PR #1, opened 2018-11-11
Recurring transfers module

g-r-a-n-t:
This implements the [recurring transfer module](https://github.com/GnosisEcosystemFund/Gnosis-Bounties-/issues/2)

Thanks for taking a look. I've enjoyed working on this and hope to get it completed soon.


contracts/RecurringTransfersModule.sol
```diff
@@ -0,0 +1,225 @@
...
+        uint256 tokenPriceNum = 1;
+        uint256 tokenPriceDen = 1;
+        // they are transfering ETH
+        if(token != 0) {
+            (tokenPriceNum, tokenPriceDen) = dutchExchange.getPriceOfTokenInLastAuction(token);
+        }
+
+        uint256 ratePriceNum;
+        uint256 ratePriceDen;
+        (ratePriceNum, ratePriceDen) = dutchExchange.getPriceOfTokenInLastAuction(rate);
```
rmeissner: I would maybe rename `rate` to `rateToken` just to make it clear that this is a token address

contracts/RecurringTransfersModule.sol
```diff
@@ -0,0 +1,225 @@
...
+        // transfer does not need to be adjusted since no rate is given
+        if(rate == 0) {
+            return amount;
+        }
+
+        uint256 tokenPriceNum = 1;
+        uint256 tokenPriceDen = 1;
+        // they are transfering ETH
+        if(token != 0) {
+            (tokenPriceNum, tokenPriceDen) = dutchExchange.getPriceOfTokenInLastAuction(token);
```
rmeissner: @anxolin could you check the dutchx logic here

contracts/RecurringTransfersModule.sol
```diff
@@ -0,0 +1,225 @@
...
+    /// @param gasPrice The maximum gas price value for this token.
+    /// @param dataGas The maximum data gas value for this token.
+    function setGasLimits(
+        address token,
+        uint256 gasPrice,
+        uint256 dataGas
+    )
+        public
+    {
+        require(OwnerManager(manager).isOwner(msg.sender), "Method can only be called by an owner");
```
rmeissner: if we want to use this, then it should use the `authorized` modifier

package.json
```diff
@@ -27,20 +29,23 @@
...
     "run-with-testrpc": "^0.3.0",
-    "solidity-coverage": "^0.4.2",
+    "solidity-coverage": "^0.4.15",
     "truffle": "^4.1.14",
     "web3": "0.20.6"
   },
   "dependencies": {
+    "@gnosis.pm/dx-contracts": "^1.0.7",
+    "ethereum-datetime": "^1.0.0",
+    "gnosis-safe": "github:gnosis/safe-contracts#v0.0.2-alpha",
```
rmeissner: could we update this to v0.1.0

test/blockTime.js
```diff
@@ -0,0 +1,47 @@
+
+function getCurrentBlockTime() {
+    return web3.eth.getBlock(web3.eth.blockNumber).timestamp
+}
+
+function fastForwardBlockTime(seconds) {
```
rmeissner: @digix/tempo includes some of that logic

contracts/RecurringTransfersModule.sol
anxolin: The logic looks ok. 

The only 2 small inputs I can give are:
- To get the new amount for GNO-DAI it's getting the prices of GNO-ETH and DAI-ETH. I find this solution simple and good enough. But note that the GNO-DAI market may also be created so `getPriceInPastAuction(token, rate, lastAuction)` could get that specific price
- The math operations, should be protected with a safe math lib, right?
- The function is internal, but I find that a user interface might find useful to show the user the amount they will pay in any token before executing the transaction. So I would make it public.

contracts/RecurringTransfersModule.sol
anxolin: I think that's a great idea. I was a bit confused at the beginning too

contracts/RecurringTransfersModule.sol
```diff
@@ -0,0 +1,225 @@
...
+        require(amount != 0, "amount must be greater than 0");
+        require(transferDay < 29, "transferDay must be less than 29");
+        require(transferHourStart > 0, "transferHourStart must be greater than 0");
+        require(transferHourEnd < 23, "transferHourEnd must be less than 23");
+        require(transferHourStart < transferHourEnd, "transferHourStart must be less than transferHourEnd");
+        recurringTransfers[receiver] = RecurringTransfer(delegate, token, rate, amount, transferDay, transferHourStart, transferHourEnd, 0);
```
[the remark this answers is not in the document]
g-r-a-n-t: I seem to remember having a conversation about this in the past. I believe this is the desired functionality.

contracts/RecurringTransfersModule.sol
```diff
@@ -0,0 +1,225 @@
...
+            require(manager.execTransactionFromModule(receiver, transferAmount, "", Enum.Operation.Call), "Could not execute Ether transfer");
+        } else {
+            bytes memory data = abi.encodeWithSignature("transfer(address,uint256)", receiver, transferAmount);
+            require(manager.execTransactionFromModule(recurringTransfer.token, 0, data, Enum.Operation.Call), "Could not execute token transfer");
+        }
+
+        recurringTransfers[receiver].lastTransferTime = now;
+
+        // We transfer the calculated tx costs to the tx.origin to avoid sending it to intermediate contracts that have made calls
+        if (gasPrice > 0) {
```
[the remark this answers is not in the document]
g-r-a-n-t: I've updated the contract with something a little closer to what's desired. There's still the issue where the delegate can sign whatever gas values they want tho.

contracts/RecurringTransfersModule.sol
```diff
@@ -0,0 +1,225 @@
...
+        address receiver,
+        address delegate,
+        address token,
+        address rate,
+        uint256 amount,
+        uint8 transferDay,
+        uint8 transferHourStart,
+        uint8 transferHourEnd
+    )
+        public
```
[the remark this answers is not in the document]
g-r-a-n-t: Both add and remove have been changed to use authorized.

contracts/RecurringTransfersModule.sol
g-r-a-n-t: I will make the following changes:
- attempt to get the value between two tokens before calculating based on the eth value
- use safe math
- make the method public
- create an interface and implement it

g-r-a-n-t: Thanks for the feedback @anxolin

g-r-a-n-t:
@GnosisEcosystemFund Could the Gitcoin expiration get changed on this again? It will expire in the next 24 hours.

contracts/RecurringTransfersModule.sol
```diff
@@ -0,0 +1,289 @@
...
+    string public constant NAME = "Recurring Transfers Module";
+    string public constant VERSION = "0.1.0";
+
+    DutchExchange public dutchExchange;
+
+    // recurringTransfers maps the composite hash of a token and account address to a recurring transfer struct.
+    mapping (address => RecurringTransfer) public recurringTransfers;
+
+    // gasPriceLimits maps a token address to a gas price limit for transfer refunds
+    mapping (address => uint256) public gasPriceLimits;
```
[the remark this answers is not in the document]
g-r-a-n-t: Hmm. I though that I had removed the refund limit functionality, but I guess not. Do you think this provides any value or should I just remove it?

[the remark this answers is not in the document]
g-r-a-n-t: I've gone ahead and removed these values for now. If you feel as though it is worthwhile to bring these back, I would be happy to do so. :)

PR #2, opened 2018-11-26
Transfer limit module

s1na:
This PR  implements the [TransferLimitModule](https://github.com/GnosisEcosystemFund/Gnosis-Bounties-/issues/1).

Note: I thought it doesn't make sense to repeat all the common code between this PR and #1 (configs, `Imports.sol`, etc.), as it'd result in merge conflicts. When that one is merged, I'll pull the changes and fix this PR.


contracts/TransferLimitModule.sol
```diff
@@ -0,0 +1,417 @@
...
+
+import "@gnosis.pm/dx-contracts/contracts/DutchExchange.sol";
+import "@gnosis.pm/dx-contracts/contracts/Oracle/PriceOracleInterface.sol";
+import "openzeppelin-solidity/contracts/math/SafeMath.sol";
+
+
+/// @title Transfer Limit Module - Allows to transfer limited amounts of ERC20 tokens and Ether.
+contract TransferLimitModule is Module, SignatureDecoder, SecuredTokenTransfer {
+    string public constant NAME = "Transfer Limit Module";
+    string public constant VERSION = "0.0.2";
```
rmeissner: Let's use a version specific to the module (not to the Safe ;) ). So either 0.0.1 or 0.1.0 oder 1.0.0 ... up to you ;)

contracts/TransferLimitModule.sol
```diff
@@ -0,0 +1,417 @@
...
+        uint256 startGas = gasleft();
+        bytes32 txHash = getTransactionHash(
+            token, to, amount,
+            safeTxGas, dataGas, gasPrice, gasToken, refundReceiver,
+            nonce
+        );
+        require(checkSignatures(txHash, signatures), "Invalid signatures provided");
+        // Increase nonce and execute transaction.
+        nonce++;
+        require(gasleft() >= safeTxGas, "Not enough gas to execute safe transaction");
```
rmeissner: This check should probably be done after we call `handleTransferLimits` (since that also uses up some gas)

contracts/TransferLimitModule.sol
```diff
@@ -0,0 +1,417 @@
...
+
+        // Validate that transfer is not exceeding transfer limit, and
+        // update state to keep track of spent values.
+        require(handleTransferLimits(token, amount), "Transfer exceeds limits");
+
+        // Perform transfer by invoking manager
+        if (token == 0) {
+            require(manager.execTransactionFromModule(to, amount, "", Enum.Operation.Call), "Could not execute ether transfer");
```
rmeissner: In the safe don't revert if the user transaction fails. Would be interesting if that makes sense here too (same for `handleTransferLimits`)

Lets assume somebody submits a transferlimit request but at the same time somebody else transfers out the last tokens from the Safe. Who should be held responsible and cover the gas costs?

contracts/TransferLimitModule.sol
```diff
@@ -0,0 +1,417 @@
...
+        uint256 gasPrice,
+        address gasToken,
+        address refundReceiver,
+        uint256 _nonce
+    )
+        public
+        view
+        returns (bytes32)
+    {
+        return keccak256(
```
rmeissner: I would like to have EIP-712 signatures here pls

contracts/TransferLimitModule.sol
```diff
@@ -0,0 +1,417 @@
...
+        TransferLimit storage transferLimit = transferLimits[token];
+        // If time period is over, reset expenditure.
+        if (isPeriodOver()) {
+            transferLimit.spent = 0;
+            totalWeiSpent = 0;
+            totalDaiSpent = 0;
+        }
+
+        // Transfer + previous expenditure shouldn't exceed limit specified for token.
+        if (SafeMath.add(transferLimit.spent, amount) > transferLimit.transferLimit) {
```
rmeissner: Can you use `SafeMath` via the `using` statemaent (https://github.com/gnosis/safe-contracts/blob/development/contracts/GnosisSafe.sol#L15)

contracts/TransferLimitModule.sol
```diff
@@ -0,0 +1,417 @@
...
+            return false;
+        }
+
+        // If a global cap is set, transfer amount + value of all
+        // previous expenditures (for all tokens) shouldn't exceed global limit.
+        if (!isUnderGlobalCap(token, amount)) {
+            return false;
+        }
+
+        transferLimits[token].spent = SafeMath.add(transferLimits[token].spent, amount);
```
rmeissner: Same here

contracts/TransferLimitModule.sol
```diff
@@ -0,0 +1,417 @@
...
+        }
+
+        // Calculate value in ether.
+        uint256 ethNum;
+        uint256 ethDen;
+        (ethNum, ethDen) = getEthAmount(token, amount);
+
+        // Convert ether to wei
+        uint256 weiAmount = SafeMath.div(SafeMath.mul(ethNum, 10**18), ethDen);
+        if (globalWeiCap > 0 && SafeMath.add(totalWeiSpent, weiAmount) > globalWeiCap) {
```
rmeissner: we could store `SafeMath.add(totalWeiSpent, weiAmount)` to avoid calculating it twice ... but not super important (might save a couple gas ... not sure)

contracts/TransferLimitModule.sol
```diff
@@ -0,0 +1,417 @@
...
+        }
+
+        return true;
+    }
+
+    function isPeriodOver()
+        internal
+        returns (bool)
+    {
+        if (rolling && getNow() > lastStartTime + timePeriod) {
```
rmeissner: > “rolling”, i.e. always consider the exact last 24h or last 7 days.

@tschubotz to my understanding this code (check for `rolling`) should be the other way around, right?

contracts/TransferLimitModule.sol
```diff
@@ -0,0 +1,417 @@
...
+        } else if (!rolling && currentStartTime() > lastStartTime) {
+            lastStartTime = currentStartTime();
+            return true;
+        }
+
+        return false;
+    }
+
+    // Return now from function to allow mocking.
+    function getNow()
```
rmeissner: hmm not the biggest fan of that, since this adds unnecessary gas costs. To change "now" during tests you could use @digix/tempo (more info here https://ethereum.stackexchange.com/questions/15755/simulating-the-passage-of-time-with-testrpc)

contracts/TransferLimitModule.sol
```diff
@@ -0,0 +1,417 @@
...
+        returns (uint256, uint256)
+    {
+        // Amount is in wei
+        if (token == 0) {
+            return (amount, 10**18);
+        }
+
+        uint256 num;
+        uint256 den;
+        (num, den) = dutchx.getPriceOfTokenInLastAuction(token);
```
rmeissner: @anxolin can you check the interaction with dutchx? :)

contracts/TransferLimitModule.sol
```diff
@@ -0,0 +1,417 @@
...
+        return (SafeMath.mul(amount, num), den);
+    }
+
+    function getDaiAmount(uint256 ethNum, uint256 den)
+        internal
+        view
+        returns (uint256, uint256)
+    {
+        PriceOracleInterface priceOracle = PriceOracleInterface(dutchx.ethUSDOracle());
+        uint256 ethDaiPrice = priceOracle.getUSDETHPrice();
```
rmeissner: @anxolin here too :)

contracts/TransferLimitModule.sol
```diff
@@ -0,0 +1,417 @@
...
+        uint256 amount = ((gasUsed - gasleft()) + dataGas) * gasPrice;
+        // Make sure refund is within transfer limits, to prevent
+        // attacker with a compromised key to empty the safe.
+        require(handleTransferLimits(gasToken, amount), "Gas refund exceeds transfer limit");
+
+        // solium-disable-next-line security/no-tx-origin
+        address receiver = refundReceiver == address(0) ? tx.origin : refundReceiver;
+        if (gasToken == address(0)) {
+            // solium-disable-next-line security/no-send
+            require(receiver.send(amount), "Could not pay gas costs with ether");
```
rmeissner: This would send tokens from the module, which is obviously not possible since there should be no funds on the module.

rmeissner:
Didn't check the tests in detail yet, but would be nice to have some that make use of the gas refund mechanism :)

contracts/TransferLimitModule.sol
s1na: Hmm, as it is now it's the relayer who bears the cost when the user transaction fails, which doesn't seem right. I'll modify it so as not to revert on user transaction failure, so that relayer can still get refunded (if any fund is in the safe!).

Regarding `handleTransferLimits`, I went with reverting because apart from validation it also modifies the state (spent values).

s1na: I just realized that if the module doesn't revert on transfer failure, the spent values could get updated in `handleTransferLimits`, without any actual transfer taking place! (for example when there's not enough balance in safe)

Should I check limits before transfer, and only update spent values after?

s1na:
@rmeissner I've tried to address some of your comments (marked resolved), but I have some questions for the others:

- Added EIP-712 signatures, should I also add support for EIP-1271 signatures?
- Re. rolling, I used the [condition](https://github.com/gnosis/MultiSigWallet/blob/master/contracts/MultiSigWalletWithDailyLimit.sol#L80) you mentioned. I can change it if its not what you meant.
- I've tried to simplify gas refund mechanism, so we can add complexity as required in next steps. Currently user simply specifies the amount they are willing to pay (`gasLimit * gasPrice` is refunded). As the tx is to a great extent deterministic, user can estimate gas (add a bit for reward), relayer can verify that this amount is enough. It would be possible to calculate `gasUsed` until `handlePayment`, and add basic tx cost + `handlePayment` costs, and cap it with `gasLimit` as an alternative. Do you think that'd be better?
- If `handleTransferLimits` returns false, module reverts. Because for current nonce the result of `handleTransferLimits` is deterministic, it can be verified off-chain by relayer before submitting transaction.
- Regarding reverting when `execTransactionFromModule` fails, I can see points for both reverting and not reverting. I'll happily oblige with what you think is best.

contracts/TransferLimitModule.sol
```diff
@@ -185,7 +192,7 @@ contract TransferLimitModule is Module, SignatureDecoder, SecuredTokenTransfer {
 
         // We transfer the calculated tx costs to the tx.origin to avoid sending it to intermediate contracts that have made calls
         if (gasPrice > 0) {
-            handlePayment(startGas, gasLimit, gasPrice, gasToken, refundReceiver);
+            handlePayment(gasLimit, gasPrice, gasToken, refundReceiver);
```
rmeissner: In this approach, why do we need `gasLimit` and `gasPrice` and not just `refund` (`gasLimit * gasPrice` can be calculated off-chain)

We can leave it like this for now, but the idea why we don't fail the whole transaction if the inner transaction fails is that the relay service will else pay for failing transactions. And these do happen even if we estimate before (e.g. investing into an ico -> this is sometimes time bound).

contracts/TransferLimitModule.sol
```diff
@@ -171,6 +170,14 @@ contract TransferLimitModule is Module, SignatureDecoder, SecuredTokenTransfer {
         // Increase nonce and execute transaction.
         nonce++;
 
+        // If time period is over, reset expenditure.
+        TransferLimit storage transferLimit = transferLimits[token];
```
rmeissner: Why to we retrieve the transferLimit here, before we know if it is required. Every storage access costs 200 gas, so that is quite expensive

contracts/TransferLimitModule.sol
s1na: @rmeissner good points. Merged `gasLimit` and `gasPrice` to `refund`.

The module now doesn't revert on `execTransactionFromModule`. It updates spent values after transfer success, and on failure emits an event. Even on failure relayer is refunded if there's enough of `gasToken` in safe.

One edge case I can think of is if a token contract returns false on failing transfers, instead of reverting. When this happens, `execTransactionFromModule` still returns true.

contracts/TransferLimitModule.sol
```diff
@@ -0,0 +1,445 @@
...
+        require(amount > 0, "Invalid amount provided");
+
+        bytes32 txHash = getTransactionHash(
+            token, to, amount,
+            refund, gasToken, refundReceiver,
+            nonce
+        );
+        require(checkSignatures(txHash, signatures), "Invalid signatures provided");
+        // Increase nonce and execute transaction.
+        nonce++;
```
rmeissner: It would be more secure to increase the nonce before we call `checkSignatures`.
In this case it should not matter too much since right now this code doesn't support contract signatures, but as soon as this is the case this would be a security risk.

contracts/TransferLimitModule.sol
```diff
@@ -0,0 +1,445 @@
...
+        }
+
+        return false;
+    }
+
+    function updateTransferLimits(address token, uint256 amount, uint256 weiSpent, uint256 daiSpent)
+        internal
+    {
+        transferLimits[token].spent = transferLimits[token].spent.add(amount);
+        if (weiSpent > 0 && weiSpent > totalWeiSpent) {
```
rmeissner: Can you comment this code pls. This might be potentially confusing. (I assume this is to check if the totalWeiSpent has increased we should update it, since 0 indicates that we should not update it)

contracts/TransferLimitModule.sol
```diff
@@ -0,0 +1,445 @@
...
+    uint256 public totalWeiSpent;
+
+    // Global limit on transfers, specified in usd (dai).
+    uint256 public globalDaiCap;
+
+    // Total amount of dai spent in current time period.
+    uint256 public totalDaiSpent;
+
+    // Number of required confirmations for a transfer.
+    uint256 public threshold;
```
rmeissner: This threshold should probably be updatable by the safe, since it is possible that the safe owner counts changes after the setup of the module and then this module cannot be used

contracts/TransferLimitModule.sol
```diff
@@ -0,0 +1,445 @@
...
+        if (token == address(0)) {
+            ok = manager.execTransactionFromModule(to, amount, "", Enum.Operation.Call);
+        } else {
+            bytes memory data = abi.encodeWithSignature("transfer(address,uint256)", to, amount);
+            ok = manager.execTransactionFromModule(token, 0, data, Enum.Operation.Call);
+        }
+
+        // Only update transfer limits if transfer happened.
+        if (ok) {
+            updateTransferLimits(token, amount, weiSpent, daiSpent);
```
rmeissner: Does this allow a reentrency attack? Since we perform the transfer that could trigger another transaction (with a valid signature and nonce) to this module and both could use the maximum limit.

s1na:
@rmeissner Apologies for the delay. Took your comments into account.

* Regarding reentrancy: I hadn't considered the owners themselves could attack and withdraw more than the limits. I added a mutex which should prevent this. If you don't like this approach, other alternatives which I thought of include: Setting spent values when checking them (before call), and if transfer fails reducing spent values again, or again setting spent values when checking them and reverting on transfer fail (as it was before).

* I rebased from master (and hence had to do a force push). After rebasing I couldn't get the tests to pass, due to an error from `bitcoin-lib` saying there are two instances, so locally I ran `rm -r node_modules/bitcore-mnemonic/node_modules/bitcore-lib` which allowed me to run tests.

s1na:
Hi folks! Is there anything I can do on my side to help with the review process?

PR #3, opened 2019-03-14
Dx module prepare release

denisgranha:


package.json
```diff
@@ -6,11 +6,15 @@
...
-    "test"
+    "test",
+    "build"
   ],
   "scripts": {
+    "preversion": "npm run restore",
+    "restore": "rm -rf build && npm run compile && tnt iN && tnt cB",
     "test-norpc": "truffle test",
     "test": "run-with-testrpc -l 20000000 --noVMErrorsOnRPCResponse true 'truffle test'",
+    "install": "cd $INIT_CWD && npm explore truffle -- npm install solc@0.5.0",
```
rmeissner: maybe use 0.5.3, since that is the version we use in the safe now

denisgranha: okk, I'm gonna update it

test/dutchXCompleteModule.js
```diff
@@ -0,0 +1,290 @@
...
+const DutchXModule = artifacts.require("./DutchXCompleteModule.sol");
+const GnosisSafe = artifacts.require("./GnosisSafe.sol");
+const ProxyFactory = artifacts.require("./ProxyFactory.sol");
+const DutchXInterface = artifacts.require("./DutchXInterface.sol")
+const TokenInterface = artifacts.require("./DutchXTokenInterface.sol")
+const MockContract = artifacts.require("./MockContract.sol")
+
+
+// Helper function to deploy a mocked tocken
+let deployWETHToken = async function(deployer) {
```
rmeissner: I would call with method `deployTestToken`

denisgranha: makes sense

contracts/DutchXBaseModule.sol
```diff
@@ -0,0 +1,151 @@
...
+        }
+        else{
+            manager = ModuleManager(_manager);
+        }
+
+        dutchXAddress = dx;
+
+        for (uint256 i = 0; i < tokens.length; i++) {
+            address token = tokens[i];
+            require(token != address(0), "Invalid token provided");
```
rmeissner: should we check `isWhitelistedToken[token] == false` to avoid duplicate tokens in the array?

contracts/DutchXBaseModule.sol
```diff
@@ -0,0 +1,151 @@
...
+            isWhitelistedToken[token] = true;
+        }
+
+        whitelistedTokens = tokens;
+
+        for (uint256 i = 0; i < operators.length; i++) {
+            address operator = operators[i];
+            require(operator != address(0), "Invalid operator address provided");
```
rmeissner: Same as above to avoid duplicate operators in array?

contracts/DutchXBaseModule.sol
denisgranha: true, we can add it as a patch in a later commit. For now is up to the user to use it well

PR #6, opened 2020-04-29
Convert to mono repo; each module has their own folder

rmeissner:
Why am I doing this?
- Allow to use different compilers for each module (old modules do not need to be updated to newest version if not used anymore)
- Clearly separate logic


denisgranha:
I like it too!

PR #16, opened 2020-10-19
Fix audit comments

rmeissner:
Reports:
1. `removeDelegate` doesn't update `delegatesStart` which will break the delegate linked list.
2. `removeDelegate` doesn't functionally remove the delegates, they can still have allowances and they can't be added back in with `addDelegate`. `removeDelegate` only affects the output of `getDelegates`
3. `ecrecover` returns 0 on failure, which will lead to this check passing, this isn't a security vulnerability if there are no allowances for the `0` address, but is still unexpected. https://github.com/gnosis/safe-modules/blob/master/allowances/contracts/AlowanceModule.sol#L241

Changes:
- Added tests for all reports
- Enforce that allowances cannot be set for the 0 address
- Enforce that delegates or their derived index cannot be the 0 address
- Enforce that the contract reverts on ecrecover failure
- Properly set `delegatesStart` if first element in linked list is removed

### DIFF
--- a/allowances/contracts/AlowanceModule.sol
+++ b/allowances/contracts/AlowanceModule.sol
@@ -305,7 +305,7 @@ contract AllowanceModule is SignatureDecoder {
         require(index != uint(0), "index != uint(0)");
         address currentDelegate = delegates[msg.sender][index].delegate;
         if(currentDelegate != address(0)) {
-            // We have a collision for the indeces of delegates
+            // We have a collision for the indices of delegates
             require(currentDelegate == delegate, "currentDelegate == delegate");
             // Delegate already exists, nothing to do
             return;
--- a/allowances/contracts/AlowanceModule.sol
+++ b/allowances/contracts/AlowanceModule.sol
@@ -73,6 +73,7 @@ contract AllowanceModule is SignatureDecoder {
     function setAllowance(address delegate, address token, uint96 allowanceAmount, uint16 resetTimeMin, uint32 resetBaseMin)
         public
     {
+        require(delegate != address(0), "delegate != address(0)");
         require(delegates[msg.sender][uint48(delegate)].delegate == delegate, "delegates[msg.sender][uint48(delegate)].delegate == delegate");
         Allowance memory allowance = getAllowance(msg.sender, delegate, token);
         if (allowance.nonce == 0) { // New token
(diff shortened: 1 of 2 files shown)


allowances/contracts/AlowanceModule.sol
```diff
@@ -296,11 +301,11 @@ contract AllowanceModule is SignatureDecoder {
     /// @dev Allows to add a delegate.
     /// @param delegate Delegate that should be added.
     function addDelegate(address delegate) public {
-        require(delegate != address(0), "Invalid delegate address");
         uint48 index = uint48(delegate);
+        require(index != uint(0), "index != uint(0)");
         address currentDelegate = delegates[msg.sender][index].delegate;
         if(currentDelegate != address(0)) {
-            // We have a collision for the indeces of delegates 
+            // We have a collision for the indeces of delegates
```
Uxio0, line 308: indexes

allowances/contracts/AlowanceModule.sol
```diff
@@ -333,8 +338,15 @@ contract AllowanceModule is SignatureDecoder {
                 emit DeleteAllowance(msg.sender, delegate, token);
             }
         }
-        delegates[msg.sender][current.prev].next = current.next;
-        delegates[msg.sender][current.next].prev = current.prev;
+        if (current.prev == 0) {
```
Uxio0, line 341: In `if(current.delegate == address(0)) return;`, shouldn't you `revert`? This would prevent users spending gas, as `call()` would look successful.

One space is missing also `if (current.delegate == address(0)) return;`

Uxio0 (approved):

allowances/contracts/AlowanceModule.sol
rmeissner: indices would be the correct one https://www.yourdictionary.com/indices#:~:text=The%20definition%20of%20indices%20is,noun

allowances/contracts/AlowanceModule.sol
rmeissner: you could revert ... you don't really safe gas (actually in case the tx is submitted the `revert` with a message, costs more gas). Only case would be that when you try to estimate it would show in some wallets.

Why I don't revert is also that you can do stuff like the following:
We have the setup with `Delegate A` and 1 safe owner proposes to `remove A and add B` and another propose to `remove A and add C` ... if both get approved you can in the current solution just execute both (as multisend tx), but if you revert, the one executed later will fail. 
Here the advantage is not that visible but in the `addDelegate` case this makes the client logic a lot easier as you can just always try to add a delegate, before setting an allowance (as just nothing will happen in case it already exists)

allowances/contracts/AlowanceModule.sol
Uxio0: D'oh :sweat_smile:

allowances/contracts/AlowanceModule.sol
Uxio0: Ok, understood :+1:

PR #22, opened 2020-12-25
Bequest module

vporton:
**Bequest Module** - Allows to bequest all funds on the wallet to be withdrawn after a given time. Moreover, after the given time the heir can execute any transaction on the inherited wallet.


vporton:
Added tests.

PR #26, opened 2021-01-15
Social recovery module

vporton:
This contract allows recovering your lost wallet by a "voting" of given number of "friends".

Tested partly. But I don't mark it as WIP, because I am not sure if other modules in this repo were also tested/audited.


vporton:
Not that this is an enhanced version of Social Recovery module from `gnosis-contracts` (from where we plan to remove modules).

rmeissner:
This PR contains 2 modules (bequest and recovery module). 

Also please provide some more information in the readme and make sure that the code is sufficiently tested. 

Example information that would be nice for the readme:
- How does this exactly work?
- Is this used anywhere?
- How can it be used?

It should be similar to what we have for the Allowance module (https://github.com/gnosis/safe-modules/tree/master/allowances).

Once this information is provided we will review this PR in detail.

vporton:
Oh, sorry, I forked a wrong branch, `bequest` should not be here. So I have rewritten the commit history to remove it.

I also added a missing feature: `turnOffSocialRecovery()` method.

The README was improved.

I didn't add more tests yet. Feel free to mark this PR as a draft. But are other modules in `safe-modules` sufficiently tested? (I don't know.) If not yet, then this PR should be merged instead of being market as a draft, because in this case it does follow the same standards as others.

social_recovery/contracts/SocialRecoveryModule.sol
```diff
@@ -0,0 +1,148 @@
...
+
+    /// @dev Returns hash of data encoding owner replacement.
+    /// @param data Data payload.
+    /// @return Data hash.
+    function getDataHash(bytes memory data)
+        public
+        pure
+        returns (bytes32)
+    {
+        return keccak256(data);
```
rmeissner, line 146: This should probably be following EIP-712

social_recovery/contracts/SocialRecoveryModule.sol
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    /// @dev Allows a friend to confirm a Safe transaction.
+    /// @param dataHash Safe transaction hash.
+    function confirmTransaction(bytes32 dataHash)
+        public
+        onlyFriend
+    {
+        require(!isExecuted[dataHash], "Recovery already executed");
+        isConfirmed[dataHash][msg.sender] = true;
```
rmeissner, line 100: This should probably trigger an event

social_recovery/contracts/import.sol
```diff
@@ -0,0 +1,3 @@
+pragma solidity >=0.5.0 <0.7.0;
```
rmeissner, line 1: this filename should have a capital `I`

social_recovery/contracts/libraries/CreateAndAddModules.sol
```diff
@@ -0,0 +1,43 @@
+pragma solidity >=0.5.0 <0.7.0;
+import "@gnosis.pm/safe-contracts/contracts/base/Module.sol";
+
+
+/// @title Create and Add Modules - Allows to create and add multiple module in one transaction.
+/// @author Stefan George - <stefan@gnosis.pm>
+/// @author Richard Meissner - <richard@gnosis.pm>
+contract CreateAndAddModules {
```
rmeissner, line 8: Why do we need this here? This contract should be part of the gnosis safe package

social_recovery/contracts/SocialRecoveryModule.sol
```diff
@@ -0,0 +1,148 @@
+pragma solidity >=0.5.0 <0.7.0;
+import "@gnosis.pm/safe-contracts/contracts/base/Module.sol";
+import "@gnosis.pm/safe-contracts/contracts/base/ModuleManager.sol";
+import "@gnosis.pm/safe-contracts/contracts/base/OwnerManager.sol";
+import "@gnosis.pm/safe-contracts/contracts/common/Enum.sol";
+
+
+/// @title Social Recovery Module - Allows to replace an owner without Safe confirmations if friends approve the replacement.
+/// @author Stefan George - <stefan@gnosis.pm>
```
rmeissner, line 9: I would adjust this to include yourself and me (as the original version of stefan looked quite different ;) )

patcito:
Hey @rmeissner, this looks great but what if your friends turn evil or get hacked? They could steal your fund without notice. A way to make this less likely would be to force a timelapse after the recovery vote. So the `setup` function would take an additional `timelapse` params that would be a number of seconds.

When friends vote to `recoverAccess`, it would set a `voteTimestamp` variable to `now` and then they would need to call a `confirmRecoverAccess` later that would check if current timestamp is greater than `voteTimestamp+timelapse`. During this time, the owner should be able to call `turnOffSocialRecovery()` to cancel the vote and remove his friends. What do you think?

PR #66, opened 2023-02-07
Do not allow implicit casting of `USE_SAFE_SINGLETON_FACTORY` variable, add celo config

mmv08:
This PR:
- fixes a bug where any string set as `USE_SAFE_SINGLETON_FACTORY` would eval as a truthy value
- add configuration for celo network


allowances/migrations/1_deploy_allowance_modules.js
```diff
@@ -1,13 +1,13 @@
-const { deployTruffleContract } = require('@gnosis.pm/singleton-deployer-truffle');
-const AllowanceModule = artifacts.require("AllowanceModule");
+const { deployTruffleContract } = require("@gnosis.pm/singleton-deployer-truffle")
```
mikhailxyz, line 1: my IDE was extremely unhappy about the JS formatting in this file so it formatted it

rmeissner (approved):

PR #67, opened 2023-03-26
Add Neon Devnet Support

Sj-001:
RPC: https://devnet.neonevm.org
@Uxio0


Sj-001:
Hey @Uxio0, I am trying to deploy AllowanceModule using the singleton factory 0x914d7fec6aac8cd542e72bca78b30650d45643d7
I get the deployed address as 0xDF3F30394dCdE3b807878569d1d520703c8f769b
It doesn't seem to be right. Can you please suggest some fix if needed?

RPC: https://devnet.neonevm.org
Explorer: https://neonscan.org/

mmv08:
The repo for safe modules deployments is here - https://github.com/safe-global/safe-modules-deployments. Please create a PR there.

Regarding the different addresses, it seems like truffle attaches compiler metadata to the deployment bytecode, resulting in a different address. You can try replaying the creation tx on your chain, and it should deploy it to the same address.

PR #88, opened 2023-10-06
Closes #81: implement 4337 exec flow

### LINKED ISSUE #81: [4337 Module] Implement execution flow
- Uses legacy fallback handler and module functionality
- Uses `checkSignatures` to provide multisignature flow (analog to `execTransaction` flow`)
- Can assume that the entrypoint is trusted
- Only has to be tested against unit tests
  - This implies that the implementation might not be 100% compatible at this point and might need to be adjusted when tests with prod bundlers are added

Technical details:
- Module contract handle signature validation and fee payment
  - Should check that the execution function is `execTransactionFromModule`
- Execution is done by whitelisting the entry point as a module and calling `execTransactionFromModule`

rmeissner:
Closes #81

- Make contract compatible with EntryPoint v0.6.0
  - Use nonce from EntryPoint


rmeissner:
Consider for the future to use the hash provided by the entry point

mmv08 (approved):

PR #89, opened 2023-10-06
Closes #82: implement 4337 deploy flow

### LINKED ISSUE #82: [4337 Module] Implement deployment flow
- Make it possible that the module can be enabled on Safe deployment

Possible approaches:
- Use custom library for initialization
- Use custom deployment factory

rmeissner:
Closes #82

- Update documentation to reflect the necessity for a new library contract
- Add `AddModulesLib` to add multiple modules on deployment
- Update Safe dependency to 1.4.1
- Add simple tests for deployment flow


4337/README.md
```diff
@@ -48,22 +48,18 @@ Important to note that [ERC-4337](https://eips.ethereum.org/EIPS/eip-4337#first-
 
 > The initCode field (if non-zero length) is parsed as a 20-byte address, followed by "calldata" to pass to this address.
 
+To deploy a Safe with 4337 directly enabled we require a setup library that enables multiple modules (`AddModulesLib`). This is necessary because to enable a Module the Safe has to do a call to itself. Before calling setup we do not know the address of the Safe yet (as the address depends on the setup parameters) and using MultiSend will not result in the correct `msg.sender` for a selfcall.
```
mmv08, line 51:
```suggestion
To deploy a Safe with 4337 directly enabled, we require a setup library that enables multiple modules (`AddModulesLib`). This is necessary because to enable a Module, the Safe has to do a call to itself. Before calling the setup method, we do not know the address of the Safe yet (as the address depends on the setup parameters) and using MultiSend will not result in the correct `msg.sender` for a selfcall.
```

4337/contracts/interfaces/Safe.sol
```diff
@@ -49,4 +49,11 @@ interface ISafe {
      * @return next Start of the next page.
      */
     function getModulesPaginated(address start, uint256 pageSize) external view returns (address[] memory array, address next);
+
+    /**
+     * @notice Enables the module `module` for the Safe.
+     * @dev This can only be done via a Safe transaction.
+     * @param module Module to be whitelisted.
```
mmv08, line 56:
```suggestion
     * @param module Module to be enabled.
```

mmv08 (commented):
minor docs improvements

mmv08 (approved):

4337/src/deploy/deploy_libraries.ts
```diff
@@ -0,0 +1,20 @@
+import { DeployFunction } from 'hardhat-deploy/types'
+import { HardhatRuntimeEnvironment } from 'hardhat/types'
+
+const ENTRY_POINT = process.env.DEPLOY_ENTRY_POINT
```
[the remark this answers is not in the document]
rmeissner: It is for the prod deployment, as we need to set the trusted entrypoint. And currently the one for our tests is different than the one on prod.

PR #95, opened 2023-10-16
ERC4337 module feature: Extract execution to a separate function and require success in the internal call

mmv08:
This PR:
- Extracts the execution function of a user operation into a separate function on the module (previously, the entrypoint would directly call `execTransactionFromModule` and had to be enabled as a module as well
- This was done to require `success` and revert on failing transactions to enable a proper gas estimation

This PR is a prototype only at the moment, and the goal is to seek the team's feedback on the approach.

### DIFF
--- a/4337/contracts/EIP4337Module.sol
+++ b/4337/contracts/EIP4337Module.sol
@@ -9,7 +9,7 @@ import {ISafe} from "./interfaces/Safe.sol";
 
 
 /// @title EIP4337Module
-abstract contract EIP4337Module is HandlerContext, CompatibilityFallbackHandler {
+contract Simple4337Module is HandlerContext, CompatibilityFallbackHandler {
     using UserOperationLib for UserOperation;
 
     // value in case of signature failure, with no time-range.
@@ -24,11 +24,9 @@ abstract contract EIP4337Module is HandlerContext, CompatibilityFallbackHandler 
         );
 
     address public immutable supportedEntryPoint;
-    bytes4 public immutable expectedExecutionFunctionId;
 
-    constructor(address entryPoint, bytes4 executionFunctionId) {
+    constructor(address entryPoint) {
         supportedEntryPoint = entryPoint;
-        expectedExecutionFunctionId = executionFunctionId;
     }
 
     /// @dev Validates user operation provided by the entry point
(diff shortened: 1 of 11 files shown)


4337/contracts/EIP4337Module.sol
```diff
@@ -59,6 +61,18 @@ abstract contract EIP4337Module is HandlerContext, CompatibilityFallbackHandler
...
 
+    /// @notice Executes user operation provided by the entry point
+    /// @dev Reverts if unsuccessful
+    /// @param executionData Execution data. Expects it to be a call to the `execTransactionFromModule` function of the Safe.
+    ///                      Validated in the `validateUserOp` function.
+    function executeUserOp(bytes calldata executionData) external {
+        address entryPoint = _msgSender();
+        require(entryPoint == supportedEntryPoint, "Unsupported entry point");
+
+        (bool success, ) = msg.sender.call(executionData);
```
mmv08, line 72: do you think we'd benefit from an additional check for the `execTransactionFromModule` signature?

4337/contracts/EIP4337Module.sol
```diff
@@ -59,6 +61,21 @@ abstract contract EIP4337Module is HandlerContext, CompatibilityFallbackHandler
...
 
+    /// @notice Executes user operation provided by the entry point
+    /// @dev Reverts if unsuccessful
+    /// @param executionData Execution data. Expects it to be a call to the `execTransactionFromModule` function of the Safe.
+    ///                      Validated in the `validateUserOp` function.
+    function executeUserOp(bytes calldata executionData) external {
+        address entryPoint = _msgSender();
+        require(entryPoint == supportedEntryPoint, "Unsupported entry point");
+
+        (bool outerCallSuccess, bytes memory returnData) = msg.sender.call(executionData);
```
rmeissner, line 72: why not use `execTransactionFromModule` directly, as we do in the validate function. This way we do not need to do any manual decoding here and asume types.

I assume that there would be some gas overhead right? If it is about this, then I would at least check the first 4 bytes to avoid that it calls another function

4337/contracts/EIP4337Module.sol
```diff
@@ -59,6 +61,21 @@ abstract contract EIP4337Module is HandlerContext, CompatibilityFallbackHandler
...
+        bool internalCallSuccess = returnData[31] == 0x01;
```
rmeissner, line 73: We should do a length check on `returnData`

rmeissner (commented):
The approach should work and be gas efficient, would just add 2 checks

rmeissner (approved):

rmeissner (approved):

PR #98, opened 2023-10-18
Feature: Add an optional ERC20 transfer in the `runOp.ts` script to test contract calls

mmv08:
This PR:
- Adds an optional ERC20 transfer execution in the `runOp` script to test contract calls
- Executes if the environment variable for the token address is set. The pre-requisite is Safe should own at least one token. It will transfer precisely 1.
- The token name is coming from https://www.behindthename.com/random/random.php?gender=both&number=2&sets=1&surname=&all=yes

Notes:
- I had to bump the compiler version to `0.8.20` because of openzeppelin's token requirements and change the default EVM target from `Shanghai` to `London`, because Mumbai didn't support the `PUSH0` opcode added in `Shanghai`
- I also had to bump the target ES spec to `es2020` to use BigInts

### DIFF
--- a/4337/.env.sample
+++ b/4337/.env.sample
@@ -19,6 +19,8 @@ SCRIPT_ADD_MODULES_LIB_ADDRESS=""
 SCRIPT_PROXY_FACTORY_ADDRESS=
 # Singleton address to use in the user op test script
 SCRIPT_SAFE_SINGLETON_ADDRESS=
+# ERC20 Token to use in the user op test script for testing contract calls
+SCRIPT_ERC20_TOKEN_ADDRESS=
 
 # Entrypoint address to use for the module deployment
 DEPLOY_ENTRY_POINT=""
--- a/4337/.env.sample
+++ b/4337/.env.sample
@@ -12,16 +12,16 @@ SCRIPT_DEBUG=true
 # Bundler to use in the user op test script
 SCRIPT_BUNDLER_URL=""
 # Module to use in the user op test script
-SCRIPT_MODULE_ADDRESS=
+SCRIPT_MODULE_ADDRESS=""
 # Library to add modules in the user op test script
 SCRIPT_ADD_MODULES_LIB_ADDRESS=""
 # Proxy factory to use in the user op test script
-SCRIPT_PROXY_FACTORY_ADDRESS=
+SCRIPT_PROXY_FACTORY_ADDRESS=""
 # Singleton address to use in the user op test script
-SCRIPT_SAFE_SINGLETON_ADDRESS=
+SCRIPT_SAFE_SINGLETON_ADDRESS=""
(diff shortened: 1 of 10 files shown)


4337/.env.sample
```diff
@@ -19,6 +19,9 @@ SCRIPT_ADD_MODULES_LIB_ADDRESS=""
 SCRIPT_PROXY_FACTORY_ADDRESS=
 # Singleton address to use in the user op test script
 SCRIPT_SAFE_SINGLETON_ADDRESS=
+# ERC20 Token to use in the user op test script for testing contract calls. Token should have 18 decimals. Safe should have 1 token.
+# Optional. If set, the runOp script will execute a token transfer out of the Safe
+SCRIPT_ERC20_TOKEN_ADDRESS=
```
nlordell, line 24: nitpick: Not related to this PR, but I noticed some slight inconsistencies in this file when declaring variables with empty values (some are `VAR=""` and some are `VAR=`); might be nice to clean up for consistency.

4337/hardhat.config.ts
```diff
@@ -47,7 +47,7 @@ const userConfig: HardhatUserConfig = {
     sources: 'contracts',
   },
   solidity: {
-    compilers: [{ version: primarySolidityVersion, settings: soliditySettings }, { version: '0.6.12' }, { version: '0.5.17' }],
+    compilers: [{ version: primarySolidityVersion, settings: { evmVersion: 'london', ...soliditySettings }}, { version: '0.6.12' }, { version: '0.5.17' }],
```
nlordell, line 50: question: out of curiosity, why is specifying `london` EVM version needed?

4337/contracts/test/TestToken.sol
```diff
@@ -0,0 +1,11 @@
+// SPDX-License-Identifier: LGPL-3.0-only
+pragma solidity >=0.8.0;
+
+import "@openzeppelin/contracts/token/ERC20/ERC20.sol";
+import "@openzeppelin/contracts/token/ERC20/extensions/ERC20Permit.sol";
+
+contract HariWillibaldToken is ERC20, ERC20Permit {
+    constructor(address initialMintDest) ERC20("Hari Willibald Token", "HWT") ERC20Permit("Hari Willibald Token") {
```
nlordell, line 8: Who is "Hari Wallibal"?! 🤣

mmv08: @rmeissner might have more context on this

nlordell (approved):
LGTM

4337/hardhat.config.ts
mmv08: I actually realised that `london` is not the latest one before the Shanghai upgrade, was tricked by some random website. I've adjusted it to Paris here: https://github.com/safe-global/safe-modules/pull/99

PR #99, opened 2023-10-18
Feature: Adjust compiler settings and document optimizer usage

mmv08:
This PR:
- Adjusts the compiler settings to use the latest `solc` compiler and document the settings

### DIFF
--- a/4337/README.md
+++ b/4337/README.md
@@ -136,6 +136,14 @@ npx hardhat --network <network> etherscan-verify
 npx hardhat --network <network> local-verify
 ```
 
+### Compiler settings
+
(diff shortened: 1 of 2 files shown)


nlordell (approved):

nlordell:
Out of curiosity, is the opcode `PUSH0`, and which chains don't support it still?

mmv08:
> Out of curiosity, is the opcode `PUSH0`, and do you know off the top of your head which chains don't support it still (if not, don't worry about answering)?

Yes, it's the one. In my case, it was polygon Mumbai and a local hardhat node :D

nlordell:
Just noticed, but the `solc` version in `package.json` is still `0.8.14`. I think this would affect the "local verify" task.

mmv08:
> Just noticed, but the `solc` version in `package.json` is still `0.8.14`. I think this would affect the "local verify" task.

Yes, and also the code size one. I don't know if relying on a separate solc dependency is optimal instead of using the one from Hardhat. But in the meantime, it would be great if you adjust it in your testing PR, or I can create a separate PR to address this.

Issue #87, opened 2023-10-05
[4337 Module] Increase test coverage to 99+%


rmeissner:
- Update hardhat version
- All branches need to be covered
- (optional) Add separate tests using latest bytecode for entrypoint on official address

nlordell:
Updating hardhat version being worked on in #100
Reduced code to cover in #104

nlordell:
Looks like GitHub auto-closed this issue with the #108 PR ("partially fixes $ISSUE" is interpreted as "fixes $ISSUE") 🙈.

nlordell:
Closing as the required tasks are implemented. The good to have isn't quite done, but there is a draft PR open for it.

PR #100, opened 2023-10-19
Update Hardhat and Related Depdendencies

nlordell:
Part of #87 

This PR updates `hardhat` and related dependencies to the latest version. The choice to also update the `hardhat-*` packages was mostly to be sure that there aren't any unexpected incompatibilities with the actual `hardhat` version being used (those changes can be reverted if necessary).

Some notable changes that may seem unrelated but worth clarifying:
- `solidity-coverage`: bump is required to fix integration with [newer `hardhat` versions](https://github.com/sc-forks/solidity-coverage/pull/796).
- `solc`: A follow up to #99, where the direct `solc` version didn't change to reflect the `hardhat` compiler settings
- `@types/node`: Updated to the LTS version of node, which we intend to use.

Another interesting change is the `@nomiclabs/hardhat-waffle` dependency was changed to an exact version. More recent versions of this package pull in `ethereum-waffle` v4, which transitively pulls in `ganache` v7, whose TypeScript definitions are causing compilation errors (I tried updating the compiler with hopes that it might fix the build errors but it did not).


mmv08 (commented):
I suggest using the `@nomicfoundation/hardhat-toolbox` umbrella package to compose all the dependencies in one umbrella package. It's also the current hardhat's recommended way to bootstrap a project. More info can be found here:
https://hardhat.org/hardhat-runner/plugins/nomicfoundation-hardhat-toolbox

For an example code, you can check out the allowance module in this repo or safe-contracts repository

mmv08:
Also, when I was migrating the safe-contracts repo to the latest hardhat tooling, I had many problems with Waffle, so I had to remove it, and it was well worth it, I didn't lose any functionality. All the APIs exist in hardhat. Another reason for this is zksync compatibility - achieving this with as few dependencies as possible is easier.

nlordell:
Superceded by #103

PR #103, opened 2023-10-19
Migrate to `hardhat-toolbox` and Ethers.js v6

nlordell:
Supersedes #100 

This PR implements the suggestion from https://github.com/safe-global/safe-modules/pull/100#pullrequestreview-1687802400 (thanks @mmv08 for the suggestion!), and migrates the 4337 package to use the latest `hardhat` with `@nomicfoundation/hardhat-toolbox` in order to simplify dependency management (it also removes a lot of extra dependencies that were pulled in by `ethereal-waffle` 🎉).

Note that, in order to use the latest version of `hardhat` and `hardhat-toolbox`, the codebase needed to migrate to Ethers.js v6 which caused some fairly small changes to the existing TypeScript codebase.

Some additional dependency version changes that may seem unrelated but worth clarifying:

* `solc`: A follow up to https://github.com/safe-global/safe-modules/pull/99, where the direct `solc` version didn't change to reflect the hardhat compiler settings
* `@types/node`: Updated to the LTS version of node, which we intend to use.

Additionally, for testing I was able to execute the `runOp` script ([tx](https://goerli.etherscan.io/tx/0xc1edbf6eb28c37f9df8df1bf44b79bf87fe037b1bf114278718276379bc75ace)).


4337/src/utils/execution.ts
```diff
@@ -79,8 +77,8 @@ export const safeApproveHash = async (
   if (!skipOnChainApproval) {
     if (!signer.provider) throw Error('Provider required for on-chain approval')
     const chainId = (await signer.provider.getNetwork()).chainId
-    const typedDataHash = utils.arrayify(calculateSafeTransactionHash(safe, safeTx, chainId))
-    const signerSafe = safe.connect(signer)
+    const typedDataHash = ethers.getBytes(await calculateSafeTransactionHash(safe, safeTx, chainId))
+    const signerSafe = safe.connect(signer) as typeof safe;
```
nlordell, line 81: I'm not sure if this is the idiomatic way of doing this. What is happening here is that `safe` is of type `Contract` (which behaves like a `Record<string, BaseContractMethod>`), but connect returns a `BaseContract` without access to methods.

The rationale for using `as typeof safe` instead of `as Contract` is to convey that `connect` returns the same thing as the original `safe` (regardless of what type it is), and would continue to work if we were to change `safe` to use a TypeChain generated strongly typed contract type.

4337/src/utils/execution.ts
```diff
@@ -1,9 +1,7 @@
-import { Contract, Wallet, utils, BigNumber, BigNumberish, Signer, PopulatedTransaction } from 'ethers'
-import { TypedDataSigner } from '@ethersproject/abstract-signer'
-import { AddressZero } from '@ethersproject/constants'
+import { BigNumberish, Contract, ContractTransaction, Signer, TransactionResponse, Wallet, ethers } from 'ethers'
 
-export const TRANSACTION_TO_EXECUTE_SLOT = utils.keccak256(utils.toUtf8Bytes('eip4337diatomic.transaction_hash_to_execute'))
-export const SAFE_EIP4337_NONCE_SLOT = utils.keccak256(utils.toUtf8Bytes('eip4337diatomic.nonce'))
+export const TRANSACTION_TO_EXECUTE_SLOT = ethers.id('eip4337diatomic.transaction_hash_to_execute')
+export const SAFE_EIP4337_NONCE_SLOT = ethers.id('eip4337diatomic.nonce')
```
mmv08, line 4: I think these can be removed because they're coming from an older iteration of the module where the entrypoint didn't track the nonce, and we had to track validated transactions and nonce in the module.

4337/src/utils/execution.ts
```diff
@@ -91,7 +89,7 @@ export const safeApproveHash = async (
 }
 
 export const safeSignTypedData = async (
-  signer: Signer & TypedDataSigner,
+  signer: Signer,
   safe: Contract,
```
mmv08, line 93: We could change the parameter type to the safe address, couldn't we? (Just saying, this may be on the border of the scope of the PR)

4337/src/utils/userOp.ts
```diff
@@ -156,32 +166,28 @@ export const buildUserOperationFromSafeUserOperation = ({
...
-  })
+  const requiredGas = getRequiredGas(userOp)
+  const requiredPrefund = (BigInt(requiredGas) * BigInt(userOp.maxFeePerGas)).toString()
+  console.log({ requiredGas, requiredPrefund })
 
-  return BigNumber.from(getRequiredGas(userOp)).mul(BigNumber.from(userOp.maxFeePerGas)).toString()
+  return requiredPrefund
 }
 
 export const calculateIntermediateTxHash = (callData: string, nonce: BigNumberish, entryPoint: string, chainId: BigNumberish): string => {
```
mmv08, line 185: This can also be removed, I think. We no longer have intermediate tx hashes because the validated transactions are tracked in the entry point now.

4337/test/eip4337/EIP4337ModuleExisting.spec.ts
```diff
@@ -12,14 +11,13 @@ import {
 import { chainId } from '../utils/encoding'
 
 describe('EIP4337Module - Existing Safe', async () => {
-  const [user1] = waffle.provider.getWallets()
+  const [user1] = await ethers.getSigners()
```
mmv08, line 14: When I was doing the migration for the safe-contracts repo, I had an issue with top-level async functions. I described it in this PR: https://github.com/safe-global/safe-contracts/pull/641

Quote:
> The migration showed some problems in how the tests were run. Some asynchronous functions were not properly accounted for. This was hard to spot if you run tests 1 by 1, but if you check out to the latest version and run hardhat test --parallel, then you can observe that the test suite finishes without waiting for all tests to finish
Here's an example of a problematic code block:
```
describe("SignMessageLib", async () => {
    const [user1, user2] = waffle.provider.getWallets();
```
> It looks like the test runner doesn't properly handle top-level async functions. As a solution, the wallet retrieval was moved into the setupTests function for each test, and the individual test can access it from the function's return values.

This was also confirmed by one of the hardhat maintainers here: https://github.com/NomicFoundation/hardhat/issues/3117#issuecomment-1235346851

This may be fine now, but we shall keep it in mind. It'd be great to address it in this PR if you have time.

4337/src/utils/execution.ts
mmv08: Hardhat toolbox has `typechain` types generation out of the box, so you could import the Safe type from the type folder:
```
import type { Safe } from '../../typechain-types/'
```

And then change the `safe` variable type to `Safe`, and then `.connect` should be correctly typed to return `Safe`. This may require adjustments to places where the function is used.

4337/package.json
```diff
@@ -43,46 +43,37 @@
...
     "typescript": "^4.7.3",
     "yargs": "^17.5.1"
   },
-  "peerDependencies": {
-    "ethers": "^5.1.4"
-  },
   "overrides": {
     "@safe-global/safe-contracts": {
-      "ethers": "^5.6.8"
+      "ethers": "^6.8.0"
```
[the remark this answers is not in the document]
nlordell: > Or does npm not allow installing dependencies because of incompatible packages in case there's no override?

The issue I ran into is that both `@nomicfoundation/hardhat-toolbox` and `@safe-global/safe-contracts` have an `ethers` peer dependency with non-overlapping version ranges. AFAICT, NPM does not support this (looked for a bit online but didn't find anything conclusive, I may be wrong though).

Furthermore, I noticed that we use `@safe-global/safe-contracts` only for contracts and not exported TypeScript/JavaScript code, so even if we override the `ethers` version, it should not cause any problems.

And finally, the other thing that I found confusing was that [the `package.json` in the `safe-contracts` repo](https://github.com/safe-global/safe-contracts/blob/main/package.json) does not include an `ethers` dependency or peer dependency, while the one [published to NPM does](https://unpkg.com/browse/@safe-global/safe-contracts@1.4.1/package.json).

4337/test/eip4337/EIP4337ModuleExisting.spec.ts
nlordell: Oh, sneaky.

Yes, I can just move this to the setup function right? Will include in this PR.

4337/src/utils/execution.ts
nlordell: One thing I noticed with `typechain` is that it adds a `connect` method with correct types, but it does not do the same for the `attach` method.

So, if you have:

```
let safe: Safe;
let otherSafe = safe.attach(otherAddress);
```

Then `otherSafe` will have a `BaseContract` type 😞.

4337/test/eip4337/EIP4337ModuleExisting.spec.ts
mmv08: Yes, that's what I did in the Safe contracts repo

4337/src/utils/execution.ts
mmv08: > One thing I noticed with `typechain` is that it adds a `connect` method with correct types, but it does not do the same for the `attach` method.
> 
> So, if you have:
> 
> ```ts
> let safe: Safe;
> let otherSafe = safe.attach(otherAddress);
> ```
> 
> Then `otherSafe` will have a `BaseContract` type 😞.

Yes, it is true, and that's annoying. But you can use the `getContractAt("ContractName", address)` (I hope that's the correct function name) and it'll be correctly typed.

4337/package.json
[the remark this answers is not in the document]
mmv08: > And finally, the other thing that I found confusing was that [the package.json in the safe-contracts repo](https://github.com/safe-global/safe-contracts/blob/main/package.json) does not include an ethers dependency or peer dependency, while the one [published to NPM does](https://unpkg.com/browse/@safe-global/safe-contracts@1.4.1/package.json).

I think it was a result of the upgrade to the hardhat-toolbox as well since ethers is already a dependency of the hardhat-toolbox package. But this upgrade happened after 1.4.1, which has not been released yet.

4337/src/utils/execution.ts
nlordell: Oh, thanks.

It turns out that the TypeChain configuration was not setup correctly (missing including the type chain output directory in the tsconfig).

`getContractAt` worked like a charm - thanks for the tip.

4337/src/utils/userOp.ts
nlordell: It is indeed not used anywhere else.

4337/src/utils/execution.ts
nlordell: I changed where possible to use the `safeAddress` as a parameter instead of a full `Safe` contract instance.

4337/test/eip4337/EIP4337ModuleExisting.spec.ts
nlordell: Fixed.

mmv08 (approved):
Good job sir

nlordell:
Thanks for your help 😄

PR #104, opened 2023-10-23
Increase Coverage By Removing Code

nlordell:
In implementing #87, I realised that a lot of the missing code coverage by lines were in the `UserOperationLib` source file. Since this is only used by test contracts (in particular the test entry point and the mock 4337 Safe), I moved the code around/cleaned up dead code to increase the coverage of the "main" contracts. In particular:

* The `UserOperation` struct definition moved to the 4337 interface file (felt natural as the struct is part of the 4337 account interface)
* Moved the `UserOperationLib` library to the `test/` subdirectory, the only place it was being used
  * I also got rid of some dead code, library functions that weren't being used anywhere
* Deduplicated the `Account` interface from the test entry point
* Fixed some Solidity warning lints (that weren't causing CI errors - should we change CI to fail on these warnings as well?)


4337/contracts/test/SafeMock.sol
```diff
@@ -164,7 +163,8 @@ contract Safe4337Mock is SafeMock {
         _validateSignatures(entryPoint, userOp);
 
         if (requiredPrefund != 0) {
-            entryPoint.call{value: requiredPrefund}("");
+            (bool success, ) = entryPoint.call{value: requiredPrefund}("");
+            success;
```
nlordell, line 167: Suppress the Solidity compiler warning that "result from a low level call was not used".

4337/contracts/test/SafeMock.sol
```diff
@@ -307,7 +307,7 @@ contract Safe4337Mock is SafeMock {
         checkSignatures(operationHash, operationData, userOp.signature);
     }
 
-    function validateReplayProtection(UserOperation calldata userOp) internal {
+    function validateReplayProtection(UserOperation calldata userOp) internal view {
```
nlordell, line 310: Fixes a Solidity compiler warning that the access of the function could be restricted.

mmv08:
> Fixed some Solidity warning lints (that weren't causing CI errors - should we change CI to fail on these warnings as well?)

Good with me

mmv08 (approved):

PR #107, opened 2023-10-23
Skip Coverage for Test Contracts

nlordell:
As test coverage for test contracts is not really important, we should skip it in the coverage report (so that it only reflects test coverage for the core contracts of the repo).


mmv08 (approved):
The increase in coverage is impressive!

PR #108, opened 2023-10-24
Increase branch Coverage

### LINKED ISSUE #87: [4337 Module] Increase test coverage to 99+%

nlordell:
Partially fixes #87

This PR adds unit tests to cover the remaining branches from the `SimpleEIP4337Module` contract. With this, we should have 100% coverage for the core contracts in the repo (although, not all paths in the test contracts are covered, which should be fine).

In terms of test organisation, I added a `EIP4337Module.spec.ts` for unit tests directly targeting the module functions (`validateUserOp` and `execUserOp*`).


4337/test/eip4337/EIP4337Module.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+import {
+  buildSafeUserOp,
+  calculateSafeOperationHash,
+  buildUserOperationFromSafeUserOperation,
+  buildSafeUserOpTransaction,
+  signSafeOp,
+} from '../../src/utils/userOp'
+import { chainId } from '../utils/encoding'
+
+describe('Simple4337Module', async () => {
```
mmv08, line 15:
```suggestion
describe('Simple4337Module', () => {
```
This doesn't need to be async anymore

4337/test/eip4337/EIP4337ModuleExisting.spec.ts
```diff
@@ -128,6 +127,22 @@ describe('EIP4337Module - Existing Safe', async () => {
       expect(emittedRevert).to.be.true
     })
 
+    it('executeUserOpWithErrorString should execute contract calls', async () => {
```
mmv08, line 130: One thing I actually forgot to do in the `executeUserOpWithErrorString` PR is to add tests for it to the EIP4337ModuleNew.spec.ts 🙈 would be amazing if you add them

mmv08 (approved):
Looks good, just two minor suggestions

4337/test/eip4337/EIP4337Module.spec.ts
nlordell: Nice catch - can probably be removed everywhere actually, now that the `let [user] = await ethers.getSigners()` moved to the async `setup` function.

4337/test/eip4337/EIP4337ModuleExisting.spec.ts
nlordell: Added - the revert propagation test was also missing which I added (asserting that the Safe gets deployed even if the user op reverts).

4337/test/eip4337/EIP4337Module.spec.ts
nlordell: 300395be039b3963834216463be2cab93705ef1d

4337/test/eip4337/EIP4337ModuleExisting.spec.ts
nlordell: 4f8cd914bbd7a3fd86c4a4506f1d1b2f6b052e5c

PR #109, opened 2023-10-24
Format Source

nlordell:
This PR just formats the source in the 4337 module repo so it is consistent. It is done with `prettier --ignore-path .gitignore --write .` (I updated the `package.json` to include a `fmt` script that calls this exact incantation). Additionally, I noticed that `.vscode/launch.json` was formatted, but it contained old Truffle stuff that is clearly not being used so I opted to remove the file instead of including its formatted source in this PR.

~~As a follow up question - do you think it is worth enforcing formatting of TypeScript code at a CI level or not?~~

Furthermore, this adds a check in CI to ensure that files are properly formatted.


mmv08:
> As a follow up question - do you think it is worth enforcing formatting of TypeScript code at a CI level or not?

Absolutely

mmv08 (approved):

PR #115, opened 2023-10-26
4337 Module: Import interfaces from the `@aa/contracts` package

mmv08:
After @nlordell added the `@account-abstraction/contracts` package in https://github.com/safe-global/safe-modules/pull/111, I thought it would be a good idea to remove the interfaces and import them from the package instead.


4337/contracts/test/SafeMock.sol
```diff
@@ -125,8 +125,6 @@ contract Safe4337Mock is SafeMock {
             "Unsupported execution function id"
         );
 
-        // We need to make sure that the entryPoint's requested prefund is in bounds
-        require(requiredPrefund <= userOp.requiredPreFund(), "Prefund too high");
```
mmv08, line 129: This check comes from the 1st iteration of the module in early 2022, where we wanted to validate everything entrypoint passes. We don't have this check in the production module now, so removing it from the mock is also fine.

nlordell (approved):
LGTM

PR #137, opened 2023-11-07
Fix TypeScript Lints

nlordell:
This PR just fixes some TypeScript lints. Additional comments included inline 👇.

### DIFF
--- a/4337/src/utils/safe.ts
+++ b/4337/src/utils/safe.ts
@@ -88,6 +88,7 @@ const actionCalldata = (action: MetaTransaction): string => {
 }
 
 export interface RpcProvider extends Provider {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
   send(method: string, params: unknown[]): Promise<any>
 }
 
@@ -98,6 +99,7 @@ export class MultiProvider4337 extends JsonRpcProvider {
     this.generalProvider = generalProvider
   }
 
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
   send(method: string, params: unknown[]): Promise<any> {
     if (
       [
--- a/4337/test/e2e/ReferenceEntryPoint.spec.ts
+++ b/4337/test/e2e/ReferenceEntryPoint.spec.ts
@@ -44,8 +44,7 @@ describe('E2E - Reference EntryPoint', () => {
   const deployEntryPoint = async (deployer: Signer, relayer: Signer) => {
     const { abi, bytecode } = EntryPointArtifact
     const transaction = await deployer.sendTransaction({ data: bytecode })
-    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-    const receipt = (await transaction.wait())!
+    const receipt = await transaction.wait()
     const contractAddress = receipt.contractAddress
     if (contractAddress === null) {
       throw new Error(`contract deployment transaction ${transaction.hash} missing address`)
@@ -81,8 +80,7 @@ describe('E2E - Reference EntryPoint', () => {
     )
 
     const transaction = await logGas('Execute UserOps with reference EntryPoint', entryPoint.handleOps(userOps, await relayer.getAddress()))
-    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-    const receipt = (await transaction.wait())!
+    const receipt = await transaction.wait()
 
     const transfers = ethers.parseEther('0.1') * BigInt(userOps.length)
     const deposits = receipt.logs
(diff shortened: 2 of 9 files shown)


4337/src/utils/safe.ts
```diff
@@ -88,6 +88,7 @@ const actionCalldata = (action: MetaTransaction): string => {
 }
 
 export interface RpcProvider extends Provider {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
   send(method: string, params: unknown[]): Promise<any>
```
nlordell, line 92: This is the interface of a `JsonRpcProvider` in `ethers`, so I just disabled the lint for the `send` function.

4337/test/e2e/ReferenceEntryPoint.spec.ts
```diff
@@ -44,8 +44,13 @@ describe('E2E - Reference EntryPoint', () => {
   const deployEntryPoint = async (deployer: Signer, relayer: Signer) => {
     const { abi, bytecode } = EntryPointArtifact
     const transaction = await deployer.sendTransaction({ data: bytecode })
-    const receipt = await transaction.wait()
-    return new ethers.Contract(receipt!.contractAddress!, abi, relayer)
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    const receipt = (await transaction.wait())!
```
nlordell, line 48: I chose to disable `no-non-null-assertion` lints for all `transaction.wait()!` instances (there are more below as well). The rationale is that `receipt` will only be `null` if called with `wait(0)`, but not if called with `wait(n)` where `n > 0` or `wait()` (which defaults to `n = 1`). I consider this a typing issue in `ethers`, although I'm not familiar enough with TypeScript to know if it is possible or how to encode dependant types this way.

4337/test/eip4337/EIP4337ModuleExisting.spec.ts
```diff
@@ -10,6 +11,8 @@ import {
 } from '../../src/utils/userOp'
 import { chainId } from '../utils/encoding'
 
+type ParseLog = Parameters<Interface['parseLog']>[0]
```
nlordell, line 14: This is an unfortunate issue where `ethers.Interface.parseLog` expects a type that is not compatible with `ethers.Log` (but the data is compatible). I proposed a fix for it here https://github.com/ethers-io/ethers.js/pull/4459.

4337/src/utils/safe.ts
```diff
@@ -98,6 +99,7 @@ export class MultiProvider4337 extends JsonRpcProvider {
     this.generalProvider = generalProvider
   }
 
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
mmv08, line 102: Any is just a warning, right? Is it necessary to turn it off, or keeping it as a "todo" list is better?

4337/test/e2e/ReferenceEntryPoint.spec.ts
mmv08: It's possible with [function overloads](https://www.typescriptlang.org/docs/handbook/2/functions.html#function-overloads) but it has to be done on the ethers' side

mmv08 (approved):

4337/src/utils/safe.ts
nlordell: In this case, I don't think it is a `TODO` lint - as this is the function signature required by `ethers`, so it will always be the `any` and the lint is a false-positive.

That being said, I do agree with "keeping it as a "todo" list is better", so at least for the `eslint-disable-next-line @typescript-eslint/no-non-null-assertion` it would make sense to keep the lint warning around...

4337/test/e2e/ReferenceEntryPoint.spec.ts
nlordell: Thanks so much for the link... I went down a rabbit hole with this and found a better solution. In fact, you can augment the `ethers` module with additional overloads for these methods 🤯.

Anyway, pushed a new version that does not require the `eslint-disable-next-line` and gets rid of the `ParseLog` type and casting from 👇.

4337/test/eip4337/EIP4337ModuleExisting.spec.ts
nlordell: Fixed with module augmenting.

nlordell:
@mmv08 can you take another look? I augmented the `ethers` module to fix some of the typing issues I ran into.

mmv08:
> @mmv08 can you take another look? I augmented the `ethers` module to fix some of the typing issues I ran into.

In general, it looks good to me. I don't know how reliable overwriting ethers' types is in case there are any breaking changes. But anyway, if there are any, they should be caught by tests.

Issue #145, opened 2023-11-10
[4337] E2E Test with Reference Bundler


nlordell:
## Context / issue

Currently, while we have E2E tests that verify that the 4337 module with the Safe contracts work with the reference entry point contract, we do not have the same for the reference bundler implementation.

Implementing this would allow quicker iterations when exploring new features and integrations with 4337.

## Proposed solution

Migrate the `docker-compose.yaml` file from the https://github.com/5afe/playground-4337 repository and create a dedicated E2E test that verifies support.

## Additional Context

It should also be possible to extend the setup to additional open source bundlers (such as Alchemy `rundler`, Pimlico `alto-bundler`).

PR #151, opened 2023-11-14
Minor Cleanup to Project and Hardhat Configuration

nlordell:
In preparation for #145, I was making sure that the deployments work on Geth and found some small cleanups that I would like to do in a separate PR (just to reduce the noise from any followups).

Namely:
- Move safe contract dependency imports to their own file instead of `SafeMock.sol`.
- Use single solidity version (we had 3 before, but only 0.8.21 was actually ever used to build anything)
- Use SafeL2 in the E2E test for two reasons:
  1. it does slightly more things than `Safe` which is worth testing
  2. we explicitly depend on `SafeL2` in the new `SafeContracts.sol` file, while `Safe` gets pulled in “by chance”.
- Bump Solidity to the latest 0.8.23 version


4337/hardhat.config.ts
```diff
@@ -25,16 +25,33 @@ if (PK) {
...
   throw new Error(`Could not find Infura key in env, unable to connect to network ${argv.network}`)
 }
 
 import './src/tasks/local_verify'
 import './src/tasks/deploy_contracts'
 import './src/tasks/show_codesize'
 
-const primarySolidityVersion = SOLIDITY_VERSION || '0.8.21'
-const soliditySettings = !!SOLIDITY_SETTINGS ? JSON.parse(SOLIDITY_SETTINGS) : { optimizer: { enabled: true, runs: 10_000_000 } }
+const solidityVersion = SOLIDITY_VERSION || '0.8.21'
```
mmv08, line 36:
```suggestion
const solidityVersion = SOLIDITY_VERSION || '0.8.23'
```

mmv08: I think we can also pump the solc version :)

mmv08 (approved):

4337/hardhat.config.ts
nlordell: I would keep that change separate for a couple reasons:
- It means that we may potentially have the Safe4337Module end up on a different address that what is currently in the deployments
- This PR is intended to not change behaviour at all

nlordell: **Edit**: I think you are right… we are already developing for a 0.2 version anyway, so bumping makes sense.

mmv08: > I would keep that change separate for a couple reasons:

> It means that we may potentially have the Safe4337Module end up on a different address that what is currently in the deployments

Whoever wants to reproduce the deployment should use the release tag anyway :)

PR #154, opened 2023-11-14
Refactor Deployment Scripts

nlordell:
This PR refactors the deployments scripts to take advantage of deployment tagging and network tagging. Correct deployment order is guaranteed through the use of the `dependencies` feature.

First, it defines three network tags:
- `dev` used to tag “development” networks, such as public testnets (Sepolia for example) and `localhost`. This ensures all required contracts are deployed including the `EntryPoint` (which will land on the same address as the reference v0.6.0 version by the mystical magic of EVM `CREATE2`)
- `safe` used to additionally deploy Safe contracts (proxy factory and singleton)
- `test` to deploy all contracts needed used for `hardhat test`. In particular, this ensures that the `TestEntryPoint` contract is used for the entry point deployment instead of the reference `EntryPoint` contract

In addition, I added tags to all the different deployment groups to `hardhat deploy --tags ...` work for redeploying specific contracts instead of everything.

For example - when deploying on Görli:
```
$ npx hardhat deploy --network goerli
Nothing to compile
No need to generate any newer typings.
reusing "EntryPoint" at 0x5FF137D4b0FDCD49DcA30c7CF57E578a026d2789
deploying "AddModulesLib" (tx: 0xe060a563a7cf1796ae48356b0312697f0e78566c39c96063fb9f1e094a383258)...: deployed at 0x8EcD4ec46D4D2a6B64fE960B3D64e8B94B2234eb with 200879 gas
deploying "Safe4337Module" (tx: 0xd9e3cf3fee9c65ce44926be225f8a2030b1a9892b3ac5f31120f4ab174fe935e)...: deployed at 0x4E6d4819dB440578caDE32563DD263FD15c2F536 with 1794049 gas
deploying "HariWillibaldToken" (tx: 0x191a78dc2c5aafd37c1dd971655483a2aa17bbe65f3ff3931d8a751f328202b5)...: deployed at 0x650d14991F1Dc8079D750e81c600b0E8F800aF8D with 1202554 gas
```

Note that the `EntryPoint` contract gets “reused” and matches the official deployment (🎉) and only contracts that are needed for the `runOp` script actually get deployed.

And when deploying on `localhost`:

```
$ npx hardhat deploy --network localhost
Nothing to compile
No need to generate any newer typings.
sending eth to create2 contract deployer address (0x3fab184622dc19b6109349b94811493bf2a45362) (tx: 0x45fa394b1605577da486658f2dbceff49f75e39c97cd092fc80dd53eb1ec06c4)...
deploying create2 deployer contract (at 0x4e59b44847b379578588920ca78fbf26c0b4956c) using deterministic deployment (https://github.com/Arachnid/deterministic-deployment-proxy) (tx: 0xeddf9e61fb9d8f5111840daef55e5fde0041f5702856532cdbb5a02998033d26)...
deploying "EntryPoint" (tx: 0xe21b49e928e292e97f6dda4b32f16863e6829d40bc58fbd628a20516c0155afc)...: deployed at 0x5FF137D4b0FDCD49DcA30c7CF57E578a026d2789 with 5318436 gas
deploying "AddModulesLib" (tx: 0x6695728e0011bbb665cf042f9b7bda1155da105dc2461b536aeff19d326222e1)...: deployed at 0x8EcD4ec46D4D2a6B64fE960B3D64e8B94B2234eb with 200879 gas
deploying "Safe4337Module" (tx: 0x9cecf3aef34994659a57ddba39cc90bb24d51bfe5d594b6ac893925670aa1c3e)...: deployed at 0x4E6d4819dB440578caDE32563DD263FD15c2F536 with 1794049 gas
deploying "SafeProxyFactory" (tx: 0x9dc11f5a40c631ab57ee15df0c2805883b96e472732936c9bc3938ea23eb5357)...: deployed at 0x68BAcEBAa1B6C8548ba76b2EFea8B83b1165C0D1 with 569183 gas
deploying "SafeL2" (tx: 0xd26105ec6fb0d7fdc307f26d7d25cd37f1f1162fb12d1879b13a498619624b29)...: deployed at 0x02E89c18a598f587515fc452c1154E2bcC5Be1Ef with 3792060 gas
deploying "HariWillibaldToken" (tx: 0xf3bf64af88363f5230dd67bbf1e956a01b959bcb0fbd25336e73372821cc4d86)...: deployed at 0xF59F15ee0456Ddd6DF312f23A518b8C457B0349d with 1202542 gas
```

Note how, not only does the `CREATE2` deployer get auto-magically deployed, the `EntryPoint` lands on the exact same address as the official deployments (🧙) and Safe contracts also get deployed. This will allow `localhost` development nodes to be used for proper E2E tests for #145


4337/src/deploy/entrypoint.ts
```diff
@@ -0,0 +1,33 @@
+import EntryPoint from '@account-abstraction/contracts/artifacts/EntryPoint.json'
+import { DeployFunction } from 'hardhat-deploy/types'
+
+const ENTRY_POINT = process.env.DEPLOY_ENTRY_POINT
```
mmv08, line 4: Is it me, or the env var name is a little bit misleading? I'd expect it to be a boolean to deploy an entry point, but it seems like a fallback address if the entry point is not deployed.

nlordell: Note that this was the environment variable before this PR. I _think_ the rationale behind it was that all environment variables for the `runOp` script were prefixed with `SCRIPT_*` and all the environment variables for deployment were prefixed with `DEPLOY_*`, but agree that it is confusing and sounds like a true/false variable for “deploy entry point?”

Do you think `DEPLOYMENT_ENTRY_POINT` / `DEPLOYMENT_OVERRIDE_ENTRY_POINT` / `DEPLOYMENT_ENTRY_POINT_ADDRESS` would be clearer?

nlordell: Renamed to `DEPLOYMENT_ENTRY_POINT_ADDRESS`

mmv08: legend

mmv08 (approved):

PR #155, opened 2023-11-15
Implement E2E Test Using Reference Bundler

### LINKED ISSUE #145: [4337] E2E Test with Reference Bundler

nlordell:
Fixes #145

This PR adds an E2E test that interacts with the reference bundler implementation to ensure that the ERC-4337 module does not break any rules during user operation validation.

In addition, some documentation on how to run the E2E tests was added as well as a new CI step for executing the E2E tests.


.github/workflows/ci_4337.yml
```diff
@@ -14,14 +14,41 @@ jobs:
...
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 18.x
+          cache: npm
+          cache-dependency-path: 4337/package-lock.json
+      - run: npm ci
+      - run: |
+          docker compose up -d
+          # wait for containers to start up
+          SECONDS=0
```
mmv08, line 41: TIL https://unix.stackexchange.com/a/646605

4337/docker/bundler/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM docker.io/library/node:18
```
mmv08, line 1: The latest LTS is 20. Would it make sense to use the latest version? Or is it the bundler's requirement?

4337/src/utils/safe.ts
```diff
@@ -116,6 +115,10 @@ export class MultiProvider4337 extends JsonRpcProvider {
       return this.generalProvider.send(method, params)
     }
   }
+
+  public async sendUserOperation(userOp: UserOperation, entryPoint: AddressLike): Promise<string> {
+    return await super.send('eth_sendUserOperation', [userOp, await ethers.resolveAddress(entryPoint, this)])
```
mmv08, line 120: even resolving ens names here. attention to detail 🥇

.github/workflows/ci_4337.yml
mmv08: It doesn't seem like the timeout works, though. It took longer than 30 seconds for cold boot on my local machine, and according to the CI logs, it also takes more than 30 seconds.

4337/docker/bundler/Dockerfile
nlordell: It is what they use in their CI: https://github.com/eth-infinitism/bundler/blob/436f12241b06397989fd7a522af0deca0d46ac09/.github/workflows/build.yml#L22

.github/workflows/ci_4337.yml
nlordell: This timeout only starts after the `docker componse up -d` command ends (I reset the counter to 0 here).

Building the docker image usually is what takes the longest (~1minute), but it isn’t counted in the timeout.

mmv08: Ahh it's for the start only, sorry I misunderstood it

mmv08 (approved):

4337/docker/bundler/Dockerfile
mmv08: > Node.js 18 will be supported until April 2025

Ok we have a plenty of time

Issue #143, opened 2023-11-10
[4337] Add Support for `validUntil` and `validAfter`


nlordell:
## Context / issue

In its current version, the 4337 module only supporting validating user operations with no validity associated with them (they are valid for all timestamps). There are use cases where user operations should only be valid for a specific time frame, and the 4337 module should be enhanced to support those.

## Proposed solution

Encode the `validUntil` and `validAfter` values in the signature bytes.

nlordell:
Implemented in #156

PR #156, opened 2023-11-15
Feat: Support `validAfter` and `validUntil` timestamps 

mmv08:
This PR:
- Implements https://github.com/safe-global/safe-modules/issues/143 by prepending a uint96 that contains two 48-bit timestamps, `validAfter` and `validUntil` to the signature bytes
- I added an e2e with the reference entrypoint that ensures that the timestamps are correctly bubbled up 
- packing two uint48s into uint96 was @nlordell's suggestion to overcome the `stack too deep` error


4337/test/eip4337/EIP4337ModuleExisting.spec.ts
```diff
@@ -27,28 +22,7 @@ describe('Safe4337Module - Existing Safe', () => {
     }
   })
 
-  describe('getOperationHash', () => {
```
mmv08, line 30: Moved this to the general `EIP4337Module` test suite, IMO test suite for existing safes wasn't the right place

4337/contracts/Safe4337Module.sol
```diff
@@ -84,7 +88,7 @@ contract Safe4337Module is IAccount, HandlerContext, CompatibilityFallbackHandle
      * @param data Data payload of the user operation.
      * @param operation Operation type of the user operation.
      */
-    function executeUserOp(address to, uint256 value, bytes memory data, uint8 operation) external {
+    function executeUserOp(address to, uint256 value, bytes calldata data, uint8 operation) external {
```
mmv08, line 91: i removed allocating memory for data from the execution functions, saved ~200 gas

nlordell: Good catch, and makes sense - it would have to copy the `memory` bytes when encoding the Safe `execTransactionFromModuleReturnData` (which I _think_ currently involves a loop since there is no `MCOPY` instruction) instead of just doing a `CALLDATACOPY`.

4337/contracts/Safe4337Module.sol
```diff
@@ -175,13 +182,17 @@ contract Safe4337Module is IAccount, HandlerContext, CompatibilityFallbackHandle
...
-        try ISafe(payable(userOp.sender)).checkSignatures(operationHash, "", userOp.signature) {
-            validationData = SIG_VALIDATION_SUCCESS;
+        // The timestamps are validated by the entry point, therefore we will not check them again
+        uint48 validAfter = uint48(signatureTimestamps >> 48);
+        uint48 validUntil = uint48(signatureTimestamps);
+        try ISafe(payable(userOp.sender)).checkSignatures(operationHash, "", signatures) {
+            validationData = _packValidationData(false, validUntil, validAfter);
         } catch {
-            validationData = SIG_VALIDATION_FAILED;
+            validationData = _packValidationData(true, validUntil, validAfter);
```
nlordell, line 195: I would not return the `validUntil, validAfter` in the signature error here for two reasons:
1. The signature is incorrect, so you don’t actually know what the real validity is
2. An invalid signature is invalid always, so the validity doesn’t really conceptually make sense IMO.

I think both are handled identically by the `EntryPoint` contract, so there is no real reason to pay the (minor) additional gas overhead of computing the shared signature.

4337/contracts/Safe4337Module.sol
```diff
@@ -175,13 +182,17 @@ contract Safe4337Module is IAccount, HandlerContext, CompatibilityFallbackHandle
...
             entryPoint
         );
 
-        try ISafe(payable(userOp.sender)).checkSignatures(operationHash, "", userOp.signature) {
-            validationData = SIG_VALIDATION_SUCCESS;
+        // The timestamps are validated by the entry point, therefore we will not check them again
+        uint48 validAfter = uint48(signatureTimestamps >> 48);
+        uint48 validUntil = uint48(signatureTimestamps);
+        try ISafe(payable(userOp.sender)).checkSignatures(operationHash, "", signatures) {
+            validationData = _packValidationData(false, validUntil, validAfter);
```
nlordell, line 193: If we kept the same ordering as in ERC-4337, then here we could just return:

```solidity
validationData = uint256(signatureTimestamps);
```

4337/contracts/Safe4337Module.sol
```diff
@@ -166,6 +172,7 @@ contract Safe4337Module is IAccount, HandlerContext, CompatibilityFallbackHandle
      * @return validationData An integer indicating the result of the validation.
      */
     function _validateSignatures(address entryPoint, UserOperation calldata userOp) internal view returns (uint256 validationData) {
+        (uint96 signatureTimestamps, bytes memory signatures) = abi.decode(userOp.signature, (uint96, bytes));
```
nlordell, line 175: Sorry to be annoying here 🙈.

I think it might be worth having a discussion on what we want to use for the signature format for ERC-4337 signatures.

Personally, I do agree with the choice, but I would argue it is not obvious:
- Packed encoding? It might save some gas (might be worth experimenting at least)
- More complicated encoding (similar to the 4-byte selector scheme that we had for the `SignatureManager` in the protocol, where if the signature starts with `bytes4(keccak(“SignatureWithValidity(uint96,bytes)”))` then we parse it this way, otherwise we use `signatureTimestamps = 0` and the signature bytes be the signature bytes).

One benefit of manual decoding of the signature, is that we can pass `memory signatures` to the `getOperationHash` function, but keep it `bytes calldata` for the `checkSignatures(operationHash, "", signatures)` call to avoid a loop for copying memory (not 100% sure this will have an impact, but I assume it is similar to the gas optimization you made in the execution function).

nlordell (commented):
Interesting choice to change the order of the `valid*` in how it is packed. While it may make sense from a natural language perspective, I think it might go against developer expectations if they look at how the validation data is typically computed in ERC-4337.

4337/contracts/Safe4337Module.sol
mmv08: The reason why I decided to return it in both cases is:
1. This is how it was implemented in the reference contracts, and also, the reference function has a boolean for signature failure for some reason.
2. They may matter for simulation (erc4337 simulation is a black box, but for example, they don't want you to revert in `validateUserOp`). Looking at the simulation methods in the contract, the timestamps are being used: https://github.com/eth-infinitism/account-abstraction/blob/674b1f51164e641a18d5c141895bab9c96a68e9d/contracts/core/EntryPointSimulations.sol#L87-L123

4337/contracts/Safe4337Module.sol
mmv08: As discussed in the standup, I've implemented the packed encoding. We decided not to implement the 4-byte selector as it would be very complicated.

4337/package.json
```diff
@@ -15,20 +15,20 @@
...
+    "lint": "npm run lint:sol && npm run lint:ts",
     "lint:sol": "solhint 'contracts/**/*.sol'",
     "lint:ts": "eslint ./src && eslint ./test",
     "fmt": "prettier --ignore-path .gitignore --write . && prettier --write 'contracts/**/*.sol'",
     "fmt:check": "prettier --ignore-path .gitignore --check . && prettier --check 'contracts/**/*.sol'",
-    "prepack": "yarn build",
+    "prepack": "npm run build",
     "prepare": "cd .. && husky install 4337/.husky",
-    "prepublish": "yarn rimraf build && yarn build && yarn build:ts"
+    "prepublish": "npx rimraf build && npm run build && npm run build:ts"
```
mmv08, line 31: Funny thing: today, I worked on a fresh computer that didn't have `yarn` installed, only `npm`. All the scripts failed :D Although we use npm

4337/package-lock.json
```diff
@@ -160,9 +160,9 @@
       }
     },
     "node_modules/@eslint/eslintrc": {
-      "version": "2.1.2",
```
nlordell, line 163: Are the version bumps intended?

4337/src/utils/userOp.ts
```diff
@@ -187,3 +194,36 @@ export const getSupportedEntryPoints = async (provider: ethers.JsonRpcProvider):
   console.log({ supportedEntryPoints })
   return supportedEntryPoints.map(ethers.getAddress)
 }
+
+export const encodeSignatureTimestamp = (validUntil: BigNumberish, validAfter: BigNumberish) => {
+  // Ensure that the values don't exceed 48 bits
+  if (BigInt(validAfter) > 0xffffffffffff || BigInt(validUntil) > 0xffffffffffff) {
```
nlordell, line 200: Doesn’t this need to be 
```suggestion
  if (BigInt(validAfter) > 0xffffffffffffn || BigInt(validUntil) > 0xffffffffffffn) {
```

nlordell (commented):
Code looks good. Just think that the version bumps are slightly unrelated to the changes here.

4337/package.json
nlordell: 🤣

4337/src/utils/userOp.ts
mmv08: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt

> The boolean-returning operators allow mixing numbers and BigInts as operands:
> 
> [Relational operators](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators#relational_operators) and [equality operators](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators#equality_operators): >, <, >=, <=, ==, !=, ===, !==
> [Logical operators](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators#binary_logical_operators) only rely on the [truthiness](https://developer.mozilla.org/en-US/docs/Glossary/Truthy) of operands

4337/package-lock.json
mmv08: Jein. I was dealing with some bizzare issues, which were fixed by `rm -rf node_modules && rm -rf package-lock.json && npm i`

4337/src/utils/userOp.ts
nlordell: I’m in the lucky 10000 today! https://xkcd.com/1053/

nlordell (approved):

4337/src/utils/userOp.ts
mmv08: Honestly, it was a copilot-generated code and I didn't know that myself.

PR #158, opened 2023-11-16
Created `onlySupportedEntryPoint(...)` modifier avoiding code redundancy

### LINKED ISSUE #157: [4337] Using `modifier` for checking `SUPPORTED_ENTRYPOINT`
## Description

The EntryPoint is checked in the validation and execution of user operation in 4337 which is done as below:

```
address entryPoint = _msgSender();
require(entryPoint == SUPPORTED_ENTRYPOINT, "Unsupported entry point");
```

The above code lines are duplicated in three functions (`validateUserOp(...)`, `executeUserOp(...)` and `executeUserOpWithErrorString(...)`) which can be abstracted away with a `modifier`.

## Proposed Solution

Using a modifier which does the same check and integrating it with the above-mentioned functions for better readable code.

remedcu:
Closes #157

This PR removes the redundant code to check if the entry point contract is actually supported or not.

### DIFF
--- a/4337/contracts/Safe4337Module.sol
+++ b/4337/contracts/Safe4337Module.sol
@@ -45,10 +45,18 @@ contract Safe4337Module is IAccount, HandlerContext, CompatibilityFallbackHandle
     }
 
     /**
+     * @notice Validates the call is initiated by the entry point.
+     */
+    modifier onlySupportedEntryPoint() {
+        require(_msgSender() == SUPPORTED_ENTRYPOINT, "Unsupported entry point");
+        _;
+    }
+
+    /**
      * @notice Validates a user operation provided by the entry point.
      * @inheritdoc IAccount
      */
-    function validateUserOp(UserOperation calldata userOp, bytes32, uint256 missingAccountFunds) external returns (uint256 validationData) {
+    function validateUserOp(UserOperation calldata userOp, bytes32, uint256 missingAccountFunds) external onlySupportedEntryPoint returns (uint256 validationData) {
         address payable safeAddress = payable(userOp.sender);
         // The entry point address is appended to the calldata in `HandlerContext` contract
         // Because of this, the relayer may manipulate the entry point address, therefore we have to verify that
@@ -63,7 +71,6 @@ contract Safe4337Module is IAccount, HandlerContext, CompatibilityFallbackHandle
         );
 
         address entryPoint = _msgSender();
-        require(entryPoint == SUPPORTED_ENTRYPOINT, "Unsupported entry point");
 
         // The userOp nonce is validated in the entry point (for 0.6.0+), therefore we will not check it again
         validationData = _validateSignatures(entryPoint, userOp);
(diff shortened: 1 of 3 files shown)


remedcu:
I have read the CLA Document and I hereby sign the CLA

mmv08 (commented):
I don't know if it's worth checking this for a modifier used only thrice, but I'm curious about the impact on the code size? (we have a code size hardhat task)

remedcu:
> I don't know if it's worth checking this for a modifier used only thrice, but I'm curious about the impact on the code size? (we have a code size hardhat task)

Before:
```
Safe4337Module 8100 bytes (limit is 24576)
```

After:
```
Safe4337Module 8135 bytes (limit is 24576)
```

**35 bytes higher.**

I checked by changing the `modifier` to an `internal` function. And then the code size was `7767 bytes`, **around 300+ bytes smaller**. Should I make the change?

mmv08:
> I checked by changing the modifier to an internal function. And then the code size was 7767 bytes, around 300+ bytes smaller. Should I make the change?

I'd say so. You could also keep the modifier and call that internal function in the modifier.

mmv08:
> > call that internal function in the modifier
> 
> The downside is slightly higher runtime cost. Personally, I would trade improving runtime cost for codesize here, because this contract is a “deploy once, use 1000000000 times” kind of contract.

Does this apply to the internal function + modifier or standalone internal function as well?

remedcu:
Based on checking with the demo code, it does apply to both using an internal function and an internal + modifier.

Source Demo Code: https://pastebin.com/M2qdK91m

mmv08:
> Based on checking with the demo code, it does apply to both using an internal function and an internal + modifier.
> 
> Source Demo Code: https://pastebin.com/M2qdK91m

Nice, thanks for the analysis

4337/test/eip4337/EIP4337Module.spec.ts
```diff
@@ -81,7 +81,7 @@ describe('Safe4337Module', () => {
         signature,
       })
 
-      await expect(safeModule.validateUserOp(userOp, ethers.ZeroHash, 0)).to.be.revertedWith('Unsupported execution function id')
+      await expect(entryPoint.executeUserOp(userOp, 0)).to.be.revertedWith('Unsupported execution function id')
```
mmv08, line 84: Why was this changed?

remedcu: As the modifier gets executed first, this will lead to getting the error `Unsupported entry point` instead of `Unsupported execution function id`.

4337/contracts/Safe4337Module.sol
```diff
@@ -63,7 +75,6 @@ contract Safe4337Module is IAccount, HandlerContext, CompatibilityFallbackHandle
         );
 
         address entryPoint = _msgSender();
```
nlordell, line 77: ubernit: Can you move this down into the `if` block where it is used?

remedcu: Actually, it is used before the `if` block for `validationData`.

nlordell (commented):
Code changes look good in general, just have the same question as @mmv08

4337/test/eip4337/EIP4337Module.spec.ts
nlordell: Another way to change this while keeping the test calling `validateUserOp` would be to call the `validateUserOp` function from the `entryPoint` contract.

I would slightly prefer that over calling `executeUserOp` which only implicitely calls `validateUserOp`

4337/contracts/Safe4337Module.sol
nlordell: 🤦 oops, sorry.

4337/test/eip4337/EIP4337Module.spec.ts
remedcu: Something like this will do the trick @nlordell 

```solidity=
const entryPointAddress = await ethers.getSigner(await entryPoint.getAddress())
const safeFromEntryPoint = safeModule.connect(entryPointAddress)
await expect(safeFromEntryPoint.validateUserOp.staticCall(userOp, ethers.ZeroHash, 0)).to.be.revertedWith('Unsupported execution function id')
```

mmv08 (approved):
👏

Issue #149, opened 2023-11-13
Launchpad Account Deployer to Work Around 4337 Limitations


nlordell:

nlordell:
> Out of curiosity, have you looked into how other projects solve this?

Not yet.

PR #159, opened 2023-11-17
[4337] User Operation with Custom Singleton Signers

### LINKED ISSUE #148: [4337] Evaluate Singleton Signers + Staked Factory
## Context / issue

The 4337 module currently allows creating a new account (using `UserOperation.initCode`) and executing an action atomically.

Furthermore, support for additional signing schemes (such as WebAuthn using P256 curve signatures) is already supported by the Safe by creating a dedicated signer contract per owner (in the case of WebAuthn, one contract per public key) implementing the signature validation logic using ERC-1271.

However, ERC-4337 limits the number of `CREATE2` calls to 1 when using `initCode` meaning that for example, it is currently not possible to deploy a Safe using account abstraction with nothing but a WebAuthn signer.

## Proposed solution

Implement a singleton `isValidSignature` implementation that can be shared across multiple Safes. You can, for example, deploy sequential `new Signer{salt: n}()` for sequential `n` to allow multiple owners that use this signing strategy. Each signer would keep a `mapping(safe => data)` for each Safe's data required for signature verification.

The downside is that this would still require a staked factory (because the `Signer` would access associated storage).

## Alternatives

- #147 
- #149 

## Outcome

A script similar to `runOp` where an owner can be deployed with the account as part of the single bundle.

The focus is **not** to have a useful signer contract, just a process that is supported under 4337.

nlordell:
Fixes #148 

This PR introduces an E2E test using the official bundler where we deploy a Safe that uses a custom (and silly 🤪) signing scheme over ERC-1271. Specifically, the signing scheme is deployed as a singleton contract that can be shared by multiple Safes using a `mapping(safe => data)` to differentiate between signers. In theory, this could be adapted to, for example, checking a signature using P256 curve in order to integrate with WebAuthn, and is meant to “prove” that the Safe with the ERC-4337 module would work with existing WebAuthn signer implementations.

The actual implementation is a little convoluted for a few reasons. Namely, ERC-4337 only allows a **single** `CREATE2` per user operation that **MUST** result in the `userOp.sender` account. Because of this, deploying ad-hoc signers in the `Safe.setup` call is not possible, and we have instead opted for a “singleton signer” pattern, where the logic for verifying signatures is shared and a mapping is used to store per-Safe context relevant to signature verification.

Additionally, In order for this to work with ERC-4337, a staked factory is needed (as accessing associated storage for user operations with `initCode` require a factory). Unfortunately, staked factories add all kinds of complexities:
- Currently, the staked factory contract is quite dumb (can only stake, but can’t unstake). A full staked factory contract requires us to iron out some requirements and make a proper implementation with documentation and tests (out of scope for this PR and issue, and a non-trivial amount of work)
- Staking is determined per bundler, so while some amount of stake may be sufficient for a bundler, it may not be sufficient for all (which adds additional off-chain management dependencies).

**For these reasons, I would recommend exploring #149**.


4337/contracts/test/TestSingletonSigner.sol
```diff
@@ -0,0 +1,42 @@
+// SPDX-License-Identifier: LGPL-3.0-only
+/* solhint-disable one-contract-per-file */
+pragma solidity >=0.8.0;
+
+import {ISignatureValidator} from "@safe-global/safe-contracts/contracts/interfaces/ISignatureValidator.sol";
+
+contract TestSingletonSigner is ISignatureValidator {
+    struct Key {
+        uint256 _dummy;
```
nlordell, line 9: This is just to verify this model would work for P256 signatures (which require 2 `uint256` slots, one for the X and one for the Y coordinate of the public key - which is a point on the curve).

akshay-ap (approved):

PR #163, opened 2023-11-21
Fix Revert message propagation

nlordell:
This PR fixes a revert message propagation issue where the length of revert bytes was the **offset** of the return data in memory and not its actual length 🤦. It turns out that because of how ABI encoding works, the tests were not catching this issue since in that particular test the offset was alway larger than the actual length of the revert message. In theory, this is UB (as the offset it not guaranteed to be larger than the message, although, in practice it is not a serious issue, as it appears the compiler is allocating memory for the raw `call` result and then **copying with a loop** the ABI-decoded `returnData` memory bytes after it, so the offset of `returnData` will always be greater than its size, and the current implementation just returns too many bytes).

Adjusted tests to ensure the exact revert data is propagated.


mmv08 (approved):
I knew there was something wrong 🤦 Great catch!

PR #165, opened 2023-11-22
Send encoded Safe operation bytes to `checkSignatures` call

### LINKED ISSUE #160: [4337] Safe ERC-4337 Module Does not Support Contract Signatures
## Description

A Safe whose owners are smart contracts and use smart contract signature validation through the `ISignatureValidator` interface are not supported by the ERC-4337 module. This is because it always passes the `""` as data bytes to the `ISafe.checkSignatures` function.

https://github.com/safe-global/safe-modules/blob/a06f0cb6e10451840d76aa1d28b04a3e48ec246b/4337/contracts/Safe4337Module.sol#L197

## Additional context

Blocks #148 and #149

nlordell:
Fixes #160 

This PR fixes the `checkSignatures` call to the Safe to include both the Safe operation hash for recovery, but also the encoded operation bytes, which is required when using an ERC-1271 signer. In order to test this, the `SafeMock` now requires that `dataHash == keccak256(data)` and I added a unit test using a nested Safe set (note that this would require a staked paymaster to work with 4337 bundlers, but the test does check that the contract signatures works as expected).

### DIFF
--- a/4337/test/eip4337/ReferenceEntryPoint.spec.ts
+++ b/4337/test/eip4337/ReferenceEntryPoint.spec.ts
@@ -6,10 +6,11 @@ import EntryPointArtifact from '@account-abstraction/contracts/artifacts/EntryPo
 import { getFactory, getAddModulesLib } from '../utils/setup'
 import { buildSignatureBytes, logGas } from '../../src/utils/execution'
 import {
-  buildUserOperationFromSafeUserOperation,
   buildSafeUserOpTransaction,
-  signSafeOp,
+  buildUserOperationFromSafeUserOperation,
+  calculateSafeOperationData,
   encodeSignatureTimestamp,
+  signSafeOp,
 } from '../../src/utils/userOp'
 import { chainId } from '../utils/encoding'
 import { Safe4337 } from '../../src/utils/safe'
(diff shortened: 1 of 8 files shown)


4337/test/eip4337/ReferenceEntryPoint.spec.ts
```diff
@@ -152,6 +155,75 @@ describe('Safe4337Module - Reference EntryPoint', () => {
     expect(await ethers.provider.getBalance(safe.address)).to.be.eq(accountBalance - transfers - deposits)
   })
 
+  it('should support a Safe signer (NOTE: would require a staked paymaster for ERC-4337)', async () => {
```
mmv08, line 158: is that because the signer would access its storage and that goes against the rules?

mmv08 (approved):

mmv08:
Have just seen the public data commit. I think it's fine to leave it as internal, because it can be computed offchain

mmv08:
> > I think it's fine to leave it as internal, because it can be computed offchain
> 
> Reverted. I was unsure about the change - but decided to propose it anyway just in case. The rationale was that it would be useful for computing `SafeMessage` hashes (where the `bytes` preimage is used instead of the `bytes32` value).

cant wait for 1.5.0 to get released 😄

PR #166, opened 2023-11-22
[4337] Feature: Upgrade dependencies

mmv08:
This PR upgrades dependencies to the latest versions, even with a few major upgrades, no changes were required


4337/package.json
```diff
@@ -46,31 +46,31 @@
...
     "@openzeppelin/contracts": "^5.0.0",
-    "@types/chai": "^4.3.1",
-    "@types/mocha": "^9.1.1",
-    "@types/node": "^18.18.6",
-    "@types/yargs": "^17.0.10",
-    "@typescript-eslint/eslint-plugin": "^5.28.0",
-    "@typescript-eslint/parser": "^5.28.0",
+    "@types/chai": "^4.3.11",
+    "@types/mocha": "^10.0.6",
+    "@types/node": "^20.9.4",
```
nlordell, line 53: We should update the CI to run on Node 20 instead of 18 then (so the types version matches what we actually use):

https://github.com/safe-global/safe-modules/blob/7cfb9ddeb5f4c741987a360477928d1c581f22cd/.github/workflows/ci_4337.yml#L14

nlordell (approved):

PR #167, opened 2023-11-22
[4337] Separate Signature Fields for `validAfter` and `validUntil`

nlordell:
Follow up to #143 and #156

This PR refactors signature validation to use the immutable `SUPPORTED_ENTRYPOINT` value instead of passing it into various functions. This has the benefit of freeing up a stack slot for an additional variable. This means that we can now sign `validAfter` and `validUntil` as separate fields in our `SafeOp` struct (which should provide a nicer user experience for signing and be less cryptic than using a “signatureTimestamps” field that combines both values).

The caveat is that `getOperationHash` is tied to the module’s supported `EntryPoint` value, but IMO this makes more sense that what we had before, as signatures for other `EntryPoint`s are useless to it anyway.

Additionally, in theory, we don’t need to include the `entryPoint` in the signature, since it is implied by the `verifyingContract` address of the ERC-4337 module, seeing as `SUPPORTED_ENTRYPOINT` is an immutable. However, I decided to leave it in, the rationale being that it adds additional information to the user signature and improves UX around collecting EIP-712 signatures.

I added validity timestamps to the user operation in one of the E2E tests for better coverage.

### DIFF
--- a/4337/contracts/test/SafeMock.sol
+++ b/4337/contracts/test/SafeMock.sol
@@ -3,7 +3,6 @@
 pragma solidity >=0.8.0;
 
 import {IAccount} from "@account-abstraction/contracts/interfaces/IAccount.sol";
-import {INonceManager} from "@account-abstraction/contracts/interfaces/INonceManager.sol";
 import {UserOperation, UserOperationLib} from "@account-abstraction/contracts/interfaces/UserOperation.sol";
 
 contract SafeMock {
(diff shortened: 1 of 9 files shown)


4337/contracts/test/SafeMock.sol
```diff
@@ -123,21 +122,17 @@ contract Safe4337Mock is SafeMock, IAccount {
         bytes32,
         uint256 missingAccountFunds
     ) external onlySupportedEntryPoint returns (uint256) {
-        address entryPoint = msg.sender;
-
-        _validateReplayProtection(userOp);
-
```
nlordell, line 129: We don’t do replay protection checks in the module, so I removed this to compare apples to apples.

mmv08 (approved):
genius 👏

PR #176, opened 2023-11-29
Check paymaster and data as well

nlordell:
This PR includes the paymaster and data in the `SafeOp` signature. This is still a draft since this changes the signature scheme, so it would imply that signatures would be different, in which case we can include the `valid*` fields already.


nlordell:
Closing in favour of #177

PR #177, opened 2023-11-29
Sign Full User Operation Data

nlordell:
This PR adjusts the Safe module to sign **all** of the `UserOperation` struct data instead of the reduced set of fields it was originally signing. Namely, this means that the `SafeOp` EIP-712 struct needs to also include the fields:

* `initCode`: ensures that initialisation works exactly as the user specified
* `paymasterAndData`: ensures that user operations are tied to specific paymaster calls

Unfortunately, this required some non-trivial changes to the module code in order to accommodate all of these fields. Namely:
- `SafeOp` field order was changed to match that of the `UserOperation`. This was done for two reasons:
  1. Make it clearer and easier to verify that all `UserOperation` fields are accounted for
  2. Allows for more efficient encoding of `SafeOp` for computing the signing message when copying from a `UserOperation calldata`. While we don’t make use of this in the code, I believe it is a nice property to have.
- `getOperationHash` now takes a `UserOperation calldata` instead of all user operation fields (because of the Solidity “stack too deep” limitation).

https://github.com/safe-global/safe-modules/pull/177/commits/b3e1785548477343f6c2a9f048585d6bf20d6100 contains other versions of possible implementations for computing the signing message for the Safe operation, namely:
1. Using assembly and `calldatacopy` to setup the `SafeOp` struct in memory for hashing, this is the most efficient by around 310 gas compared to the current implementation, but has a lot of magic and is not easy to understand, hence it was not used. ([code](https://github.com/safe-global/safe-modules/blob/b3e1785548477343f6c2a9f048585d6bf20d6100/4337/contracts/Safe4337Module.sol#L168-L214))
2. Encoding values into a `uint256[14] memory`, while this requires slightly less boiler plate, it is not as easy to understand (IMO) and was less gas-efficient. ([code](https://github.com/safe-global/safe-modules/blob/b3e1785548477343f6c2a9f048585d6bf20d6100/4337/contracts/Safe4337Module.sol#L217-L262))
3. Using `abi.encode` as usual, but using some careful tricks to avoid “stack too deep” errors. Ironically, this was the **least** gas efficient of all the methods, and when building with instrumentation for generating a coverage report, would run into “stack too deep” errors anyway. ([code](https://github.com/safe-global/safe-modules/blob/b3e1785548477343f6c2a9f048585d6bf20d6100/4337/contracts/Safe4337Module.sol#L264-L302))


4337/contracts/Safe4337Module.sol
```diff
@@ -199,62 +158,147 @@ contract Safe4337Module is IAccount, HandlerContext, CompatibilityFallbackHandle
...
+                bytes1(0x19),
+                bytes1(0x01),
+                domainSeparator(),
+                structHash
+            );
+        }
+    }
+
+    /*
+    function _getSafeOpLessAssembly(
```
nlordell, line 217: This is an alternative implementation that consumes ~600 - ~1000 more gas per user operation recovery. Is is the “code” version of what we are trying to accomplish with the assembly ☝️.

4337/contracts/Safe4337Module.sol
```diff
@@ -199,62 +182,55 @@ contract Safe4337Module is IAccount, HandlerContext, CompatibilityFallbackHandle
...
+            signatures = sig[12:];
+        }
+
+        // It is important that **all** user operation fields are represented in the `SafeOp` data somehow, to prevent
+        // user operations from being submitted that do not fully respect the user preferences. The only exception is
+        // the `signature` bytes. Note that even `initCode` needs to be represented in the operation data, otherwise
+        // it can be replaced with a more expensive initialization that would charge the user additional fees.
+        {
+            // In order to work around Solidity "stack too deep" errors related to too many stack variables, manually
+            // encode the `SafeOp` fields into a memory `struct` for computing the EIP-712 struct-hash.
```
rmeissner, line 210: Maybe it is worth mentioning that this works, because a struct that has no dynamic fields is represented as continuous memory in solidity. Therefore we can just pass it directly to keccak

4337/contracts/Safe4337Module.sol
```diff
@@ -199,62 +182,55 @@ contract Safe4337Module is IAccount, HandlerContext, CompatibilityFallbackHandle
...
+                maxPriorityFeePerGas: userOp.maxPriorityFeePerGas,
+                paymasterAndDataHash: keccak256(userOp.paymasterAndData),
+                validAfter: validAfter,
+                validUntil: validUntil,
+                entryPoint: SUPPORTED_ENTRYPOINT
+            });
+            bytes32 structHash;
+            // solhint-disable-next-line no-inline-assembly
+            assembly ("memory-safe") {
+                structHash := keccak256(structFields, 448)
```
rmeissner, line 230: would it work if we use solidity only? i.e. `keccak256(abi.encodePacked(structFields))`

Not saying that we should do this, but was wondering if it would work without assembly

rmeissner (commented):
The solidity implementations looks good. 

I could not find test that check that initCode and paymasterData can not be changed.

4337/contracts/Safe4337Module.sol
```diff
@@ -20,28 +20,56 @@ import {ISafe} from "./interfaces/Safe.sol";
  *      - Replay protection is handled by the entry point.
  */
 contract Safe4337Module is IAccount, HandlerContext, CompatibilityFallbackHandler {
+    /**
+     * @dev A structure used internally for computing the EIP-712 signing message for a Safe operation.
+     */
+    struct SafeOpFields {
```
rmeissner, line 26: nitpick: I know style wise not optimal, but might be easier for varification if this is defined right after the `SAFE_OP_TYPEHASH` constant

4337/contracts/Safe4337Module.sol
```diff
@@ -20,28 +20,56 @@ import {ISafe} from "./interfaces/Safe.sol";
...
+        bytes32 paymasterAndDataHash;
+        uint48 validAfter;
+        uint48 validUntil;
+        address entryPoint;
+    }
+
+    /**
+     * @notice The EIP-712 type-hash for the domain separator used for verifying Safe operation signatures.
+     */
     bytes32 private constant DOMAIN_SEPARATOR_TYPEHASH = keccak256("EIP712Domain(uint256 chainId,address verifyingContract)");
```
rmeissner, line 46: Would it be more gas efficient if we hardcode the hash? (might be that the optimizer handles this)

4337/contracts/Safe4337Module.sol
rmeissner, line 230: If we keep the assembly it would be nice to comment on the length (i.e. `the SafeOpFields struct has 14 static fields which each have the size of 32 bytes. Therefore the length of the struct in memory is 14*32=448 bytes`)

4337/contracts/Safe4337Module.sol
rmeissner, line 26: Minor nitpick: I feel that a name like `EncodedSafeOpType` or similar might be more descriptive ...

nlordell:
> I could not find test that check that initCode and paymasterData can not be changed.

The TS library changed to additionally sign `initCode` and `paymasterAndData`, so we implicitly check this with the current tests. We don’t explicitly check they cannot change (as that is a property of EIP-712) but I agree this would be a nice test to have (given that we now want to guarantee that **all** `UserOperation` fields are signed).

4337/contracts/Safe4337Module.sol
nlordell: It should be equivalent to `abi.encode(structFields)` because there are no dynamic types. I will double check.

nlordell: Confirmed that it works with `abi.encode(structFields)` for an additional ~900 gas. It's not so expensive if you think it is worth avoiding the assembly.

4337/contracts/Safe4337Module.sol
nlordell: I ended up calling it `EncodedSafeOpStruct` as this is used in [the `hashStruct`](https://eips.ethereum.org/EIPS/eip-712#definition-of-hashstruct) algorithm (while [`encodeType`](https://eips.ethereum.org/EIPS/eip-712#definition-of-encodetype) already has a special meaning - encoding the type definition used in computing the type hash).

nlordell: ca530e28a16a649a35906a8bd7ed20889e351065

4337/contracts/Safe4337Module.sol
nlordell: ca530e28a16a649a35906a8bd7ed20889e351065

4337/contracts/Safe4337Module.sol
nlordell: Let me check. I think this is a separate change, so I will keep it out of this PR for now, but do a follow-up in case it helps.

nlordell: WOW - I am flabbergasted that it actually makes a difference… It is small (26 gas) but still…

nlordell:
@rmeissner - added a new unit test in 467df98a52be80fe10434b92aee69f8bfc381ed5 that verifies that if any `UserOperation` field changes, then the operation hash will also change.

4337/contracts/Safe4337Module.sol
nlordell: Captured in https://github.com/safe-global/safe-modules/issues/179

rmeissner (approved):

PR #178, opened 2023-11-29
Update 4337 readme

Uxio0:


4337/README.md
```diff
@@ -147,9 +147,11 @@ npm test
 
 End-to-end tests with the reference bundler implementation are also provided. These automated tests verify that no user operation validation rules are broken by the implementation.
 
+Note: Geth is used as the RPC as the reference bundler implementation relies on Geth specific tracing APIs.
```
mmv08, line 150:
```suggestion
Note: Geth is used as the RPC because the reference bundler implementation relies on Geth-specific tracing APIs.
```
took me a few moments to understand because of the double as, I think this version reads easier

Uxio0: Agree, thanks señor

mmv08 (approved):

Uxio0:
I don't have permissions to merge in this repo either, one of you have to do it 😢

Issue #179, opened 2023-11-30
Use Hardcoded Constants


nlordell:
## Context / issue

It turns out that using hardcoded constants instead of `keccak256(“…”)` has implications on gas consumption 🤦.

For example changing:

https://github.com/safe-global/safe-modules/blob/75e24b90204fa26ed607e2d150ed5b5cc10b6870/4337/contracts/Safe4337Module.sol#L23

to:

```solidity
    bytes32 private constant DOMAIN_SEPARATOR_TYPEHASH = 0x47e79534a245952e8b16893a336b85a3d9ea9fa8c573f3d803afb92a79469218;
```

Has some marginal gas benefits (~13 gas).

## Proposed solution

Use hardcoded constants.

PR #184, opened 2023-12-06
Safe 4337 Signer Launchpad

nlordell:
Related to #149, follow-up to #182

This PR provides another approach to Safe initialization to accommodate for custom signers that are deployed along with the Safe. See #182 for more context.

This launchpad implementation is very specific to custom signers and introduces a new `IUniqueSignerFactory` interface that must be implemented to work with it. The name “unique” signer is not great, but named this way since for a given factory and some “signer data”, a `ISignatureVerifier` implementation will correspond to a **unique** address. In the example of P-256 and WebAuthn signers, this means a P-256 public key will correspond to a unique on-chain address with an `ISignatureVerifier` implementation.

This implementation works by tying the Safe address to the `signerFactory`, `signerData` and Safe `setup` parameters. This means a couple of things:
1. The Safe address is unique for a specific initial configuration (owner, singleton, fallback, modules, etc.); the same property holds for the existing `SafeProxyFactory`
2. Unlike #182, the Safe address **is not** tied to the first user operation (in fact, the user operation is signed with the same owner as the Safe will use!)

```mermaid
sequenceDiagram
    actor B as Bundler
    participant E as EntryPoint
    participant F as SafeProxyFactory
    participant P as SafeProxy (Account)
    participant L as SafeSignerLaunchpad
    participant S as Safe (Singleton)
    participant M as Safe4337Module
    participant U as IUniqueSignerFactory
    participant A as Signer
    actor T as Target

    B  ->> +E: handleOps([userOp])
    E  ->> +F: CALL(userOp.initCode[:20])
    F  ->>  P: CREATE2
    F  ->>  P: setup(initHash)
    P  ->>  L: setup(initHash)
    note over L: SSTORE(initHash)
    F -->> -E: ok

    E  ->> +P: validateUserOp(userOp)
    P  ->> +L: validateUserOp(userOp)
    note over L: require(getInitHash(userOp) == initHash)
    L  ->> +U: isValidSignatureForSigner(operationData, signature, signerData)
    U -->> -L: magic
    L -->> -P: validationData
    P -->> -E: validationData

    E  ->> +P: initializeThenUserOp<br>(singleton, signerFactory, signerData, ...setupParams, callData)
    P  ->>  L: initializeThenUserOp(…)
    note over L: SafeStorage.singleton = singleton
    L  ->> +U: createSigner(signerData)
    U  ->>  A: CREATE2
    U -->> -L: owner
    L  ->>  S: setup([owner], 1, ...setupParams)
    note over S: standard Safe setup
    L  ->>  P: DELEGATECALL(callData)
    P  ->>  S: DELEGATECALL(callData)
    S  ->>  M: execUserOp(…)
    M  ->>  P: executeFromModule(…)
    P  ->>  S: executeFromModule(…)
    S  ->>  T: Ether
    P -->> -E: ok
```

### Implementation Notes

Currently, the implementation only supports a single signer, but can be changed to support multiple (by turning `signerFactory` and `signerData` into arrays).

### Downsides

The downsides of this implementation over #182 is that it is a solution _very_ specific to deploying Safes with custom signing schemes that require additional contract deployments and not a general solution to complex Safe setups in the context of ERC-4337.


4337/contracts/test/SafeSignerLaunchpad.sol
```diff
@@ -0,0 +1,266 @@
...
+        bytes calldata signerData
+    ) external view returns (bytes4 magicValue);
+}
+
+/**
+ * @title SafeOpLaunchpad - A contract for Safe initialization with custom unique signers that would violate ERC-4337 factory rules.
+ * @dev The is intended to be set as a Safe proxy's implementation for ERC-4337 user operation that deploys the account.
+ */
+contract SafeSignerLaunchpad is IAccount, SafeStorage {
+    bytes32 private constant DOMAIN_SEPARATOR_TYPEHASH = keccak256("EIP712Domain(uint256 chainId,address verifyingContract)");
```
mmv08, line 48: Didn't you figure out that pre-computed constants are cheaper? 😄

4337/contracts/test/SafeSignerLaunchpad.sol
```diff
@@ -0,0 +1,266 @@
...
+
+    receive() external payable {}
+
+    function setup(bytes32 initHash, address to, bytes calldata preInit) external onlyProxy {
+        _setInitHash(initHash);
+        if (to != address(0)) {
+            (bool success, ) = to.delegatecall(preInit);
+            require(success, "Pre-initialization failed");
+        }
+    }
```
mmv08, line 105: Just to confirm I understand things correctly: this shall be called through the factory during the deployment?

mmv08: nit: i know that the arity is different but it's too similar to the safe's core `setup` function. I'd name it `preValidationSetup` to be more descriptive

4337/contracts/test/TestUniqueSigner.sol
```diff
@@ -3,6 +3,16 @@
 pragma solidity >=0.8.0;
 
 import {ISignatureValidator} from "@safe-global/safe-contracts/contracts/interfaces/ISignatureValidator.sol";
+import {IUniqueSignerFactory} from "./SafeSignerLaunchpad.sol";
+
+function checkSignature(bytes memory data, uint256 signature, uint256 key) pure returns (bytes4 magicValue) {
+    uint256 message = uint256(keccak256(data));
+
+    // A very silly signing scheme where the `message = signature ^ key`
```
mmv08, line 11: almost a one-time pad: https://en.wikipedia.org/wiki/One-time_pad 😆

mmv08 (approved):
I like the PR

4337/contracts/test/SafeSignerLaunchpad.sol
nlordell: Yes, [by around 13 gas units](https://github.com/safe-global/safe-modules/issues/179). Since this is a test contract, I just went for what was easier to code.

4337/contracts/test/SafeSignerLaunchpad.sol
nlordell: Good point - I will rename it.

4337/contracts/test/TestUniqueSigner.sol
nlordell: Except the key is public :clown_face:. :1st_place_medal: cryptography here we go!

4337/contracts/test/SafeSignerLaunchpad.sol
nlordell: Renamed in cf404bfb60f270ee9b1bd50c5a49a5f4aed2efad

PR #189, opened 2023-12-11
Update `deploy` section in the readme

mmv08:
1. Fix the dead link for the releases page (follow-up: do tags like 0.1.0 work well for a monorepo?)
2. Remove the compiler settings from the section; we have a separate section about compiler settings.

### DIFF
--- a/4337/README.md
+++ b/4337/README.md
@@ -156,9 +156,9 @@ npm run test:e2e
 
 ### Deploy
 
-> :warning: **Make sure to use the correct commit when deploying the contracts.** Any change (even comments) within the contract files will result in different addresses. The tagged versions used by the Safe team can be found in the [releases](https://github.com/safe-modules/releases).
+> :warning: **Make sure to use the correct commit when deploying the contracts.** Any change (even comments) within the contract files will result in different addresses. The tagged versions used by the Safe team can be found in the [releases](https://github.com/safe-global/safe-modules/releases).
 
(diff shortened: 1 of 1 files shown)


4337/README.md
```diff
@@ -156,9 +156,9 @@ npm run test:e2e
 
 ### Deploy
 
-> :warning: **Make sure to use the correct commit when deploying the contracts.** Any change (even comments) within the contract files will result in different addresses. The tagged versions used by the Safe team can be found in the [releases](https://github.com/safe-modules/releases).
+> :warning: **Make sure to use the correct commit when deploying the contracts.** Any change (even comments) within the contract files will result in different addresses. The tagged versions used by the Safe team can be found in the [releases](https://github.com/safe-global/safe-modules/releases).
```
rmeissner, line 159: Currently our releases are not up to date, something we need to follow up on

rmeissner (approved):

4337/README.md
mmv08: If we plan to have more modules in the monorepo, I would employ the same setup with changelogs and stuff as we did for the safe-apps-sdk using changesets: https://github.com/safe-global/safe-apps-sdk/releases

nlordell:
> follow-up: do tags like 0.1.0 work well for a monorepo?)

What I have seen work well for monorepos is `project-path/vX.Y.Z` tag/releases.

So here, I would create a `4337/v0.2.0` tag

PR #191, opened 2023-12-13
Vendor FreshCryptoLib Libraries

nlordell:
This PR vendors the latest version for the FreshCryptoLib into the repository. The rationale is that both @mmv08 and myself will be doing some related work around WebAuthn PoCs using the Safe and the 4337 module and wanted to have a common set of Solidity code as a base for our work.

Some notes:
- The exact vendored version is recorded in `VERSION.md`
- Prettier formatting is ignored for these files
- The prettier check and write commands were adjusted to work better with v3.1.0. The existing prettier commands were needed for pre-v3 versions of prettier where the Solidity files were not automatically being considered for formatting, but no longer needed :)


mmv08:
> Prettier formatting is ignored for these files

Why? Let's make it pretty.

mmv08 (approved):

PR #195, opened 2023-12-21
Implement WebAuthn Shim for Passkeys tests

nlordell:
This PR adds a `tests/utils/webauthn.ts` module that implements a shim for the Browser WebAuthn API (sometimes conflated with Passkeys).

I verified the implementation by testing both credential creation and assertion (i.e. signing) using a reference WebAuthn/Passkey server implementation (specifically `SimpleWebAuthn` which was suggested on the [Passkeys.dev website](https://passkeys.dev/docs/tools-libraries/libraries/#typescript)).

OMG debugging signature issues is annoying :sob: :sob: :sob:

### DIFF
--- a/4337/.solhintignore
+++ b/4337/.solhintignore
@@ -1,0 +1,1 @@
+contracts/test/FCL/
--- a/4337/test/utils/webauthn.ts
+++ b/4337/test/utils/webauthn.ts
@@ -1,0 +1,231 @@
+/**
+ * This module provides a minimal shim to emulate the Web Authentication API implemented in browsers. This allows us to
+ * write tests where we create and authenticate WebAuthn credentials that are verified on-chain.
+ *
+ * This implementation is inspired by software authenticators found in the Awesome WebAuthn list [1].
+ *
+ * [1]: <https://github.com/herrjemand/awesome-webauthn#software-authenticators>
+ */
+
+import { p256 } from '@noble/curves/p256'
+import { ethers } from 'hardhat'
+import CBOR from 'cbor'
+
+export interface CredentialCreationOptions {
+  publicKey: PublicKeyCredentialCreationOptions
+}
+
+/**
+ * Public key credetial creation options, restricted to a subset of options that this module supports.
+ * See <https://w3c.github.io/webauthn/#dictionary-makecredentialoptions>.
+ */
+export interface PublicKeyCredentialCreationOptions {
+  rp: { id: string; name: string }
+  user: { id: Uint8Array; displayName: string; name: string }
+  challenge: Uint8Array
+  pubKeyCredParams: [{ type: 'public-key'; alg: -7 }]
+  timeout: undefined
+  excludeCredentials: undefined
+  authenticatorSelection: undefined
+  hints: undefined
+  attestation?: 'none'
+  attestationFormats: undefined
+  extensions: undefined
+}
+
+export interface CredentialRequestOptions {
+  publicKey: PublicKeyCredentialRequestOptions
+}
+
+/**
+ * Public key credetial request options, restricted to a subset of options that this module supports.
+ * See <https://w3c.github.io/webauthn/#dictionary-assertion-options>.
+ */
+export interface PublicKeyCredentialRequestOptions {
+  challenge: Uint8Array
+  timeout: undefined
+  rpId: string
+  allowCredentials: [
+    {
+      type: 'public-key'
+      id: Uint8Array
+      transports: undefined
+    },
+  ]
+  userVerification: undefined
+  hints: undefined
+  attestation?: 'none'
+  attestationFormats: undefined
+  extensions: undefined
+}
+
+/**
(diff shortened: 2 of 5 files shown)


4337/.solhintignore
```diff
@@ -0,0 +1 @@
+contracts/test/FCL/
```
nlordell, line 1: unrelated to this PR, just noticed when running `npm run lint` locally.

4337/test/utils/webauthn.ts
```diff
@@ -0,0 +1,244 @@
...
+ *
+ * This implementation is inspired by software authenticators found in the Awesome WebAuthn list [1].
+ *
+ * [1]: <https://github.com/herrjemand/awesome-webauthn#software-authenticators>
+ */
+
+import { p256 } from '@noble/curves/p256'
+import { ethers, BytesLike } from 'ethers'
+import CBOR from 'cbor'
+import { isArrayBuffer } from 'util/types'
```
mmv08, line 13: AFAIK this method is only available in Node. Ethers has a similar method which we could use. The motivation is to provide a better reference implementation for in-browser usage.

4337/test/utils/webauthn.ts
```diff
@@ -0,0 +1,244 @@
...
+
+  constructor(
+    public rp: string,
+    public user: Uint8Array,
+  ) {
+    this.pk = p256.utils.normPrivateKeyToScalar(p256.utils.randomPrivateKey())
+    this.id = ethers.dataSlice(ethers.keccak256(ethers.dataSlice(p256.getPublicKey(this.pk, false), 1)), 12)
+  }
+
+  public cosePublicKey(): string {
```
mmv08, line 94: Took me a while to find out what cose is; perhaps adding a comment together with the method definition is not a bad idea (I know that the actual implementation links the spec, but I was confused by the method name) https://datatracker.ietf.org/doc/html/rfc8152

4337/test/utils/webauthn.ts
```diff
@@ -0,0 +1,244 @@
...
+ *
+ * @param data data to encode to `base64url`
+ * @returns the `base64url` encoded data as a string.
+ */
+export function base64UrlEncode(data: BytesLike | ArrayBufferLike): string {
+  const bytes = isArrayBuffer(data) ? new Uint8Array(data) : ethers.getBytes(data)
+  return Buffer.from(bytes).toString('base64url')
+}
+
+function b2ab(buf: Buffer): ArrayBuffer {
```
mmv08, line 242: Imo it's better to stick to Uint8Array here (buffer extends Uint8Array)

Fun read: https://sindresorhus.com/blog/goodbye-nodejs-buffer

4337/test/webauthn/WebAuthn.spec.ts
```diff
@@ -0,0 +1,135 @@
...
+  }
+
+  const user = {
+    id: ethers.getBytes(ethers.id('chucknorris')),
+    name: 'chucknorris',
+    displayName: 'Chuck Norris',
+  }
+
+  describe('navigator.credentials.create()', () => {
+    it('Safe with 4337 Module Deployment', async () => {
```
mmv08, line 34: Is the test description off, or is the test incomplete?

mmv08 (approved):

4337/test/utils/webauthn.ts
nlordell: The spec is referenced below twice (for example here: https://github.com/safe-global/safe-modules/blob/26fdff1d5d252962e431fd63ece40fa8fb5d7518/4337/test/utils/webauthn.ts#L101), but I’ll add doc comments referencing the spec again as it wasn’t super clear that those are links to the spec.

4337/test/webauthn/WebAuthn.spec.ts
nlordell: Its a copy-pasta

4337/test/utils/webauthn.ts
nlordell: We now don't use `Buffer` anymore :smile: acf4235e71d0535c3b0b8cddeb4bb16ec62aff7b

4337/test/utils/webauthn.ts
nlordell: Added more context to the function documentation.

acf4235e71d0535c3b0b8cddeb4bb16ec62aff7b

4337/test/utils/webauthn.ts
nlordell: `Buffer` be gone!

acf4235e71d0535c3b0b8cddeb4bb16ec62aff7b

4337/test/webauthn/WebAuthn.spec.ts
nlordell: Fixed test descriptions in acf4235e71d0535c3b0b8cddeb4bb16ec62aff7b

Issue #173, opened 2023-11-27
Gas profile WebAuthn FCL signer implementation


rmeissner:
This issue is a follow-up to #172.

We should evaluate gas characteristics of existing WebAuthn implementations which may help guide us to choose the optimal WebAuthn library.

As a start, we should setup tests in the 4337 module that can with the FCL WebAuthn signer implementation compute gas usage for both:
- initial Safe deployment
- standalone user operation execution

#172 already has a rough comparison between different signature verification gas comparisons, so we can extrapolate gas amounts based on the findings in that repository. As a followup, we may choose to vendor additional implementations to compute exact gas amounts.

## Expected Output

- Vendor FCL code into our repository
- Manual process is fine at first (no GitHub action needed at first)

PR #197, opened 2023-12-22
Add E2E Test with WebAuthn Signer

nlordell:
A huge part of the way to #173

This PR adds an E2E test using the reference 4337 EntryPoint and bundler for a Safe deployment with a WebAuthn (i.e. Passkeys) signer.

Note that this E2E test is not 100% complete. In particular, we don't get test using the Safe after the launchpad deployment is completed. I want to add that test in a follow-up when I actually add the gas profiling for the WebAuthn signer.

### DIFF
--- a/4337/contracts/test/WebAuthnSigner.sol
+++ b/4337/contracts/test/WebAuthnSigner.sol
@@ -14,6 +14,7 @@ struct SignatureData {
 
 function checkSignature(bytes memory data, bytes calldata signature, uint256 x, uint256 y) view returns (bytes4 magicValue) {
     SignatureData calldata signaturePointer;
+    // solhint-disable-next-line no-inline-assembly
     assembly {
         signaturePointer := signature.offset
     }
--- a/4337/contracts/test/WebAuthnSigner.sol
+++ b/4337/contracts/test/WebAuthnSigner.sol
@@ -1,0 +1,90 @@
+// SPDX-License-Identifier: LGPL-3.0-only
+/* solhint-disable one-contract-per-file */
+pragma solidity >=0.8.0;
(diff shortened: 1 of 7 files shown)


4337/contracts/test/WebAuthnSigner.sol
```diff
@@ -0,0 +1,91 @@
...
+import {IUniqueSignerFactory} from "./SafeSignerLaunchpad.sol";
+
+struct SignatureData {
+    bytes authenticatorData;
+    bytes clientData;
+    uint256 challengeOffset;
+    uint256[2] rs;
+}
+
+function checkSignature(bytes memory data, bytes calldata signature, uint256 x, uint256 y) view returns (bytes4 magicValue) {
```
mmv08, line 15: TIL one can define functions outside the contract

4337/contracts/test/WebAuthnSigner.sol
```diff
@@ -0,0 +1,91 @@
...
+    SignatureData calldata signaturePointer;
+    // solhint-disable-next-line no-inline-assembly
+    assembly {
```
mmv08, line 18: memory safe? :D

mmv08 (approved):

4337/contracts/test/WebAuthnSigner.sol
nlordell: Added (and in a couple other places as well).

4337/contracts/test/WebAuthnSigner.sol
nlordell: Since 0.8

PR #201, opened 2024-01-10
Add Gas Metering Tests for WebAuthn Signers

### LINKED ISSUE #173: Gas profile WebAuthn FCL signer implementation

nlordell:
Fixes #173 

This PR adds tests that execute user operations on Safes with WebAuthn signers over the reference entrypoint and prints out the gas usage. It includes both Safe deployment as well as regular user operation usage.

The results are in:

```
  Safe4337Module - WebAuthn Owner
    executeUserOp - new account
           Used 2293926 gas for >WebAuthn signer Safe deployment<
      ✔ should execute user operation (357ms)
    executeUserOp - existing account
           Used 374142 gas for >WebAuthn signer Safe operation<
      ✔ should execute user operation (229ms)
```

### DIFF
--- a/4337/test/e2e/WebAuthnSigner.spec.ts
+++ b/4337/test/e2e/WebAuthnSigner.spec.ts
@@ -1,9 +1,8 @@
 import { expect } from 'chai'
-import CBOR from 'cbor'
 import { deployments, ethers, network } from 'hardhat'
 import { bundlerRpc, prepareAccounts, waitForUserOp } from '../utils/e2e'
 import { chainId } from '../utils/encoding'
-import { AuthenticatorAttestationResponse, WebAuthnCredentials, base64UrlEncode } from '../utils/webauthn'
+import { WebAuthnCredentials, extractChallengeOffset, extractPublicKey, extractSignature } from '../utils/webauthn'
 
 describe('E2E - WebAuthn Signers', () => {
   before(function () {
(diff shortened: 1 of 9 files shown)


4337/test/e2e/WebAuthnSigner.spec.ts
```diff
@@ -200,76 +197,4 @@ describe('E2E - WebAuthn Signers', () => {
     const safeInstance = await ethers.getContractAt('SafeL2', safe)
     expect(await safeInstance.getOwners()).to.deep.equal([signerAddress])
   })
-
```
nlordell, line 203: Moved to `utils/webauthn`.

mmv08 (approved):
Spenny

PR #213, opened 2024-01-18
Extract signature validation to a separate function in the launchpad contract

mmv08:
While working on the WebAuthn PoC app, I got stuck with the gas estimation. After debugging, I discovered that the webauthn contract reverts if you provide an invalid signature. This goes against ERC-4337 specs:
> If the account does not support signature aggregation, it MUST validate the signature is a valid signature of the userOpHash, and SHOULD return SIG_VALIDATION_FAILED (and not revert) on signature mismatch. Any other error should revert.

This PR fixes it using a function similar to the `Safe4337Module`. No tests because experimental 😄


nlordell (approved):
LGTM

PR #221, opened 2024-01-23
New Signature Encoding for Webauthn Signer

### LINKED ISSUE #220: Fix WebAuthn Implementation to Use Standard Client Data JSON Serialization
The Webauthn signer requires an offset position for challenge, but this can be simplified as the Webauthn standard mandates a specific JSON serialization format.

The expected outcome of this issue is to modify the Webauthn signer’s implementation of computing the signing message to take advantage of the strict JSON serialization format.

nlordell:
Fixes #220 

This PR proposes a new signature encoding scheme for the Webauthn signer. This takes advantage of the [`clientDataJSON` serialization specification]() which requires a specific ordering and formatting (no additional whitespace) for the fields). Notably, this scheme:

* reduces the amount of calldata by 114 bytes per signature; this is non-trivial in the context of reducing calldata for rollups (where calldata is more expensive than computation).
* no longer requires a trusted `challengeOffset` (which has some security implications).

I plan to upstream these changes to FCL after doing a bit of investigation on potentially optimizing:
- base64 encoding of the challenge (since we use a fixed 32-byte challenge while the library allows for arbitrary `bytes`)
- byte concatenation logic to try and make it slightly more gas efficient

### DIFF
--- a/modules/4337/contracts/experimental/WebAuthn.sol
+++ b/modules/4337/contracts/experimental/WebAuthn.sol
@@ -1,0 +1,43 @@
+// SPDX-License-Identifier: LGPL-3.0-only
+pragma solidity >=0.8.0;
+
+import {FCL_ecdsa_utils} from "../vendor/FCL/FCL_ecdsa_utils.sol";
+import {Base64Url} from "../vendor/FCL/utils/Base64Url.sol";
+
+library WebAuthn {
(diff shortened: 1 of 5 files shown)


mmv08 (approved):
A slight interference with my extraction PR but I'll manage it 😄

Issue #219, opened 2024-01-22
Refactor WebAuthn Signer to Allow Different Signature Verification Implementations


nlordell:
Related to #218, but the deployed `WebauthnSigner` implementation constructor should support specifying not only the public key, but also the P-256 curve implementation address.

The expected outcome of this issue is a refactored `WebauthnSigner` and `WebauthnSignerFactory` with an additional P-256 curve implementation parameter. As a bonus, it would be nice to emulate a P-256 signature verification pre-compile in tests to make sure that the `WebauthnSigner` and `WebauthnSignerFactory` implementations support the pre-compile version as well.

Issue #218, opened 2024-01-22
Move P-256 Signature Verification To Separate Contract


nlordell:
The current `WebauthnSigner` implementation embeds code for verifying P-256 signature in each instance it deploys, which leads to large deployment overhead for passkeys. As such, we should instead deploy a single P-256 curve signature verification contract that can be reused by multiple `WebauthnSigner` instances.

The interface should be designed in a way that the Solidity P-256 curve implementation can be easily replaced by the P-256 precompile that is currently proposed as an EIP: https://eips.ethereum.org/EIPS/eip-7212. For example, we can choose to have the shared P-256 signature verifying contract exactly match the pre-compile interface.

PR #223, opened 2024-01-24
Move P-256 signature verification to a separate contract, Refactor WebAuthn Signer to Allow Different Signature Verification Implementations

mmv08:
This PR:
- Implements https://github.com/safe-global/safe-modules/issues/219 and https://github.com/safe-global/safe-modules/issues/218

Https://github.com/daimo-eth/p256-verifier heavily inspired the implementation. However, I tried to reuse as much of the FCL code as possible to make future upgrades easier.

This is the final contract hierarchy:
Untitled-2024-02-05-1850

1. `P256Verifier.sol` is a `secp256r1` curve signature verifier; it follows the interface defined in https://eips.ethereum.org/EIPS/eip-7212.
2. `P256Wrapper.sol` is a wrapper that points to a verifier address (can be later replaced with the precompile's address). It implements two functions advised by the EIP: `verifySignatureWithoutMalleabilityCheck` and `verifySignature`
3. `WebAuthnVerifier.sol` inherits from `P256Wrapper.sol` and adds a few WebAuthn-specific things, like WebAuthn message generation and signature verification with authenticator flags check.
4. `WebAuthnSigner.sol` and `WebAuthnSignerFactory.sol` call into the `WebAuthnVerifier` to verify a signature. I hardcoded the authenticator flag to be `0x04` (require user verification), compared to our previously set `0x01` (user presence). More on that here: https://developers.yubico.com/WebAuthn/WebAuthn_Developer_Guide/User_Presence_vs_User_Verification.html. I believe UV makes more sense for the type of application we're building.

**Benchmarks**:
- WebAuthnSigner codesize went from 8143 to 1610 bytes
- Used 1013491 gas for WebAuthn signer Safe deployment. Before: 2339762 (Numbers are non-deterministic)
- Used 378029 gas for WebAuthn signer Safe operation. Before: 375184 (Numbers are non-deterministic)

**Notes**:
I was torn between two approaches, and both had their pros and cons:
1. Keep everything as close to the standard as possible: a singleton verifier and a wrapper contract with the functions defined in the standard. Inherit that verifier in the webauthn verifier and use that to verify the signature. The benefit of this approach is standard compatibility and the ease of switching to the precompile by just swapping an address in the wrapper contract. The downside is that it requires a lot of external calls to verify a signature. e.g., `contract -> signer -> webauthn verifier -> p256 verifier`. In my benchmarks, this approach, on average, consumed 10k more gas than the next one.
2. Remove a call to the p256 verifier by directly inheriting the p256 verification code in the webauthn verifier. It consumes less gas but is also not compatible with the standard.

I decided to use the first approach for easy future changes where we don't need to produce new code but swap an address.


nlordell:
> Used 378800 gas for WebAuthn signer Safe operation. Before: 375184 (Numbers are non-deterministic)

The extra cold contract call, numbers add up.

> Used 1263258 gas for WebAuthn signer Safe deployment. Before: 2339762 (Numbers are non-deterministic)

This means that only after `(2339762−1263258) / (378800−375184) ~= 300` transactions, the larger contract code would make more sense.

modules/4337/contracts/experimental/P256Wrapper.sol
```diff
@@ -0,0 +1,62 @@
...
+ */
+abstract contract P256Wrapper {
+    address immutable VERIFIER;
+
+    constructor(address verifier) {
+        require(verifier != address(0), "P256: invalid verifier address");
+        VERIFIER = verifier;
+    }
+
+    /// P256 curve order n/2 for malleability check
```
nlordell, line 19: uber-OCD-nit: I would use consistent comment format for all fields, here it has `///`, but in other places it has `/** */`.

modules/4337/contracts/experimental/P256Wrapper.sol
```diff
@@ -0,0 +1,62 @@
...
+    function verifySignatureAllowMalleability(
+        bytes32 message_hash,
+        uint256 r,
+        uint256 s,
+        uint256 x,
+        uint256 y
+    ) internal view returns (bool) {
+        bytes memory args = abi.encode(message_hash, r, s, x, y);
+        (bool success, bytes memory ret) = VERIFIER.staticcall(args);
+        assert(success); // never reverts, always returns 0 or 1
```
nlordell, line 40: Curious on the use of assert here. Does this compile away with optimisations? If not, I would use:

```
success;
```

To suppress the Solidity warnings.

modules/4337/contracts/experimental/P256Wrapper.sol
```diff
@@ -0,0 +1,62 @@
...
+
+        return abi.decode(ret, (uint256)) == 1;
+    }
```
nlordell, line 43: I wonder if there are significant any gas savings in using assembly by:
1. not going over `abi.encode` which allocates memory
2. not using `bytes memory ret`, but writing only 32 bytes to the scratch space with the `call` opcode directly (instead of the separate `returndatacopy` that this code generates).

Would not do it in this PR but a follow up.

nlordell (commented):
Code looks great.

The only problem I see is that the invariant `isValidSignatureForSigner(data, signature, signerData) == createSigner(signerData).isValidSignature(data, signature)` is no longer guaranteed but should be.

modules/4337/contracts/experimental/WebAuthnSigner.sol
```diff
@@ -127,8 +147,9 @@ contract WebAuthnSignerFactory is IUniqueSignerFactory, SignatureValidatorConsta
      * @param y The y-coordinate of the signer.
      * @return The address of the signer.
      */
-    function _getSigner(uint256 x, uint256 y) internal view returns (address) {
-        bytes32 codeHash = keccak256(abi.encodePacked(type(WebAuthnSigner).creationCode, x, y));
+    function _getSigner(uint256 x, uint256 y, address verifier) internal view returns (address) {
+        // We need to convert the address to uint256, so it is padded to 32 bytes
+        bytes32 codeHash = keccak256(abi.encodePacked(type(WebAuthnSigner).creationCode, x, y, uint256(uint160(verifier))));
```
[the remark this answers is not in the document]
mmv08: Oops, I actually messed that up. Initially my intention was to create a verifier-agnostic factory and later on I switched the approach but forgot to cleanup. Thanks for the catch.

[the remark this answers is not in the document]
mmv08: Actually, I think both approaches are valid 🤔 Which one do you think is more elegant?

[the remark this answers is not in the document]
nlordell: > I think both approaches are valid

Wouldn't that require deploying a `new WebauthnVerifier(...)` for each possible `P256Verifier` implementation?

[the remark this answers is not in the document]
mmv08: Yeah, but there shouldn't be many, right? So the signer would just call into the verifier contract

Also, re-reading the [issue](https://github.com/safe-global/safe-modules/issues/219), I think this is what was expected: 
> The expected outcome of this issue is a refactored WebauthnSigner and WebauthnSignerFactory with an additional P-256 curve implementation parameter. 

Not sure why I switched the approach 🤔

modules/4337/contracts/experimental/P256Verifier.sol
```diff
@@ -0,0 +1,78 @@
...
+ * - input[  0: 32] = signed data hash
+ * - input[ 32: 64] = signature r
+ * - input[ 64: 96] = signature s
+ * - input[ 96:128] = public key x
+ * - input[128:160] = public key y
+ * The result is a bytes array where the first 32 bytes represent 0x00..00 (invalid) or 0x00..01 (valid).
+ * For more details, refer to the EIP-7212 specification: https://eips.ethereum.org/EIPS/eip-7212
+ */
+contract P256VerifierWithFallback {
+    fallback(bytes calldata input) external returns (bytes memory) {
```
nlordell, line 21: You can add an:

```solidity
    function ecdsaVerify(
        bytes32 message,
        uint256 r,
        uint256 s,
        uint256 x,
        uint256 y
    ) internal view abstract returns (bool);
```

And then you could:

```solidity
contract FCLWebAuthnVerifier is WebAuthnVerifier {
    function ecdsaVerify(
        bytes32 message,
        uint256 r,
        uint256 s,
        uint256 x,
        uint256 y
    ) internal view override returns (bool) {
        return FCL_ecdsa.ecdsa_verify(message, r, s, x, y);
    }
}

contract PrecompileWebAuthnVerifier is WebAuthnVerifier {
    function ecdsaVerify(
        bytes32 message,
        uint256 r,
        uint256 s,
        uint256 x,
        uint256 y
    ) internal view override returns (bool) {
        // implement with precompile...
    }
}
```

modules/4337/contracts/experimental/P256Verifier.sol
```diff
@@ -0,0 +1,78 @@
...
+
+        bytes32 hash = bytes32(input[0:32]);
+        uint256 r = uint256(bytes32(input[32:64]));
+        uint256 s = uint256(bytes32(input[64:96]));
+        uint256 x = uint256(bytes32(input[96:128]));
+        uint256 y = uint256(bytes32(input[128:160]));
+
+        uint256 ret = FCL_ecdsa.ecdsa_verify(hash, r, s, x, y) ? 1 : 0;
+
+        return abi.encodePacked(ret);
```
nlordell, line 34: Wouldn't this be clearer as:

```suggestion
        bool success = FCL_ecdsa.ecdsa_verify(hash, r, s, x, y);

        return abi.encode(success);
```

Or was this done on purpose to be more explicit about the fact that it implements the pre-compile API? (If so, I would at a comment, as this wasn't immediately obvious).

nlordell (approved):
Looks good.

I left a suggestion with abstract contracts to make the path to having a separate pre-compile implementation more trivial, but I'm also OK with the implementation as is (seeing as this is an experimental contract and does not need to be 100% flushed out).

As a side comment, can you include points from your post on Slack in the PR for documentation purposes?

modules/4337/contracts/experimental/P256Verifier.sol
mmv08: Could you please elaborate on the hierarchy you have in mind? I _think_ I got it, but the message and the file you left a comment on confuse me.

modules/4337/contracts/experimental/P256Wrapper.sol
mmv08: I personally like to stick to `//` for regular comments and `/** */` for multiline. How does that sound?

modules/4337/contracts/experimental/P256Wrapper.sol
mmv08: @nlordell I actually implemented this. Please re-review

nlordell (approved):

PR #224, opened 2024-01-24
Improve developer experience of running e2e tests

mmv08:
The golden rule of developer experience is that everything should be one command away (running, building, testing a project, etc). This PR does precisely that for end-to-end tests, removing separate container run steps by creating a `run.sh` script


modules/4337/hardhat.config.ts
```diff
@@ -66,6 +66,7 @@ const userConfig: HardhatUserConfig = {
   },
   networks: {
     localhost: {
+      url: 'http://localhost:8545',
```
mmv08, line 69: For some reason, hardhat refused to connect to the node if this URL wasn't provided. 
According to @nlordell, some IPv4/IPv6 shenanigans could be involved, but I don't think it's worth looking into.

nlordell: VPN maybe? :stuck_out_tongue:

modules/4337/test/e2e/run.sh
```diff
@@ -0,0 +1,17 @@
...
+        echo "ERROR: timeout waiting for local node and bundler to start"
+        docker compose logs
+        exit 1
+    fi
+    sleep 1
+done
+
+hardhat test --deploy-fixture --network localhost --grep '^E2E - '
+
+docker compose down
```
nlordell, line 17: Interesting that you do a tear-down for every test run. I hadn't really thought about it, but it makes it so things run from a more deterministic state :+1:.

nlordell (approved):
:100:

modules/4337/test/e2e/run.sh
mmv08: Neither had I 😄

PR #239, opened 2024-01-26
[L-01] Specific Solidity Version for Contracts

nlordell:
This PR addresses the issue L-01 from the audit report.

Specifically, it implements the suggestion to not use floating pragmas, but specific Solidity compiler versions for the contracts.


remedcu (approved):
nit: `=` is not necessary as per [this](https://swcregistry.io/docs/SWC-103/#semver_floating_pragma_fixedsol).

PR #245, opened 2024-01-31
[M-01] Potential Obsolescence for `Safe4337Module`

mmv08:
As vaguely explained by the OpenZeppelin team, the primary motivation for M-01 is not compatibility with unreleased versions but a more prompt awareness of the new potential changes in the ERC-4337 standard, as they deemed incompatibility with the standard as one of the leading security concerns for the protocol.

This PR presents an alternative solution for M-01 by adding a secondary e2e test job that uses an upstream version of the bundler for the test. This should test both the `Entrypoint <> Module` API compatibility (because the bundler will deploy its own Entrypoint to the locally running geth chain) as well as the storage access rules. The test also runs on schedule (1st and 15th of each month)

### DIFF
--- a/.github/workflows/ci_4337_upstream_bundler.yml
+++ b/.github/workflows/ci_4337_upstream_bundler.yml
@@ -1,0 +1,21 @@
+name: 4337 Module End-to-End Tests With Upstream Bundler
+on:
+  schedule:
+    # * is a special character in YAML so you have to quote this string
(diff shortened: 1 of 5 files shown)


nlordell (approved):
Nice and pragmatic :+1:.

.github/workflows/ci_4337.yml
```diff
@@ -31,6 +31,18 @@ jobs:
       - run: |
           npm ci
           npm run test:e2e -w modules/4337
+  e2e-upstream-bundler:
```
nlordell, line 34: As suggested by OZ, can we move this to its own workflow file and make it run on a schedule?

Issue #228, opened 2024-01-25
Implement `WebauthnSigletonSigner` PoC


nlordell:
WebAuthn signature verification with the 4337 module is only implemented using the launchpad strategy, however the singleton signer strategy introduced in #159. could also be applied for WebAuthn signers. For some applications, the simplicity (and gas efficiency - no additional contract deployment!) of a singleton signer might outweigh the downsides (a singleton signer can only represent one signer for a Safe, so multiple deployments may be needed).

The expected outcome of this issue would be to create a `WebAuthnSingletonSigner` based on the `TestSingletonSigner` introduced in the aforementioned PR that works for verifying passkey signatures along with a test to verify its functionality. _Optionally_ gas consumption for the new singleton signer scheme should be recorded so it can be easily compared with the "launchpad strategy" that we use for the current WebAuthn PoC integration.

### Recommended Reading

* #148 
* #159

PR #252, opened 2024-02-07
Refactor WebAuthn Signature Validation Logic

nlordell:
This PR slightly refactors the WebAuthn signature validation logic in preparation for sharing code when implementing #228.

Essentially, it creates a `SignatureValidator` abstract base class for sharing that supports both legacy and current EIP-1271 functions. As a consequence of the refactoring, `ISafeSingletonSigner` now uses `bytes32` messages for signature validation instead of `bytes memory`, in order to be more of an analogy to the current EIP-1271 standard (instead of the legacy one).

### DIFF
--- a/modules/4337/contracts/experimental/SafeSignerLaunchpad.sol
+++ b/modules/4337/contracts/experimental/SafeSignerLaunchpad.sol
@@ -27,14 +27,14 @@ interface IUniqueSignerFactory {
     /**
      * @notice Verifies a signature for the specified address without deploying it.
      * @dev This must be equivalent to first deploying the signer with the factory, and then verifying the signature
-     * with it directly: `factory.createSigner(signerData).isValidSignature(data, signature)`
-     * @param data The data whose signature should be verified.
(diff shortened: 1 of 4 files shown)


modules/4337/contracts/experimental/SafeSignerLaunchpad.sol
```diff
@@ -27,14 +27,14 @@ interface IUniqueSignerFactory {
     /**
      * @notice Verifies a signature for the specified address without deploying it.
      * @dev This must be equivalent to first deploying the signer with the factory, and then verifying the signature
-     * with it directly: `factory.createSigner(signerData).isValidSignature(data, signature)`
-     * @param data The data whose signature should be verified.
+     * with it directly: `factory.createSigner(signerData).isValidSignature(message, signature)`
+     * @param message The singing message.
```
mmv08, line 31:
```suggestion
     * @param message The signed message.
```

mmv08 (approved):
Looks good! One tiny thing

PR #253, opened 2024-02-07
Allow Specifying Docker CLI For E2E

nlordell:
This PR modifies the E2E `run.sh` script to allow configuration of the Docker CLI to use. Why? I use `podman` instead of `docker`. Another potential use case would be to use a specific `docker` client instead of the one configured in `$PATH`.

I can now successfully run:

```
DOCKER=podman npm run test:e2e
```

Note that not specifying any configuration in the environment will cause the script to default ot the `docker` command. On my system without `docker` installed:

```
$ npm run test:e2e

> @safe-global/safe-erc4337@0.2.0 test:e2e
> ./test/e2e/run.sh

./test/e2e/run.sh: line 5: docker: command not found
```

### DIFF
--- a/modules/4337/test/e2e/run.sh
+++ b/modules/4337/test/e2e/run.sh
@@ -1,18 +1,20 @@
 #!/usr/bin/env bash
 
-bundler_container="bundler"
+DOCKER="${DOCKER:-docker}"
 
+bundler_container="bundler"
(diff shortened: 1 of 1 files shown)


remedcu (approved):
LGTM

P.S. Learned a new thing on why using `:-` instead of `=` while assigning is better, as it makes sure passing nothing in the parameter also results in using the default value.

PR #256, opened 2024-02-07
Emit `SafeOperation` Events

nlordell:
This PR changes the module to emit `SafeOperation` events in order to facilitate indexing by the Safe transaction service. Note that we emit an event with a `struct` instead of flattened event fields to avoid "stack too deep" issues.

Unfortunately the gas overhead of including this event is quite large:

| Operation                              | Before | After  | Change        |
| -------------------------------------- | ------ | ------ | ------------- |
| Safe Deployment + Native Transfers     | 364590 | 383230 | 18640 / 5.1%  |
| Native Transfer                        | 88475  | 107115 | 18640 / 21.1% |
| Safe Deployment + ERC20 Transfer       | 370552 | 390191 | 19639 / 5.3%  |
| ERC20 Transfer                         | 94437  | 114088 | 19651 / 20.1% |
| Safe Deployment + ERC721 Token Minting | 412351 | 431990 | 19639 / 4.8%  |
| ERC721 Token Minting                   | 136248 | 155887 | 19639 / 14.4% |

It is currently in draft, as there are no accompanying tests, and have not yet committed to including this change.

### DIFF
--- a/modules/4337/contracts/Safe4337Module.sol
+++ b/modules/4337/contracts/Safe4337Module.sol
@@ -70,12 +70,46 @@ contract Safe4337Module is IAccount, HandlerContext, CompatibilityFallbackHandle
     }
 
     /**
+     * @notice An error indicating that the entry point used when deploying a new module instance is invalid.
+     */
+    error InvalidEntryPoint();
+
+    /**
+     * @notice An error indicating that the caller does not match the Safe in the corresponding user operation.
+     * @dev This indicates that the module is being used to validate a user operation for a Safe that did not directly
+     * call this module.
+     */
+    error InvalidCaller();
+
+    /**
+     * @notice An error indicating that the call validating or executing a user operation was not called by the
+     * supported entry point contract.
+     */
+    error UnsupportedEntryPoint();
+
+    /**
+     * @notice An error indicating that the user operation `callData` does not correspond to one of the two supported
+     * execution functions: `executeUserOp` or `executeUserOpWithErrorString`.
+     */
+    error UnsupportedExecutionFunction(bytes4 selector);
+
+    /**
+     * @notice An error indicating that the user operation failed to execute successfully.
+     * @dev The contract reverts with this error when `executeUserOp` is used instead of bubbling up the original revert
+     * data. When bubbling up revert data is desirable, `executeUserOpWithErrorString` should be used instead.
+     */
+    error ExecutionFailed();
+
+    /**
      * @notice The address of the EntryPoint contract supported by this module.
      */
     address public immutable SUPPORTED_ENTRYPOINT;
 
     constructor(address entryPoint) {
-        require(entryPoint != address(0), "Invalid entry point");
+        if (entryPoint == address(0)) {
+            revert InvalidEntryPoint();
+        }
+
         SUPPORTED_ENTRYPOINT = entryPoint;
     }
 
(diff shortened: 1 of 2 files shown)


modules/4337/contracts/Safe4337Module.sol
```diff
@@ -69,6 +69,31 @@ contract Safe4337Module is IAccount, HandlerContext, CompatibilityFallbackHandle
...
+        uint48 validAfter;
+        uint48 validUntil;
+        address entryPoint;
+        bytes signatures;
+    }
+
+    /**
+     * @notice An event that gets emitted when a Safe operation executes.
+     */
+    event SafeOperation(SignedSafeOp operation);
```
mmv08, line 95: Nothing needs to be indexed in this event, right?

nlordell: That is a question for @Uxio0

Uxio0: I think indexing the safe address could be important for the future (I'm thinking for example in Dune analytics). Also would it be posible to emit the `SafeOpHash` or is it too much?

Uxio0: I think indexing the safe address could be important for the future (I'm thinking for example in Dune analytics). Also would it be posible to emit the `SafeOpHash` or is it too much?

Uxio0: As long as we trust the entrypoint event not to change and as we can correlate entrypoint events with ours (I don't know if it can be done), it's alright

Uxio0: The point here is not what information we have on the entrypoint. The issue is that entrypoint is missing crucial data (the signature) to build the Safe operation, and as we already have to emit an event we add all the information to build the SafeOperation without depending on the entrypoint. All the data but the signature can be obtained from the entrypoint, so following that logic we should only emit an event with the signature

nlordell: One possible alternative is to emit:
1. UserOpHash (so that it can be indexed against a user operation [event](https://github.com/eth-infinitism/account-abstraction/blob/fa61290d37d079e928d92d53a122efcc63822214/contracts/interfaces/IEntryPoint.sol#L29))
2. Signature bytes.

The issue I see, however, is that even user operation data is not emitted as part of events, so some additional indexing (like parsing calldata) might still be necessary. Additionally, AFAIU, the entry point is not guaranteed to be called by an EOA, so it might even require tracing in order to get the calldata to extract the user operation parameters.

Uxio0: I'm ok with only indexing the missing data as long as you emit `UserOpHash` so I can match the information in the entrypoint, but we cannot count on relying on tracing, as it's not available for some L2 networks

mmv08: If we're only missing a signature from the event emitted by the entrypoint, I'd contact the 4337 team to see if they can consider adding it. They also have a release coming, so I'd hurry up with this suggestion.

Uxio0: We are missing also the `initCode`, `callData`, `data` for the paymaster... If they are willing to add all the data to the event that would be awesome, but I don't think they would be willing to

mmv08: Hmm, somehow https://jiffyscan.xyz/userOpHash/0xd0c9c8761cd347d0642741c7998be32ab2855471dce4a3c28727fa0ba49dfc4c?network=mainnet can index it. Perhaps we can reach out to them to learn how they implemented it

Uxio0: I can do that for mainnet too, using tracing 😉

mmv08: Screenshot 2024-02-11 at 17 47 34

Here's the list of networks that they support - guess that's doable without tracing

Uxio0: I guess then they are using the bundler APIs https://docs.alchemy.com/reference/bundler-api-endpoints

PR #261, opened 2024-02-13
Sign All Gas Limit and Price Parameters

### LINKED ISSUE #250: Sign All Gas Parameters in ERC-4337 0.7 `PackedUserOperation`
## Context / issue

The new ERC-4337 EntryPoint v0.7.0 has a new `PackedUserOperation` format. Initial support for this user operation was added in #243, however we sign the fully packed gas parameters instead of each individual gas parameter seperately, making it less clear to the use exactly what is being signed.

## Proposed solution

Unpack the gas parameter data and include that in the EIP-712 `SafeUserOperation` struct that gets signed.

## Alternatives

Save on gas and keep signing the packed gas parameters, at the cost of clearly understanding what is being signed. There may also be Solidity "stack-too-deep" errors which may be hard to work around (although it _should_ not be an issue with how we compute the hashes in the latest code).

nlordell:
Fixes #250 

This PR changes the `SafeOperation` type to include ALL gas parameters for signing, instead of packing them. This ensures the maximum amount of data is available to the user when signing user operations, instead of trusting gas parameters from a Dapp.

Note a subtle change in the order of the "fee per gas" fields, this was done to match the order in the `PackedUserOperation` struct, as well as the order in EIP-1559 transactions. I figured, since we are changing the signing format anyway (notice the different integer types), this wouldn't be a big deal.

Additionally, we have a new test that verifies that `safeOp` hashes are computed off-chain in the same way they are computed on-chain. This is to catch issues with field orders being slightly different in the `EncodedSafeOp` struct that we would otherwise miss. Notably, the "fee per gas" parameters are always set to the same value, and our existing test suite was not able to catch swapped order in `EncodedSafeOp` fields.

### DIFF
--- a/modules/4337/test/erc4337/Safe4337Module.spec.ts
+++ b/modules/4337/test/erc4337/Safe4337Module.spec.ts
@@ -40,26 +40,42 @@ describe('Safe4337Module', () => {
 
   describe('getOperationHash', () => {
     it('should correctly calculate EIP-712 hash of the operation', async () => {
-      const { validator, safeModule, entryPoint } = await setupTests()
+      const { entryPoint, validator } = await setupTests()
 
-      const safeAddress = ethers.hexlify(ethers.randomBytes(20))
-      const validAfter = (await timestamp()) + 10000
-      const validUntil = validAfter + 10000000000
+      const safeOp = {
+        safe: '0x0000000000000000000000000000000000000011',
+        nonce: 0x12,
+        initCode: '0x13',
+        callData: '0x14',
+        preVerificationGas: 0x15,
+        verificationGasLimit: 0x16,
+        callGasLimit: 0x17,
+        maxPriorityFeePerGas: 0x18,
+        maxFeePerGas: 0x19,
+        paymasterAndData: '0x1a',
+        validAfter: 0x1b,
+        validUntil: 0x1c,
+        entryPoint: entryPoint.target as string,
+      }
+      const userOp = {
+        sender: safeOp.safe,
+        nonce: safeOp.nonce,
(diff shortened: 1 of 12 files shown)


modules/4337/test/erc4337/Safe4337Module.spec.ts
```diff
@@ -82,22 +87,63 @@ describe('Safe4337Module', () => {
...
         ['paymasterAndData', '0x2a'],
         ['signature', signature({ validAfter: 0x2b, validUntil: 0x1c })],
         ['signature', signature({ validAfter: 0x1b, validUntil: 0x2c })],
-      ]) {
+      ] as const) {
         expect(await validator.getOperationHash({ ...userOp, [name]: value })).to.not.equal(operationHash)
       }
     })
+
+    it('should match locally computed Safe operation hash', async () => {
```
mmv08, line 109: how is this different from the test on line 41 in the same file?

nlordell: Oh, I missed that existing test. It should do the same thing. The issue is that the existing test uses the default Safe operation values, which in particular means `maxFeePerGas == maxPriorityFeePerGas` and does not (as per the PR description):

> [...] catch issues with field orders being slightly different in the `EncodedSafeOp` struct that we would otherwise miss. Notably, the "fee per gas" parameters are always set to the same value, and our existing test suite was not able to catch swapped order in `EncodedSafeOp` fields.

I'm going to remove this new test case and replace the existing test with this new logic.

mmv08 (approved):

Issue #226, opened 2024-01-25
Use the production entrypoint contract for testing


mmv08:
As discussed internally, there are potential benefits to using the production version of the entry point contract for testing. We already do that for e2e testing, but the integration tests use a mock. This is the improvement we want for the upcoming release of 0.7.0 entrypoint compatibility.

Deliverables:
- Refactoring the tests to use the entrypoint 0.7.0 everywhere

Potential blockers:
- I'm not sure if the entrypoint contract 0.7.0 is available in package repositories like npm, may need to get creative with fetching it

nlordell:
Related to #211

PR #262, opened 2024-02-13
Use production entrypoint in integration tests

mmv08:
This PR:
- Implements https://github.com/safe-global/safe-modules/issues/226 by deploying the EntryPoint contract and using it through the `IEntryPoint` interface
- Removed the test entrypoint contract
- Had to adjust quite some test because of the difference in entrypoint implementations (for example the test entrypoint didn't care about the pre-fund)


modules/4337/src/utils/userOp.ts
```diff
@@ -80,19 +80,19 @@ export const buildSafeUserOp = (template: OptionalExceptFor<SafeUserOperation, '
   return {
     safe: template.safe,
     nonce: template.nonce,
-    initCode: template.initCode || '0x',
-    callData: template.callData || '0x',
-    callGasLimit: template.callGasLimit || 2000000,
-    verificationGasLimit: template.verificationGasLimit || 500000,
-    preVerificationGas: template.preVerificationGas || 60000,
+    initCode: template.initCode ?? '0x',
```
mmv08, line 83: If you pass a falsy value here, let's say (0) the `||` operator would default to the default value, the `??` operator only checks for null and undefined without coercing the value to a boolean

modules/4337/test/erc4337/ERC4337ModuleExisting.spec.ts
```diff
@@ -9,20 +9,22 @@ describe('Safe4337Module - Existing Safe', () => {
...
       user1,
       safe,
+      relayer,
       validator: module,
       entryPoint,
     }
   })
 
-  describe('executeUserOp - existing account', () => {
+  describe('handleOps [- existing] account', () => {
```
nlordell, line 27: Looks like a search and replace change?
```suggestion
  describe('handleOps - existing account', () => {
```

modules/4337/test/erc4337/ERC4337ModuleExisting.spec.ts
```diff
@@ -54,22 +59,28 @@ describe('Safe4337Module - Existing Safe', () => {
         '0x',
         '0',
         await entryPoint.getAddress(),
+        false,
+        false,
+        {
+          maxFeePerGas: '0',
+        },
```
nlordell, line 66: Interesting... I would have emulated this by doing a:

```
await entryPoint.depositTo(safe.address, { value: ethers.parseEther("1.0") })
```

modules/4337/test/erc4337/ERC4337ModuleExisting.spec.ts
```diff
@@ -99,15 +115,20 @@ describe('Safe4337Module - Existing Safe', () => {
...
     })
 
     it('reverts on failure', async () => {
       const { user1, safe, validator, entryPoint } = await setupTests()
 
-      await user1.sendTransaction({ to: await safe.getAddress(), value: ethers.parseEther('0.000001') })
-      expect(await ethers.provider.getBalance(await safe.getAddress())).to.be.eq(ethers.parseEther('0.000001'))
+      // Make sure to send enough ETH for the pre-fund but less than the transaction
+      await user1.sendTransaction({ to: await safe.getAddress(), value: ethers.parseEther('0.4') })
+      expect(await ethers.provider.getBalance(await safe.getAddress())).to.be.eq(ethers.parseEther('0.4'))
```
nlordell, line 131: You can prefund the account using `depositTo` to the EntryPoint contract, which would (IMO) make this test closer to the original and easier to understand.

modules/4337/test/erc4337/ERC4337ModuleExisting.spec.ts
```diff
@@ -119,16 +140,18 @@ describe('Safe4337Module - Existing Safe', () => {
...
-
-      const transaction = await entryPoint.executeUserOp(userOp, ethers.parseEther('0.000001')).then((tx) => tx.wait())
-      const logs = transaction.logs.map((log) => entryPoint.interface.parseLog(log))
-      const emittedRevert = logs.some((l) => l?.name === 'UserOpReverted')
-
-      expect(emittedRevert).to.be.true
+      const userOpHash = await entryPoint.getUserOpHash(userOp)
+      // Error('Execution failed')
+      const expectedReturnData =
+        '0x08c379a000000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000000010457865637574696f6e206661696c656400000000000000000000000000000000'
```
nlordell, line 146: nit: This should be doable with Ethers:
```suggestion
      const expectedReturnData = validator.interface.encodeErrorResult('Error(string)', 'Execution failed')
```

modules/4337/test/erc4337/ERC4337ModuleExisting.spec.ts
```diff
@@ -170,12 +196,13 @@ describe('Safe4337Module - Existing Safe', () => {
...
       const userOp = buildUserOperationFromSafeUserOperation({ safeOp, signature })
 
-      const transaction = await entryPoint.executeUserOp(userOp, ethers.parseEther('0.000001')).then((tx) => tx.wait())
-      const logs = transaction.logs.map((log) => entryPoint.interface.parseLog(log)) ?? []
-      const emittedRevert = logs.find((l) => l?.name === 'UserOpReverted')
-      expect(emittedRevert?.args.reason).to.equal(
-        reverterContract.interface.encodeErrorResult('Error', ['You called a function that always reverts']),
-      )
+      // Error('You called a function that always reverts')
+      const expectedRevertReason =
```
nlordell, line 200: nit: Use ethers to compute this, same as :point_up:.

modules/4337/test/erc4337/ERC4337ModuleNew.spec.ts
```diff
@@ -115,16 +121,33 @@ describe('Safe4337Module - Newly deployed safe', () => {
...
+
+      const newSafeOp = buildSafeUserOpTransaction(
+        safe.address,
+        user1.address,
+        ethers.parseEther('0.5'),
+        '0x',
+        '0',
+        await entryPoint.getAddress(),
+        false,
+        false,
```
nlordell, line 133: These are the defaults right?
```suggestion
```

modules/4337/test/erc4337/ERC4337ModuleNew.spec.ts
```diff
@@ -115,16 +121,33 @@ describe('Safe4337Module - Newly deployed safe', () => {
...
+      await entryPoint.handleOps([userOp], user1.address)
+      const newSignature = buildSignatureBytes([await signSafeOp(user1, await validator.getAddress(), newSafeOp, await chainId())])
+      const newUserOp = buildUserOperationFromSafeUserOperation({
+        safeOp: newSafeOp,
+        signature: newSignature,
+      })
+
+      await expect(entryPoint.handleOps([newUserOp], user1.address))
+        .to.be.revertedWithCustomError(entryPoint, 'FailedOp')
+        .withArgs(0, 'AA25 invalid account nonce')
```
nlordell, line 144: Why is the `newUserOp` needed instead of just trying to replay `userOp`?

modules/4337/test/erc4337/ERC4337ModuleNew.spec.ts
```diff
@@ -143,22 +166,25 @@ describe('Safe4337Module - Newly deployed safe', () => {
         safeOp,
         signature,
       })
-      const transaction = await logGas('Execute UserOp without fee payment', entryPoint.executeUserOp(userOp, 0))
-      const receipt = await transaction.wait()
-      const logs = receipt.logs.map((log) => entryPoint.interface.parseLog(log))
-      const emittedRevert = logs.some((log) => log?.name === 'UserOpReverted')
-      expect(emittedRevert).to.be.true
-      expect(await safe.isDeployed()).to.be.true
+      const expectedReturnData =
```
nlordell, line 169: nit: Same as :point_up: - use `encodeErrorResult`.

modules/4337/test/erc4337/ERC4337ModuleNew.spec.ts
```diff
@@ -234,14 +262,16 @@ describe('Safe4337Module - Newly deployed safe', () => {
...
-      const receipt = await transaction.wait()
-      const logs = receipt.logs.map((log) => entryPoint.interface.parseLog(log))
-      const emittedRevert = logs.find((log) => log?.name === 'UserOpReverted')
-      expect(emittedRevert?.args.reason).to.equal(
-        reverterContract.interface.encodeErrorResult('Error', ['You called a function that always reverts']),
-      )
-      expect(await safe.isDeployed()).to.be.true
+
+      // Error('You called a function that always revert')
+      const expectedRevertReason =
```
nlordell, line 267: Same as :point_up:

nlordell (commented):
Looks good. Mostly small nits.

I also suggested changing from `maxFeePerGas: 0` to `entryPoint.depositTo(safe.address)` to prefund an account without requiring a prefund transfer, but this is a "soft" suggestion, so don't feel like you need to implement it if you disagree.

modules/4337/test/erc4337/ERC4337ModuleExisting.spec.ts
mmv08: Don't know what happened here

modules/4337/test/erc4337/ERC4337ModuleExisting.spec.ts
mmv08: Hmm, I might've understood `without fee` too literally...

modules/4337/test/erc4337/ERC4337ModuleNew.spec.ts
mmv08: Because the other user operation contains `initCode` and it will revert with an error "sender already constructed" and I don't think this is what we're trying to validate here

nlordell: :bulb:

modules/4337/test/erc4337/ERC4337ModuleExisting.spec.ts
```diff
@@ -78,19 +89,24 @@ describe('Safe4337Module - Existing Safe', () => {
...
+      expect(await ethers.provider.getBalance(randomAddress)).to.be.eq(ethers.parseEther('0.5'))
+      await expect(entryPoint.handleOps([userOp], user1.address))
+        .to.be.revertedWithCustomError(entryPoint, 'FailedOp')
+        .withArgs(0, 'AA25 invalid account nonce')
     })
 
     it('should execute contract calls with fee', async () => {
       const { user1, safe, validator, entryPoint } = await setupTests()
+      const randomAddress = ethers.Wallet.createRandom().address
+      const randomAddress2 = ethers.Wallet.createRandom().address
```
nlordell, line 102: Suggestion: Might make sense to name the fee beneficiary, so the test is a bit more self-explanatory:

```suggestion
      const feeBeneficiary = ethers.Wallet.createRandom().address
      const randomAddress = ethers.Wallet.createRandom().address
```

modules/4337/test/erc4337/ERC4337ModuleNew.spec.ts
```diff
@@ -143,22 +164,23 @@ describe('Safe4337Module - Newly deployed safe', () => {
...
+      const expectedReturnData = validator.interface.encodeErrorResult('Error(string)', ['Execution failed'])
+      await expect(entryPoint.handleOps([userOp], user1.address))
+        .to.emit(entryPoint, 'UserOperationRevertReason')
+        .withArgs(await entryPoint.getUserOpHash(userOp), userOp.sender, 0, expectedReturnData)
     })
 
     it('should execute contract calls with fee', async () => {
       const { user1, safe, validator, entryPoint } = await setupTests()
+      const randomAddress = ethers.Wallet.createRandom().address
+      const randomAddress2 = ethers.Wallet.createRandom().address
```
nlordell, line 176: Same rename suggestion as above.

modules/4337/test/erc4337/ERC4337ModuleNew.spec.ts
```diff
@@ -234,14 +258,14 @@ describe('Safe4337Module - Newly deployed safe', () => {
...
-      const transaction = await logGas('Execute UserOp without fee payment', entryPoint.executeUserOp(userOp, 0))
-      const receipt = await transaction.wait()
-      const logs = receipt.logs.map((log) => entryPoint.interface.parseLog(log))
-      const emittedRevert = logs.find((log) => log?.name === 'UserOpReverted')
-      expect(emittedRevert?.args.reason).to.equal(
-        reverterContract.interface.encodeErrorResult('Error', ['You called a function that always reverts']),
-      )
-      expect(await safe.isDeployed()).to.be.true
+
+      // Error('You called a function that always reverts')
```
nlordell, line 262: Comment feels superfluous

```suggestion
```

modules/4337/test/erc4337/ERC4337ModuleNew.spec.ts
nlordell: On second thought, do you think it makes sense to `await safe.deploy()` so that there is no init code and really make this test "you can't replay a user operation" (which I feel was the original intent of the test)?

Not blocking PR, and I'll leave it up to you whether or not to incorporate this suggestion.

nlordell (approved):
LGTM, left a couple of suggestions that you can decide if you want to incorporate or not (as I don't feel strongly either way).

modules/4337/test/erc4337/ERC4337ModuleNew.spec.ts
mmv08: I didn't know about this method, implemented it here and optimized a few other tests

PR #273, opened 2024-02-29
Add an E2E test for nested Safes in 4337 context

### LINKED ISSUE #251: Evaluate Nested Safe Calling Convention with ERC-4337
## Context / issue

Nested Safe ownership structures are problematic in the context of 4337 user operations because of storage access restrictions. However, it should technically possible to accomplish at a UX cost, by creating a signing scheme that allows for a "leaf" Safe to execute the user operation, instead of the "root" Safe.

## Proposed solutions

Specifically, imagine an acyclic Safe ownership tree, where we want to execute a transaction with the root. The idea would be to pick a path to a "leaf" Safe (that is, a Safe with only EOA owners). Since we assume an acyclic Safe ownership tree, this must always be possible. The idea would then be to have the "leaf" Safe sign a user operation to execute `parent.execTransaction(parent.execTransaction(...(root.execTransaction(payload))))` on its parent and make use of "approved hash from executor" signatures (i.e. v = 1). All other Safes/owners sign regular SafeTransactions (i.e. NOT 4337 user operation signature) for executing the root transaction.

The problem with this scheme is that the UX is not great:
* You only know what signatures are needed once you know the signers, so signature collection is super messy (for example, imagine a 2 of 3 Safe where two owners are EOAs and one is a Safe, you can't know what kind of signatures to collect without knowing who the signers will be).
* The user operation would appears as if it was from the "leaf" Safe in "4337-user-operation-explorers".

## Alternatives

- #265 

## Expected Outcome

A PoC for nested Safe user operation execution using the proposed solution.

mmv08:
This PR adds an E2E test case for nested Safe transaction execution in the 4337 Context. The test uses the following ownership structure:
image

I implemented helper classes representing the Safe ownership tree with two node types: Safe and EOA. The tests also required functions to find an execution path and build the user operation from the execution path. The alternative would've been to do everything imperatively, but using those functions, we could theoretically represent arbitrary ownership structure, find an execution path (currently, the function doesn't find the most optimal one) and generate a user operation.

The current algorithm works as follows (thanks to @nlordell for the help):
1. **Finding an execution path:** It does a DFS on the tree until it finds the first Executor node. We consider an executor node a leaf safe owned by an EOA. In the context of the ownership structure on the screenshot, the executor would be the leftmost Leaf Safe. The execution path is represented by an array ordered from the root node to the executor node.
2. **Building the user operation:** Using the execution path and our desired `SafeTransaction` (the one executed with `execTransaction`), we recursively sign the transaction with all the node owners in the execution path. After that, we re-encode the `SafeTransaction` for the next node with a call to the preceding node in the execution path and do the signing again.  It stops when we reach the starting node (the node before the executor 4337 Safe). After that, we encode and send the SafeTransaction in a User Operation. 

- Implements #251 
- I also had to bring quite some code for executing regular Safe transactions from the `safe-smart-account` repo

I tried to document the code as much as possible, but somehow, I feel it still wasn't enough - please feel free to point me to the parts that are unclear, and I'll do my best to explain them with comments.

I'll quote the Slack message from @nlordell that describes the idea implemented in this PR:

> As you may already be aware, nested Safe ownership structures are problematic in the context of 4337 user operations because of storage access restrictions. However, I think it is technically possible to accomplish at a UX cost.
> Specifically, imagine an acyclic Safe ownership tree, where we want to execute a transaction with the root. The idea would be to pick a path to a "leaf" Safe (that is, a Safe with only EOA owners). Since we assume an acyclic Safe ownership tree, this must always be possible. The idea would then be to have the "leaf" Safe sign a user operation to execute parent.execTransaction(parent.execTransaction(...(root.execTransaction(payload)))) on its parent and make use of "approved hash from executor" signatures (i.e. v = 1). All other Safes/owners sign regular SafeTransactions (i.e. NOT 4337 user operation signature) for executing the root transaction.
> The problem with this scheme is that the UX is not great:
> - You only know what signatures are needed once you know the signers, so signature collection is super messy (for example, imagine a 2 of 3 Safe where two owners are EOAs and one is a Safe, you can't know what kind of signatures to collect without knowing who the signers will be). There may be a way to work around this by expanding the Safe4337Module to allow SafeUserOperation signatures to be used for calling 4337-enabled parent Safes, but I would need to think about it a bit more...
> - The user operation would appears as if it was from the "leaf" Safe in 4337-user-operation-explorers.

### DIFF
--- a/modules/4337/src/utils/execution.ts
+++ b/modules/4337/src/utils/execution.ts
@@ -7,6 +7,22 @@ export const EIP_DOMAIN = {
   ],
 }
 
+export const EIP712_SAFE_TX_TYPE = {
+  // "SafeTx(address to,uint256 value,bytes data,uint8 operation,uint256 safeTxGas,uint256 baseGas,uint256 gasPrice,address gasToken,address refundReceiver,uint256 nonce)"
+  SafeTx: [
+    { type: 'address', name: 'to' },
+    { type: 'uint256', name: 'value' },
+    { type: 'bytes', name: 'data' },
+    { type: 'uint8', name: 'operation' },
+    { type: 'uint256', name: 'safeTxGas' },
+    { type: 'uint256', name: 'baseGas' },
+    { type: 'uint256', name: 'gasPrice' },
+    { type: 'address', name: 'gasToken' },
+    { type: 'address', name: 'refundReceiver' },
+    { type: 'uint256', name: 'nonce' },
+  ],
+}
+
 export const EIP712_SAFE_MESSAGE_TYPE = {
   // "SafeMessage(bytes message)"
   SafeMessage: [{ type: 'bytes', name: 'message' }],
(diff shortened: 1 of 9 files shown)


modules/4337/src/utils/execution.ts
```diff
@@ -28,44 +44,85 @@ export interface SafeTransaction extends MetaTransaction {
...
+      const dynamicPartWithLength = `${dynamicPartLength}${sig.data.slice(2)}`
+
+      signatureBytes += staticSignature
+      dynamicBytes += dynamicPartWithLength
+    } else {
+      signatureBytes += sig.data.slice(2)
+    }
+  }
+
+  return signatureBytes + dynamicBytes
```
mmv08, line 125: This whole file is a copy-paste from the core safe contracts repo, which is suboptimal. The issue is that the Safe Contracts repo uses ethers v5, and we use ethers v6, and they're incompatible. Perhaps we should consider releasing a package with 1.4.1 contracts and ethers v6.

nlordell: We can add a tech debt item for the next sprint. We want to start working on Safe account contracts in the near future, so the work would be strategic as well.

modules/4337/test/e2e/4337NestedSafe.spec.ts
```diff
@@ -0,0 +1,428 @@
...
+const recursivelySignHashWithASafe = async (safe: SafeOwnershipTreeNode, hash: string, chainId: BigNumberish): Promise<SafeSignature> => {
+  const signatures = []
+
+  for (const owner of safe.owners) {
+    if (owner instanceof SafeOwnershipTreeEoaNode) {
+      const safeMsgHash = calculateSafeMessageHash(safe.value.address, hash, chainId)
+      signatures.push(await signHash(owner.value, safeMsgHash))
+    } else {
+      // When Safe contract validates the signatures, it uses the pre-image of the hash to validate the signature (legacy EIP-1271 behaviour),
+      // hence we're computing the pre-image of the hash here.
```
nlordell, line 130: Can't wait for Safe 1.5.0

nlordell (approved):
Nice. The algorithm wasn't too hard to follow IMO.

mmv08:
> Nice. The algorithm wasn't too hard to follow IMO.

I guess such an algorithm isn't an issue for someone coming from Cowswap, but it was definitely tricky to write :smile:

modules/4337/src/utils/execution.ts
mmv08: > We can add a tech debt item for the next sprint. We want to start working on Safe account contracts in the near future, so the work would be strategic as well.

This would require some planning because we already upgraded the repo with ethers v6 in the upcoming 1.5.0. The question we'd need to answer is if we need to release an incremental version of 1.4.1 with this upgrade or just wait for 1.5.0.

PR #298, opened 2024-03-04
Introduce P-256 Verifier Contract

### LINKED ISSUE #286: Introduce FCL P-256 Verifier Contract
Introduce a P-256 elliptic curve verifier contract following the P-256 interface. This corresponds to the `P256Verifier` contract from #223.

Since the code has already been implemented, the expected outcome is to first port the code to the new workspace package, author tests that specifically check the P-256 verifier contract implementation with high coverage (whether or not we want 100% coverage on FCL/Daimo.eth code is up for discussion, but my intuition is "no").

Note that we **do not** want to test the actual elliptic curve math implementation, only the interface to it.

nlordell:
Fixes #286 

This PR introduces the P-256 verifier contract first introduced in #223 into the newly created `passkey` project.

The verifier contract follows the exact interface for EIP-7212, so that it is compatible with the precompile.

Added tests that verify functionality and the interface. The `FCLP256Verifier` has 100% coverage.


mmv08 (approved):

modules/passkey/contracts/verifiers/IP256Verifier.sol
```diff
@@ -0,0 +1,121 @@
...
+     */
+    fallback(bytes calldata input) external returns (bytes memory output);
+}
+
+/**
+ * @title P-256 Elliptic Curve Verifier.
+ * @dev P-256 verifier contract that follows the EIP-7212 EC verify precompile interface. For more
+ * details, refer to the EIP-7212 specification: <https://eips.ethereum.org/EIPS/eip-7212>
+ */
+library P256VerifierLib {
```
mmv08, line 41: Question: in https://github.com/safe-global/safe-modules/pull/223, I made the verifier wrapper contract an external contract instead of a library to avoid repeating this code in every signer contract and deployed a singleton rather. Have we decided to switch the approach with this PR?

mmv08: While deploying this code as a singleton still could be achieved by deploying a WebAuthnVerifier singleton contract, I think standalone `P256Signers` that do not use webauthn can benefit from it gas-wise

mmv08: Or I'm having a brainfart that doesn't affect anything significantly..

nlordell: > While deploying this code as a singleton still could be achieved by deploying a WebAuthnVerifier singleton contract, I think standalone P256 signers that do not use webauthn can benefit from it gas-wise

The problem is that the P-256 precompile won't have this and need an intermediary contract.

Also, I think this ends up working exactly like #223, as `WebAuthnVerifier` inherits from `P256Wrapper` which calls `P256Verifier`. Here, a potential `WebAuthnVerifier` will import and call `P256VerifierLib` which calls `P256Verifier`. In both cases, the `WebAuthnVerifier` is the one that contains the code for checking malleability and will be a singleton in both cases.

Issue #270, opened 2024-02-23
Update Examples and Gas Metering to be 0.7 Compatible


nlordell:
Examples and 4337 gas metering use EntryPoint v0.6 compatible module, we should update the code to work with the latest EntryPoint.

nlordell:
The example was updated in #303 and gas metering is covered by #294 and #316

PR #300, opened 2024-03-05
Fix E2E Tests For ERC-4337 v0.7 Bundler

### LINKED ISSUE #227: Update end-to-end tests with 0.7.0 entrypoint & corresponding bundler
For the upcoming release of 0.7.0 entrypoint compatibility, we want our end-to-end tests to use the 0.7.0 entrypoint and a bundler version supporting it.

Deliverables:
- Refactoring end-to-end tests to use the entrypoint and bundler 0.7.0

Potential blockers:
- Bundler image/entrypoint may be unavailable due to undergoing security audit

nlordell:
Fixes #227 

This PR updates the E2E tests to work with the latest bundler for the EntryPoint contract v0.7. Note that there was a notable change in that the `PackedUserOperation` struct is only used on-chain, and the bundler RPC uses a different `UserOperation` type that is completely unpacked. This required some additional helper methods as well as some adjustments to tests to distinguish between the packed and unpacked versions of user operations.

Also, unfortunately the reference bundler repository is not tagged yet - so we are using a commit hash that corresponds to the latest version. I asked the 4337 team to create a new tag, at which point, I will update the docker file. Additionally, the latest bundler version uses submodules, so some dockerfile tweaking was required.

Note that the broken CI is expected until #270 is fixed.

### DIFF
--- a/modules/4337/docker/bundler/Dockerfile
+++ b/modules/4337/docker/bundler/Dockerfile
@@ -1,9 +1,11 @@
 FROM docker.io/library/node:18
 
-ARG TAG=v0.6.1
+# v0.7.0
+ARG TAG=26e4f4c
 RUN git clone https://github.com/eth-infinitism/bundler /src/bundler
 WORKDIR /src/bundler
 RUN git checkout ${TAG}
+RUN git submodule init && git submodule update
 
 RUN yarn && yarn preprocess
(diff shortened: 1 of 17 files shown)


mmv08 (approved):

mmv08:
CI is broken though, it looks like the TS files failed to compile.

PR #301, opened 2024-03-05
`WebAuthnLib` tests

### LINKED ISSUE #287: Introduce WebAuthn Signature Verifier
This issue captures the work to introduce a WebAuthn signature verifier implementation that uses the P-256 verifier introduced in #286 to validate Safe signatures.

An implementation already exists in #223, so the expected outcome of this issue is to port the existing code to the new package, and testing of the WebAuthn signature verification logic.

akshay-ap:
The PR was initially created to fix #287; however, https://github.com/safe-global/safe-modules/pull/320 heavily refactored the WebAuthn flow, and the PR is now adjusted to test the corresponding library from the #320 PR.

Changes in this PR:
- Add unit tests for the `WebAuthnLib` contract.

Tests covered:
- Return `false` when flags don't match.
- Return `true` when the verifier returns `true`
- Returns `false` when the verifier returns `false`
- Check that the signing message is correctly generated

The tests use the mock contract.


remedcu (approved):

modules/passkey/test/verifiers/WebAuthnVerifier.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+import hre from 'hardhat'
+
+describe('WebAuthnVerifier', () => {
+  const setupTests = deployments.createFixture(async ({ deployments }) => {
+    const { WebAuthnVerifier } = await deployments.fixture()
+    const webAuthnVerifier = await ethers.getContractAt('WebAuthnVerifier', WebAuthnVerifier.address)
+    return { webAuthnVerifier }
+  })
+
+  it('Should return false when invalid signature', async () => {
```
mmv08, line 12: In this tests there are mutliple fields that are invalid:
- authenticator data
- clientDataFields
- (not sure here) x, y and r, s values are all zero, not sure if that's supported by the standard

Therefore it's unclear why the test fails. It may be because of the signature or because of any other field

akshay-ap: here, `WebAuthnVerifier` returns `false` in either of bloew:
a. Authenticator flag did not match
b. P256VerifierLib returned `false`

The test case sets `a` correctly so only case to have `false` return is `b`.

mmv08: This is not clear from the test description, I'd adjust it to `Should return false when the verifier returns false` in this case

modules/passkey/test/verifiers/WebAuthnVerifier.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+        0n,
+        0n,
+      ),
+    ).to.be.false
+  })
+
+  it('Should return true on successful signature checks', async () => {
+    const mockP256Verifier = await (await hre.ethers.getContractFactory('MockContract')).deploy()
+    await mockP256Verifier.givenAnyReturnBool(true)
+    const webAuthnVerifier = await (await hre.ethers.getContractFactory('WebAuthnVerifier')).deploy(mockP256Verifier.target)
```
nlordell, line 83: We have a "dummy" verifier that always returns `true`, why not use that here instead (avoids the additional dependency)?

modules/passkey/test/libraries/P256.spec.ts
```diff
@@ -20,7 +20,7 @@ describe('P256', function () {
   it('Should return true on valid signature', async function () {
     const { verifier, p256Lib, account } = await setupTests()
 
-    const message = ethers.id('hello world')
+    const message = ethers.id('hello passkeys')
```
mmv08, line 23: I accidentally opened this file instead of WebAuthn tests and changed `world` to `passkeys` for fun, and that led to discovering https://github.com/safe-global/safe-modules/pull/338. I can revert the change but IMO, it looks more fun like this :stuck_out_tongue:

modules/passkey/contracts/test/MockContractImport.sol
```diff
@@ -0,0 +1,5 @@
+// SPDX-License-Identifier: LGPL-3.0-only
+/* solhint-disable no-global-import */
+pragma solidity >=0.8.0;
+
+import "@safe-global/mock-contract/contracts/MockContract.sol";
```
nlordell, line 5: Since we import the `MockContract`, we sould probably rewrite the P-256 verifier tests to use it instead of the `BadP265Verifier` contract.

nlordell: (not in this PR - just a side note).

modules/passkey/contracts/test/TestWebAuthnLib.sol
```diff
@@ -0,0 +1,68 @@
...
+     * @notice Generate a signing message based on the authenticator data, challenge, and client
+     * data fields.
+     * @dev The signing message are the 32-bytes that are actually signed by the P-256 private key
+     * when doing a WebAuthn credential assertion. Note that we verify that the challenge is indeed
+     * signed by using its value to compute the signing message on-chain.
+     * @param challenge The WebAuthn challenge used for the credential assertion.
+     * @param authenticatorData Authenticator data.
+     * @param clientDataFields Client data fields.
+     * @return message Signing message.
+     */
```
nlordell, line 18: nit: why natspec for test contracts? Feels like overkill, and can potentially get out of sync over time.

modules/passkey/test/libraries/WebAuthn.spec.ts
```diff
@@ -0,0 +1,100 @@
...
+import { DUMMY_CLIENT_DATA_FIELDS, DUMMY_SIGNATURE_BYTES, base64UrlEncode, getSignatureBytes } from '../utils/webauthn'
+
+describe('WebAuthn Library', () => {
+  const setupTests = deployments.createFixture(async ({ deployments }) => {
+    const { FCLP256Verifier } = await deployments.fixture()
+    const verifier = await ethers.getContractAt('FCLP256Verifier', FCLP256Verifier.address)
+
+    const WebAuthnLibFactory = await ethers.getContractFactory('TestWebAuthnLib')
+    const deployedWebAuthnLib = await WebAuthnLibFactory.deploy()
+    const webAuthnLib = await ethers.getContractAt('TestWebAuthnLib', await deployedWebAuthnLib.getAddress())
```
nlordell, line 13: Why is this needed?
```suggestion
    const webAuthnLib = await WebAuthnLibFactory.deploy()
```

modules/passkey/contracts/test/TestWebAuthnLib.sol
```diff
@@ -0,0 +1,68 @@
...
+    function verifySignatureCastSig(
+        bytes32 challenge,
+        bytes calldata signature,
+        WebAuthn.AuthenticatorFlags authenticatorFlags,
+        uint256 x,
+        uint256 y,
+        IP256Verifier verifier
+    ) public view returns (bool success) {
+        success = verifySignature(challenge, WebAuthn.castSignature(signature), authenticatorFlags, x, y, verifier);
```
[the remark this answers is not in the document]
mmv08: What if someone adjusts it so it doesn't call it anymore? I think covering only one method would mean depending too much on the implementation details.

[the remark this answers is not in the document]
nlordell: But we would see that in the coverage report right?

modules/passkey/contracts/test/TestWebAuthnLib.sol
mmv08: Copy and paste from the lib code. Will remove

modules/passkey/contracts/test/TestWebAuthnLib.sol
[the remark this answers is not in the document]
mmv08: Ooh good catch! Adjusted

modules/passkey/test/libraries/WebAuthn.spec.ts
nlordell: Was this intentially resolved?

modules/passkey/test/libraries/WebAuthn.spec.ts
```diff
@@ -0,0 +1,100 @@
...
+    const mockP256Verifier = await (await hre.ethers.getContractFactory('MockContract')).deploy()
+
+    return { webAuthnLib, verifier, mockP256Verifier }
+  })
+
+  it('Should return false when the verifier returns false', async () => {
+    const { webAuthnLib, verifier } = await setupTests()
```
nlordell, line 19: Nit: This is the only test using an actual verifier, and relies on the signature being invalid. Why not just use the Mock Verifier here again?

modules/passkey/test/utils/webauthn.ts
```diff
@@ -357,3 +357,34 @@ export function encodeWebAuthnSignature(response: AuthenticatorAssertionResponse
     [new Uint8Array(response.authenticatorData), clientDataFields, r, s],
```
nlordell, line 357: nit: This should be re-written with `getSignatureBytes`.

modules/passkey/test/libraries/WebAuthn.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+    const signature = {
+      authenticatorData,
+      clientDataFields: DUMMY_CLIENT_DATA_FIELDS,
+      r: 0n,
+      s: 0n,
+    }
+
+    expect(await webAuthnLib.verifySignature(challenge, signature, '0x01', 0n, 0n, verifier.target)).to.be.false
+
+    expect(await webAuthnLib.verifySignatureCastSig(challenge, DUMMY_SIGNATURE_BYTES, '0x01', 0n, 0n, verifier.target)).to.be.false
```
nlordell, line 35: Is it intentional that these are verifying with different inputs?

modules/passkey/test/utils/webauthn.ts
mmv08: done

modules/passkey/test/libraries/WebAuthn.spec.ts
mmv08: done

modules/passkey/test/utils/webauthn.ts
```diff
@@ -345,15 +345,43 @@ export function decodeSignature(response: Pick<AuthenticatorAssertionResponse, '
...
+ */
+export const DUMMY_SIGNATURE_BYTES = ethers.AbiCoder.defaultAbiCoder().encode(
+  ['bytes', 'string', 'uint256', 'uint256'],
+  [
+    `0x${'a0'.repeat(37)}`, // authenticatorData without any extensions/attestated credential data is always 37 bytes long.
+    DUMMY_CLIENT_DATA_FIELDS,
+    `0x${'ec'.repeat(32)}`,
+    `0x${'d5a'.repeat(21)}f`,
+  ],
+)
```
nlordell, line 387: No longer used

```suggestion
```

modules/passkey/test/utils/webauthn.ts
```diff
@@ -345,15 +345,43 @@ export function decodeSignature(response: Pick<AuthenticatorAssertionResponse, '
...
+/**
+ * Encodes the given WebAuthn signature into a string. Used for testing purposes.
+ *
+ * @param authenticatorData - The authenticator data as a Uint8Array.
+ * @param clientDataFields - The client data fields as a string.
+ * @param r - The value of r as a bigint.
+ * @param s - The value of s as a bigint.
+ * @returns The encoded string.
+ */
+export function getSignatureBytes(authenticatorData: Uint8Array, clientDataFields: string, r: bigint, s: bigint): string {
```
nlordell, line 357: ubernit:

Might be nicer to implement this as:

```suggestion
export function getSignatureBytes(signature: { authenticatorData: BytesLike, clientDataFields: string, r: BigNumberish, s: BigNumberish }): string {
```

Since you are always calling this function as:
```ts
getSignatureBytes(signature.authenticatorData, signature.clientDataFields, signature.r, signature.s)
```

nlordell (approved):
LGTM. A couple uber-nits, but fine merging as is.

modules/passkey/test/utils/webauthn.ts
mmv08: This change actually led to failing tests with `TestWebAuthnSigner` :see_no_evil:

mmv08:
@nlordell The suggestion with updating the `encodeWebAuthnSignature` to use `getSignatureBytes` led to a couple of tests failing related to the `TestWebAuthnSigner`. It was related to the `clientDataFields` being a bytes array in the old `WebAuthnSignature` struct. The current webauthnlib made it a string. I fixed it here: https://github.com/safe-global/safe-modules/pull/301/commits/b6f0b1030b5cc094a7822b29a8220f94094e2df9

Could you please take a look to see if that's correct since you have more context than me?

modules/passkey/test/utils/webauthn.ts
```diff
@@ -302,7 +303,7 @@ export function decodeClientDataFields(response: Pick<AuthenticatorAssertionResp
   }
 
   const [, fields] = match
-  return ethers.hexlify(ethers.toUtf8Bytes(fields))
+  return fields
```
nlordell, line 306: Nice catch!

nlordell (approved):

PR #303, opened 2024-03-05
Fix 0.7 Compilation Issues In Passkey Example

nlordell:
Partially addresses #270 

This PR modifies the passkey example to compile once again for the 0.7 EntryPoint changes. Note that the contract addresses are also updated, as well as the network the example is for. This is because Polygon Mumbai is not well supported seeing as it is anchored to Goerli which is deprecated.

Test transaction using the example application: https://sepolia.etherscan.io/tx/0x267fa1cd04ba44701807c10ea6f7885c45c22c3ff6646fb9ef697884d91fedc8


mmv08 (approved):
Haven't tested it but the code looks good

nlordell:
> Haven't tested it but the code looks good

I'm figuring out my passkey setup on Linux so I can test, then will merge once I have a successful test :)

Issue #288, opened 2024-02-29
Implement a Passkey `SignatureValidator` Factory


nlordell:
Passkey signers should be deployed via a factory contract using `CREATE2`. This allows addresses of signers to be deterministically computed, and used as a Safe owner even without being deployed.

This issue captures the work to introduce such a factory, as well as associated unit tests. While a rough sketch for such a factory exists in #223, this was geared towards the `Launchpad` contract which is out of scope for the current passkey implementation epic. Therefore, we may consider changing the interface for this issue.

The expected outcome is a WebAuthn `SignatureValidator` factory, where the created signature validators instances verify signatures using the WebAuthn verifier implementation introduced in #287, as well as associated tests.

PR #306, opened 2024-03-05
Feature: Implement a Passkey `SignatureValidator` factory

mmv08:
This PR is a partial solution for https://github.com/safe-global/safe-modules/issues/288. What needs to be added is a proper test suite for the factory. It only has one migrated end-to-end test from the 4337 module package.

What was done:
- Introduce `ICustomECSignerFactory` and `ICustom256BitECSignerFactory` interfaces for custom signer factories. The first is generic and should work with signing algorithms of any key length, while the second is optimized for EC signing and 256-bit key sizes. The first one is not used in the project and is kept as an example or for future reference.
- Add the `WebAuthnVerifier` and make it use the new P256Library introduced in #298 
- Migrate the `SafeSignerLaunchpad` contract and make it implement the `ICustom256BitECSignerFactory` interface thus making it `Safe256BitECSignerLaunchpad` 
- Migrate and adjust the end-to-end test from the 4337 module for the launchpad

To be done:
- Remove the migrated contracts from the 4337 package


mmv08:
@CodiumAI-Agent /describe

mmv08:
@CodiumAI-Agent /describe

modules/4337/contracts/experimental/WebAuthnSigner.sol
```diff
@@ -5,7 +5,6 @@ pragma solidity >=0.8.0;
 import {SignatureValidatorConstants} from "./SignatureValidatorConstants.sol";
 import {IUniqueSignerFactory} from "./SafeSignerLaunchpad.sol";
 import {SignatureValidator} from "./SignatureValidator.sol";
-import {SignatureValidatorConstants} from "./SignatureValidatorConstants.sol";
```
mmv08, line 8: removed a duplicated import

examples/4337-gas-metering/tsconfig.json
```diff
@@ -13,7 +13,7 @@
     /* Completeness */
     "skipLibCheck": true /* Skip type checking all .d.ts files. */,
     "outDir": "./dist" /* Redirect output structure to the directory. */,
-    "noImplicitReturns": true /* Report error when not all code paths in function return a value. */,
+    "noImplicitReturns": true /* Report error when not all code paths in function return a value. */
   },
-  "exclude": ["node_modules"],
+  "exclude": ["node_modules"]
```
mmv08, line 18: the formatter wanted me to do it

modules/4337/package.json
```diff
@@ -30,8 +31,9 @@
     "lint:ts": "eslint ./src --fix && eslint ./test --fix",
     "fmt": "prettier --write ./contracts/**/*.sol",
     "fmt:check": "prettier --check ./**/*.sol",
-    "prepack": "npm run build",
-    "prepublish": "npx rimraf build && npm run build && npm run build:ts"
+    "prepack": "npm run build:all",
+    "prepublish": "npx rimraf build && npm run build:all",
```
nlordell, line 35: maybe `rimraf build` should be part of `build:sol` script?

modules/4337/package.json
```diff
@@ -76,6 +78,6 @@
     }
   },
   "dependencies": {
-    "@safe-global/safe-contracts": "^1.4.1"
+    "@safe-global/safe-contracts": "^1.4.1-build.0"
```
nlordell, line 81: Out of curiosity, what is the difference?

modules/passkey/contracts/interfaces/ICustomSignerFactory.sol
```diff
@@ -0,0 +1,85 @@
+// SPDX-License-Identifier: LGPL-3.0-only
+pragma solidity >=0.8.0 <0.9.0;
+
+/**
+ * @title ICustomECSignerFactory
+ * @dev Interface for creating and verifying ECDSA signers. This is a generalized interface that should be
+ *  compatible with curves of any order size. Currently not used in the project and exists here for reference.
+ */
+interface ICustomECSignerFactory {
```
nlordell, line 9: nit, this isn't an EC signer - in particular, it would support an interface to passkey with RSA key.

modules/passkey/.env.sample
```diff
@@ -0,0 +1,2 @@
+# Entrypoint address to use for the Safe 4337 module deployment in 4337 tests. Default is the canonical EntryPoint v0.7 address
+TESTS_4337_MODULE_DEPLOYMENT_ENTRY_POINT_ADDRESS="0x0000000071727De22E5E9d8BAf0edAc6f37da032"
```
nlordell, line 2: nit: missing trailing newline.

modules/4337/package.json
mmv08: 1.4.1 pointed to a non-existing js file as the entry file. It didn't affect our build, but the autocomplete suggested this version as the latest one in the passkeys package.json, and when you try to use 1.4.1 and 1.4.1-build.0 together, npm will complain :shrug: 

A full overview of the changes: https://github.com/safe-global/safe-smart-account/compare/v1.4.1...v1.4.1-build.0

modules/passkey/contracts/4337/Safe256BitECSignerLaunchpad.sol
```diff
@@ -0,0 +1,277 @@
...
+        uint256 size;
+        // solhint-disable-next-line no-inline-assembly
+        assembly ("memory-safe") {
+            size := extcodesize(account)
+        }
+        return size > 0;
+    }
+}
+
+interface SafeSetup {
```
nlordell, line 266: nit: Can we move this to an `ISafe` file?

modules/passkey/contracts/interfaces/ICustomSignerFactory.sol
```diff
@@ -0,0 +1,85 @@
...
+        bytes calldata signature,
+        bytes calldata signerData
+    ) external view returns (bytes4 magicValue);
+}
+
+/**
+ * @title ICustom256BitECSignerFactory
+ * @dev Interface for creating and verifying ECDSA signers using 256-bit elliptic curves.
+ */
+interface ICustom256BitECSignerFactory {
```
nlordell, line 46: Seeing as `ICustomECSignerFactory` isn't used at all, I would actually suggest removing it. It feels a bit like we are trying to future proof, even if there is no demand for using passkeys with algorithms other than ECDSA over the P-256 curve.

modules/4337/package.json
nlordell: Thanks!

modules/passkey/contracts/interfaces/ICustomSignerFactory.sol
mmv08: Yeah I mentioned in the `@dev` natspec that it's just kept for the future reference. I can remove it.

modules/passkey/contracts/4337/Safe256BitECSignerLaunchpad.sol
```diff
@@ -0,0 +1,265 @@
...
+import {SafeStorage} from "@safe-global/safe-contracts/contracts/libraries/SafeStorage.sol";
+import {SignatureValidatorConstants} from "../SignatureValidatorConstants.sol";
+import {ICustom256BitECSignerFactory} from "../interfaces/ICustomSignerFactory.sol";
+import {ISafeSetup} from "../interfaces/ISafe.sol";
+
+/**
+ * @title SafeOpLaunchpad - A contract for Safe initialization with custom unique signers that would violate ERC-4337 factory rules.
+ * @dev The is intended to be set as a Safe proxy's implementation for ERC-4337 user operation that deploys the account.
+ */
+contract Safe256BitECSignerLaunchpad is IAccount, SafeStorage, SignatureValidatorConstants {
```
nlordell, line 16: The launchpad was introduced in a way to be general for the 4337 module, but maybe we should make it specific to the passkey feature.

My rationale is that the naming, and how the contracts relate to themselves is a bit confusing, and simplifying the code so that it is exactly what we need instead of trying to be too general would go a long way. I might be completely off here - so correct me if you think I'm wrong, but I don't see us supporting additional signers with this lauchpad contract in the next half year at least, at which point, the smart account landscape might be sufficiently different that this contract no longer makes sense.

modules/passkey/docker-compose.yaml
```diff
@@ -0,0 +1,37 @@
...
+      context: .
+      dockerfile: docker/bundler/Dockerfile
+      args:
+        TAG: 26e4f4c433916a6a2ea8ecc91ba8a56cd904f27f
+    restart: always
+    command: ['--auto', '--network=http://geth:8545']
+    ports:
+      - 3000:3000
+
+  bundler-upstream:
```
nlordell, line 28: nit: not needed for the passkey project IMO.

modules/passkey/docker/bundler/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM docker.io/library/node:18
```
nlordell, line 1: You can `ln-s` this file, so it is always the same as the 4337 file. It is a lightweight way to share the docker file between both projects :shrug:

modules/passkey/hardhat.config.ts
```diff
@@ -1,15 +1,40 @@
...
     sources: 'contracts',
   },
-  solidity: '0.8.24',
+  networks: {
+    localhost: {
+      url: 'http://localhost:8545',
+      tags: ['dev', 'safe'],
+    },
+    hardhat: {
+      blockGasLimit: 100000000,
```
nlordell, line 22: Do we need to increase the block gas limit? That doesn't sound like a good idea.

modules/passkey/contracts/4337/Safe256BitECSignerLaunchpad.sol
mmv08: I'm fine either way. I can rename/adjust it to become a `WebAuthnSingerLaunchpad`, but since the WebAuthnSignerFactory implements the `ICustom256BitECSignerFactory`, I thought it would be OK if the launchpad kept some level of abstraction.

modules/passkey/hardhat.config.ts
```diff
@@ -1,15 +1,40 @@
...
+      tags: ['test'],
+    },
+  },
+  solidity: {
+    compilers: [
+      {
+        version: '0.8.24',
+      },
+      {
+        version: '0.8.23',
```
nlordell, line 34: We can just use 0.8.23 everywhere to simplify the setup as well.

modules/passkey/docker-compose.yaml
mmv08: I don't think it for any contract intended to work with 4337 it hurts to test the compatibility with the upstream

modules/passkey/src/deploy/4337.ts
```diff
@@ -0,0 +1,24 @@
...
+  if (network.tags.dev || network.tags.test) {
+    await deploy('EntryPoint', {
+      from: deployer,
+      contract: EntryPoint,
+      args: [],
+      log: true,
+      deterministicDeployment: '0x90d8084deab30c2a37c45e8d47f49f2f7965183cb6990a98943ef94940681de3',
+    })
+  } else if (!ENTRY_POINT) {
+    throw new Error('TESTS_4337_MODULE_DEPLOYMENT_ENTRY_POINT_ADDRESS must be set')
```
nlordell, line 18: Since this is for testing, do we ever even need this? Might be simpler to always deploy the EntryPoint.

nlordell: Oh... we need it for the launchpad contract. In that case, I think it should be `DEPLOYMENT_ENTRY_POINT`, since when we deploy it for real on Polygon for example, it feels weird to use a `TEST_*` variable.

modules/passkey/hardhat.config.ts
nlordell: I would remove it unless we need it.

modules/passkey/src/deploy/4337.ts
mmv08: good idea, removed

modules/passkey/hardhat.config.ts
nlordell: Maybe it makes sense to include the Safe 4337 module only as an artifact instead of building it for the tests?

nlordell: I'm fine as-is though.

modules/passkey/contracts/4337/Safe256BitECSignerLaunchpad.sol
nlordell: My point is that we can get rid of the interface entirely and just have it use the `WebAuthnSignerFactory` directly.

nlordell: Also I would call it `SafePasskeyLaunchpad` to make sure we capitlize on the `passkey` buzzword :stuck_out_tongue:

modules/passkey/contracts/4337/Safe256BitECSignerLaunchpad.sol
```diff
@@ -0,0 +1,265 @@
...
+    }
+
+    modifier onlySupportedEntryPoint() {
+        require(msg.sender == SUPPORTED_ENTRYPOINT, "Unsupported entry point");
+        _;
+    }
+
+    receive() external payable {}
+
+    function preValidationSetup(bytes32 initHash, address to, bytes calldata preInit) external onlyProxy {
```
nlordell, line 68: One of the reasons I had hoped to introduce this code piecemeal was that we are moving "experimental" code into the "production" directory tree, and there are quite a few things missing (for example in this file, natspec documentation, unit tests, etc).

My personal opinion is that it would have been better to do it that way just to make sure we don't overlook any of these details, which would also have kept the scope of this PR down.

modules/passkey/hardhat.config.ts
mmv08: > I would remove it unless we need it.

Removing it makes the tests fail:

```
1) P256VerifierLib
       Should return true on valid signature:
     ProviderError: Transaction gas limit is 100000000 and exceeds block gas limit of 30000000
      at EdrProviderWrapper.request (/home/mmv/Projects/Safe/safe-modules/node_modules/hardhat/src/internal/hardhat-network/provider/provider.ts:424:19)
      at async HardhatEthersSigner.sendTransaction (/home/mmv/Projects/Safe/safe-modules/modules/passkey/node_modules/@nomicfoundation/hardhat-ethers/src/signers.ts:125:18)
      at async ContractFactory.deploy (/home/mmv/Projects/Safe/safe-modules/modules/passkey/node_modules/ethers/src.ts/contract/factory.ts:111:24)
      at async /home/mmv/Projects/Safe/safe-modules/modules/passkey/test/verifiers/P256VerfierLib.spec.ts:13:25
      at async /home/mmv/Projects/Safe/safe-modules/modules/passkey/node_modules/hardhat-deploy/src/DeploymentsManager.ts:341:24
      at async Context.<anonymous> (/home/mmv/Projects/Safe/safe-modules/modules/passkey/test/verifiers/P256VerfierLib.spec.ts:21:48)

  2) P256VerifierLib
       Should return false on invalid signature:
     ProviderError: Transaction gas limit is 100000000 and exceeds block gas limit of 30000000
      at EdrProviderWrapper.request (/home/mmv/Projects/Safe/safe-modules/node_modules/hardhat/src/internal/hardhat-network/provider/provider.ts:424:19)
      at async HardhatEthersSigner.sendTransaction (/home/mmv/Projects/Safe/safe-modules/modules/passkey/node_modules/@nomicfoundation/hardhat-ethers/src/signers.ts:125:18)
      at async ContractFactory.deploy (/home/mmv/Projects/Safe/safe-modules/modules/passkey/node_modules/ethers/src.ts/contract/factory.ts:111:24)
      at async /home/mmv/Projects/Safe/safe-modules/modules/passkey/test/verifiers/P256VerfierLib.spec.ts:13:25
      at async /home/mmv/Projects/Safe/safe-modules/modules/passkey/node_modules/hardhat-deploy/src/DeploymentsManager.ts:341:24
      at async Context.<anonymous> (/home/mmv/Projects/Safe/safe-modules/modules/passkey/test/verifiers/P256VerfierLib.spec.ts:31:39)

  3) P256VerifierLib
       Should check for signature signature malleability:
     ProviderError: Transaction gas limit is 100000000 and exceeds block gas limit of 30000000
      at EdrProviderWrapper.request (/home/mmv/Projects/Safe/safe-modules/node_modules/hardhat/src/internal/hardhat-network/provider/provider.ts:424:19)
      at async HardhatEthersSigner.sendTransaction (/home/mmv/Projects/Safe/safe-modules/modules/passkey/node_modules/@nomicfoundation/hardhat-ethers/src/signers.ts:125:18)
      at async ContractFactory.deploy (/home/mmv/Projects/Safe/safe-modules/modules/passkey/node_modules/ethers/src.ts/contract/factory.ts:111:24)
      at async /home/mmv/Projects/Safe/safe-modules/modules/passkey/test/verifiers/P256VerfierLib.spec.ts:13:25
      at async /home/mmv/Projects/Safe/safe-modules/modules/passkey/node_modules/hardhat-deploy/src/DeploymentsManager.ts:341:24
      at async Context.<anonymous> (/home/mmv/Projects/Safe/safe-modules/modules/passkey/test/verifiers/P256VerfierLib.spec.ts:37:48)

  4) P256VerifierLib
       Should return false for misbehaving verifiers:
     ProviderError: Transaction gas limit is 100000000 and exceeds block gas limit of 30000000
      at EdrProviderWrapper.request (/home/mmv/Projects/Safe/safe-modules/node_modules/hardhat/src/internal/hardhat-network/provider/provider.ts:424:19)
      at async HardhatEthersSigner.sendTransaction (/home/mmv/Projects/Safe/safe-modules/modules/passkey/node_modules/@nomicfoundation/hardhat-ethers/src/signers.ts:125:18)
      at async ContractFactory.deploy (/home/mmv/Projects/Safe/safe-modules/modules/passkey/node_modules/ethers/src.ts/contract/factory.ts:111:24)
      at async /home/mmv/Projects/Safe/safe-modules/modules/passkey/test/verifiers/P256VerfierLib.spec.ts:13:25
      at async /home/mmv/Projects/Safe/safe-modules/modules/passkey/node_modules/hardhat-deploy/src/DeploymentsManager.ts:341:24
      at async Context.<anonymous> (/home/mmv/Projects/Safe/safe-modules/modules/passkey/test/verifiers/P256VerfierLib.spec.ts:48:38)```

modules/passkey/hardhat.config.ts
mmv08: Forgot it was an option :facepalm: done

modules/passkey/contracts/4337/Safe256BitECSignerLaunchpad.sol
mmv08: Ah, actually, I wanted to avoid limiting it to passkeys because IMO we shouldn't overlook direct secp256r1 keys support in HSMs of mobile devices. If we make it a passkey factory, we will need a separate factory for plain secp256r1 signers

modules/passkey/contracts/4337/Safe256BitECSignerLaunchpad.sol
mmv08: I concur. It's my bad. Since I'm going on vacation tomorrow, I wasn't sure if I would be able to add a meaningful test suite for the factory and decided to take a shortcut and migrate our existing end-to-end test, which required migrating a bunch of other experimental contracts and thus bloating the scope :slightly_frowning_face: I know that 4337 wasn't a requirement for the sprint, but I believe this migration will be beneficial in the long term because for other projects looking to use Safe with passkeys 4337 support is a must.

> we are moving "experimental" code into the "production" directory tree

I think it's okay since no code should be considered production-ready without an official release. I can also add documentation in a separate PR

nlordell: OK, Don't want to bikeshed too much given the time pressure. Can you add a GH issue to clean up some of the experimental code (like adding natspec comments, etc.) so we can schedule it for next week's sprint?

modules/passkey/src/deploy/4337.ts
mmv08: Ah, I didn't read the second comment :facepalm: will restore it

modules/passkey/contracts/4337/Safe256BitECSignerLaunchpad.sol
mmv08: Done here: https://github.com/safe-global/safe-modules/issues/308

mmv08:
I'm not sure why e2e tests fail; they work on my machine (including with a fresh dependency installation). I will debug it later today,

nlordell:
Maybe related to fresh vs cached docker compose image?

nlordell (approved):

modules/passkey/hardhat.config.ts
mmv08: > Maybe it makes sense to include the Safe 4337 module only as an artifact instead of building it for the tests?

So TIL that if you use artifacts directly that will not produce hardhat artifacts and you won't be able to use ethers.getContractAt(name) to get the contract. https://github.com/safe-global/safe-modules/pull/306/commits/c7c79a668e24cef46c250a19900cf761d01b1a11

mmv08:
> Maybe related to fresh vs cached docker compose image?

I fixed the first one (I could reproduce it in the end; it required removing all the build files and reinstalling dependencies from scratch), but now there's a new one with a `SERVER_ERROR` I can't seem to reproduce: https://github.com/safe-global/safe-modules/actions/runs/8177290950/job/22358636900?pr=306. I'm not sure what it can be :shrug:

Issue #285, opened 2024-02-29
Support Multiple EC Verifier Implementations


nlordell:
The WebAuthn signer implementation should support multiple elliptic curve (EC) verifier implementations. Ideally, we would be able to test against the EIP-7212 precompile, but that is currently not available in Harhat. Instead, we can use both the Daimo.eth and an FCL-based verifier to check that signature validation functions in the same way for both verifier implementations.

The expected outcome is the required code and tests ensuring that the EC verfier implementation can be changed. Since this is already the design introduced in #223, most of the work will be around testing.

For bonus points, it would be nice to also include a test using the EIP-7212 P-256 precompile as a verifier, if we can find a way to run a chain that supports this precompile in CI.

nlordell:
It turns out that using an actual precompile is non-trivial at the moment. I checked out some execution clients and they either:
- Don't support the precompile at all ATM (Optimism for example)
- Don't enable it in Dev mode ([Polygon](https://github.com/maticnetwork/bor/issues/1182) for example)

PR #307, opened 2024-03-06
Verify Multiple P-256 Verifiers Are Supported

nlordell:
Partially addresses #285

This PR adds new tests that verify multiple P-256 verifiers are supported by our code. The `webauthn` shim code now moved to the `passkey` package and is imported by the `4337` package, as this made more sense to me. @mmv08 moved E2E tests for Passkey+4337 to the `passkey` project, so once they both merge, the dependency can be removed.

Note that for now, we use a `TestWebAuthnSignerFactor` contract just for implementing the tests, but it should be switched to using the canonical signing factory from #306 once merged.

The additional verifier that is being used is the one from [Daimo-eth](https://github.com/daimo-eth/p256-verifier). The artifact was vendored into the project, as it is a Foundry project, and this is the easiest way to include the dependency without building it (giving us identical bytecode to what is deployed on-chain).

Adding a test for using the precompile will be done in a separate PR.

---

Note that this PR contains some unrelated changes to the deployment and E2E setup. This was a result of rebasing onto changes introduced in #306 to get things working as well as some nits leftover from the aforementioned PR.


modules/passkey/test/utils/webauthn.ts
```diff
@@ -15,11 +15,7 @@ export interface CredentialCreationOptions {
   publicKey: PublicKeyCredentialCreationOptions
 }
 
-export enum UserVerificationRequirement {
-  'required',
-  'preferred',
-  'discouraged',
-}
+export type UserVerificationRequirement = 'required' | 'preferred' | 'discouraged'
```
nlordell, line 18: To match the web interface.

modules/passkey/test/utils/webauthn.ts
```diff
@@ -145,27 +140,26 @@ export class WebAuthnCredentials {
...
-          [
-            ethers.sha256(ethers.toUtf8Bytes(publicKey.rp.id)),
-            0x40 + userVerificationFlag, // flags = attested_data + user_present
-            0, // signCount
-            `0x${'42'.repeat(16)}`, // aaguid
-            ethers.dataLength(credential.id),
-            credential.id,
-            credential.cosePublicKey(),
-          ],
+      authData: b2ab(
```
nlordell, line 149: This was a small bug that existed before but went unnoticed as the previous version of the CBOR decoding library used in the tests accepted both types here. Essentially, we need the field to be an `ArrayBuffer` and not a `Uint8Array`. We got slightly different versions of the library when installing the dependencies in this package, which are more pedantic than their predecessors, and led to this bug being found.

remedcu (approved):

PR #310, opened 2024-03-07
Rename Workspace Packages

nlordell:
This PR just renames some workspace packages to be (hopefully) more consistent:

- `safe-4337-passkey` example is now `4337-passkey`. Since the path was `safe-global/safe-modules/examples/safe-4337-passkey`, it felt like the last `safe` was already implicit...
- Changed the `package.json` names to include the `safe-modules-example-` prefix (kind of like scoping). Since they are private packages, it doesn't really matter.
- `safe-erc4337` -> `safe-4337`, it is more succinct and IMO the ERC is superfluous in the package name.

Additionally, I added the missing example to the root `README`.


remedcu (approved):
Except the CI, rest LGTM 👍🏾

PR #314, opened 2024-03-07
Shared Local Bundler E2E Setup

nlordell:
This PR adds a new workspace `packages/4337-local-bundler` which exposes a `bin` executable called `4337-local-bundler-test` which runs a local bundler using Docker, and then executes `hardhat` tests using the `@4337` tag.

This allows us to have a homogeneous setup between both the 4337 and passkey repositories (as well as any future modules that we want to be 4337-compliant).

Along the way, I was also able to share 4337 and Safe deployment scripts across the projects when bootstrapping the local development node.


packages/4337-local-bundler/src/bin/test.ts
```diff
@@ -0,0 +1,68 @@
+#!/usr/bin/env node
```
nlordell, line 1: Ported from the original `run.sh` script to JS so that it can be exported as a `bin`.

package.json
```diff
@@ -10,7 +10,8 @@
   ],
   "scripts": {
     "fmt:global": "prettier --write .",
-    "fmt:global-check": "prettier --check ."
+    "fmt:global-check": "prettier --check .",
+    "postinstall": "npm rebuild --skip-scripts @safe-global/safe-4337-local-bundler"
```
nlordell, line 14: This looks weird, bit it is a work around to re-link the `bin` from the workspace package. The issue is that the JS file that is referenced in the `bin` field from the `package.json` is **generated** by the build step, and NPM can't figure out how to link it correctly.

The work-around is to manually re-build the package, this forces NPM to regenerate `bin` links. Note that, by the time the `postinstall` step happens, the package is guarateed to be built, as it is a dependency of other packages in the workspace.

akshay-ap (approved):

PR #315, opened 2024-03-08
Move 4337 Testing Utilies To Local Bundler Package

nlordell:
This PR moves the TS testing utility code for 4337 local bundler tests into the package introduced in #314. This is done in an effort to avoid duplicating the `test/utils/e2e.ts` file that contains setup functions for the local bundler setup that gets deployed from the `safe-4337-local-bundler` package.


package.json
```diff
@@ -4,8 +4,8 @@
   "description": "A monorepo for Safe modules",
   "main": "dist/index.js",
   "workspaces": [
-    "modules/*",
     "packages/*",
+    "modules/*",
     "examples/*"
```
nlordell, line 9: Its LOL, but package ordering matters for the order that scripts are executed :facepalm:

packages/4337-local-bundler/package.json
```diff
@@ -9,14 +9,17 @@
...
   "scripts": {
-    "build": "npx rimraf dist && tsc && chmod +x dist/bin/*.js",
+    "clean": "npx rimraf dist",
+    "build": "tsc && chmod +x dist/bin/*.js",
     "lint": "eslint .",
-    "prepack": "npm run build",
-    "prepublish": "npm run build",
-    "postinstall": "npm run build"
+    "preprepare": "npm run clean",
+    "prepare": "npm run build"
```
nlordell, line 16: Prepare gets run at all the same lifetimes as these scripts. Cleaning and building multiple times within an `npm ci` was causing issues with the build process, where small changes to the dependency graph could make the order in which the build happens change and cause errors.

remedcu (approved):

akshay-ap (approved):

PR #318, opened 2024-03-11
Use assembly in WebAuthnVerifier for Base64 encoding

### LINKED ISSUE #289: Optimize Base-64 Encoding Algorithm
The WebAuthn signing process requires [encoding the `challenge`](https://w3c.github.io/webauthn/#dictionary-client-data) to base64. The "challenge" is a relying party chosen value for replay protection. In the passkey design we use, this is the hash of the data we are verifying for, and assume replay protection already applies to it (for example, the `nonce` is included in the Safe transaction hash which gets verified and ensures replay protection).

The issue is that the challenge is computed as a `bytes32` on-chain, but needs to be encoded in base64 for computing the `clientDataJSON` which is ultimately used in computing the "signing message" (i.e. what is actually signed by the private key).

Currently, the base64 implementation that is used in FCL and in the experimental contracts first:
- Converts the `bytes32` to a `bytes memory` array (thus incurring unnecessary allocations)
- Operates on arbitrarily sized `bytes memory` arrays (thus adding code complexity and marginal runtime costs)
- Returns a `string memory` which needs to be concatenated to an existing `bytes memory` (which incurs a copy loop - at least until `mcopy` is available - which has a non-negligible runtime and codesize cost).

The proposal is to instead compute the base64 incoding in-place with an algorithm specialized for 32 bytes of data, thus avoiding memory allocations and copies, and additional complexity for supporting arbitrary sized arrays.

The expected outcome is a specialized base64 encoding implementation for use in the WebAuthn signature validator contract.

akshay-ap:
This PR implements #289 

The algorithm is implemented as a library instead of the suggested in-place adjustment from the original issue. Akshay's original implementation was doing it in-place (which can still be found in the commit history). While it was efficient code size-wise, but gas-wise, it consumed more gas than FCL's implementation, probably due to some inefficient stack manipulations. I tested it with Solady's [library](https://github.com/vectorized/solady/blob/main/src/utils/Base64.sol) then, and it turned out to be much more efficient even though it wasn't doing the encoding in place. 

The implementation I'm proposing is based on Solady's efficient algorithm. It seems to me that it's not adjustable to in-place encoding due to the fact the algorithm takes advantage of optimizing stack operations by writing X bytes to the scratch space and then writing them to the string at position Y+(I*N) (I=Iteration, Y=start, N=bytes written in an iteration) as a 32-byte word. Because it writes them as a 32-byte word, it’s not suitable for modifying a string in place because it will overwrite the existing string bytes with empty characters. (I'm happy if I'm wrong though)

**Changes in PR**:
- Add `codesize` task
- Add a `Base64Url.sol` library with a Base64Url encoding algorithm that is optimized for encoding `bytes32` data.
- Add tests for the base64 library
- Edit the typescript configuration to include the hardhat.config.ts so all the type extensions are picked up. Necessary to get the project compiled.

### Code size change

This PR:
`WebAuthnSigner 2693 bytes (limit is 24576)`

Default main branch:
`WebAuthnSigner 2817 bytes (limit is 24576)`

### Gas Usage - Needs additional test for benchmarking

**Solady**:
```
  Gas Benchmarking
    WebAuthnSigner
      ⛽ deployment: 550077
      ✔ Benchmark signer deployment cost
      ⛽ verification (FreshCryptoLib): 216563
      ✔ Benchmark signer verification cost with FreshCryptoLib verifier (70ms)
      ⛽ verification (daimo-eth): 347246
      ✔ Benchmark signer verification cost with daimo-eth verifier (116ms)
      ⛽ verification (Dummy): 12713
```

**FCL**:
```
  Gas Benchmarking
    WebAuthnSigner
      ⛽ deployment: 612929
      ✔ Benchmark signer deployment cost
      ⛽ verification (FreshCryptoLib): 217272
      ✔ Benchmark signer verification cost with FreshCryptoLib verifier (70ms)
      ⛽ verification (daimo-eth): 347636
      ✔ Benchmark signer verification cost with daimo-eth verifier (108ms)
      ⛽ verification (Dummy): 13844
```

**This PR**:
```
  Gas Benchmarking
    WebAuthnSigner
      ⛽ deployment: 533456
      ✔ Benchmark signer deployment cost
      ⛽ verification (FreshCryptoLib): 210645
      ✔ Benchmark signer verification cost with FreshCryptoLib verifier (64ms)
      ⛽ verification (daimo-eth): 337941
      ✔ Benchmark signer verification cost with daimo-eth verifier (108ms)
      ⛽ verification (Dummy): 12281
```

**This PR (viaIR)**:
```
  Gas Benchmarking
    WebAuthnSigner
      ⛽ deployment: 544844
      ✔ Benchmark signer deployment cost
      ⛽ verification (FreshCryptoLib): 294440
      ✔ Benchmark signer verification cost with FreshCryptoLib verifier (108ms)
      ⛽ verification (daimo-eth): 335453
      ✔ Benchmark signer verification cost with daimo-eth verifier (117ms)
      ⛽ verification (Dummy): 10352
```


modules/4337/contracts/experimental/verifiers/WebAuthnVerifier.sol
```diff
@@ -102,7 +104,61 @@ contract WebAuthnVerifier is IWebAuthnVerifier, P256Wrapper {
         bytes32 challenge,
         bytes calldata clientDataFields
     ) internal pure returns (bytes32 message) {
-        string memory encodedChallenge = Base64Url.encode(abi.encodePacked(challenge));
+
+        // Result is 44 bytes long because the encoded challenge is 32 bytes long
+        // 4*(32 + 2)/3 = 44 after rounding.
+        string memory table = ENCODING_TABLE;
+        string memory encodedChallenge = new string(44);
```
nlordell, line 111: Instead of a `new string(44)`, you can actually decode this directly into `clientDataJson` by building it with:

```solidity
        bytes memory clientDataJson = abi.encodePacked(
            '{"type":"webauthn.get","challenge":"AAAAAAAA....",',
            clientDataFields,
            "}"
        );
```

nlordell: This avoids the copy loop from `abi.encodePacked`-ing a dynamic string.

akshay-ap: done in https://github.com/safe-global/safe-modules/pull/318/commits/f4c2e88d1d7c757b63ad85cc89b93f1e70a0a6f5

modules/4337/contracts/experimental/verifiers/WebAuthnVerifier.sol
```diff
@@ -102,17 +104,47 @@ contract WebAuthnVerifier is IWebAuthnVerifier, P256Wrapper {
...
+
+            // Store group of 6 bits from the challenge to be encoded. Storing of 6 bits group can be removed but, kept here for readability.
+            let sixBitGroup
+
+            // Iterate over challenge in group of 6 bits, for each 6 bits lookup the ENCODING_TABLE, transform it and store it in the result
+            for {let i := 0} lt(i, 252)
+            {
+                i := add(i, 6)
+            } {
+                sixBitGroup := and(shr(sub(250, i), challenge), 0x3F)
```
akshay-ap, line 135: can remove `sub(250, i)` by using value of `i` such that no `sub` is needed.

akshay-ap, line 135:
```suggestion
            for {let i := 250 } lt(i, 251)
            {
                i := sub(i, 6)
            } {
                sixBitGroup := and(shr(i, challenge), 0x3F)
```

akshay-ap: @nlordell , wdyt?

modules/passkey/contracts/verifiers/WebAuthnVerifier.sol
```diff
@@ -106,15 +107,44 @@ contract WebAuthnVerifier is IWebAuthnVerifier {
...
+            // Store group of 6 bits from the challenge to be encoded. Storing of 6 bits group can be removed but, kept here for readability.
+            let sixBitGroup
+
+            // Iterate over challenge in group of 6 bits, for each 6 bits lookup the ENCODING_TABLE, transform it and store it in the result
+            for {
+                let i := 0
+            } lt(i, 252) {
+                i := add(i, 6)
+            } {
+                sixBitGroup := and(shr(sub(250, i), challenge), 0x3F)
```
akshay-ap, line 138:
```suggestion
            for {let i := 250} lt(i, 251)
            {
                i := sub(i, 6)
            } {
                sixBitGroup := and(shr(i, challenge), 0x3F)
```

nlordell:
The code in question is notable missing unit tests.

akshay-ap:
> ### Gas Usage - Needs additional test for benchmarking
> These tests to be added in #287. So, need to close #287 before merging this PR if benchmarking results are required.

yes, I was thinking to add tests and benchmarking tests in #287 or else we can even do it in this

nlordell:
I think we should add unit tests for the functions to ensure that we have 100% coverage (for example, check that it does not validate a correctly signed signature if the authentication flags aren't correctly set).

This should be different than general benchmarking.

akshay-ap:
> I think we should add unit tests for the functions to ensure that we have 100% coverage (for example, check that it does not validate a correctly signed signature if the authentication flags aren't correctly set).
> 
> This should be different than general benchmarking.

Already added here: https://github.com/safe-global/safe-modules/pull/301/files

modules/4337/contracts/experimental/verifiers/WebAuthnVerifier.sol
```diff
@@ -3,7 +3,6 @@
 pragma solidity >=0.8.0;
```
nlordell, line 3: The `experimental` code will be removed, so I wouldn't update it here as well.

modules/4337/contracts/experimental/verifiers/WebAuthnVerifier.sol
```diff
@@ -102,17 +104,47 @@ contract WebAuthnVerifier is IWebAuthnVerifier, P256Wrapper {
...
     ) internal pure returns (bytes32 message) {
-        string memory encodedChallenge = Base64Url.encode(abi.encodePacked(challenge));
+
         /* solhint-disable quotes */
+        // AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA is placeholder for the encoded challenge
         bytes memory clientDataJson = abi.encodePacked(
             '{"type":"webauthn.get","challenge":"',
-            encodedChallenge,
+            "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA",
             '",',
```
nlordell, line 113: nit: this can be a single string instead of mutiple strings (not sure how it affects the generated assembly though).

modules/4337/contracts/experimental/verifiers/WebAuthnVerifier.sol
```diff
@@ -102,17 +104,47 @@ contract WebAuthnVerifier is IWebAuthnVerifier, P256Wrapper {
...
+            }
+
+            // Load the remaining last 4 bits of challenge that are yet to be encoded and then shift left to add 2 bits at the end to make it a group of 6 bits.
+            sixBitGroup := shl(2, and(challenge, 0x0F))
+            mstore8(resultPtr, mload(add(tablePtr, sixBitGroup)))
+        }
+
         /* solhint-enable quotes */
         message = sha256(abi.encodePacked(authenticatorData, sha256(clientDataJson)));
+
```
nlordell, line 147:
```suggestion
```

modules/passkey/contracts/verifiers/WebAuthnVerifier.sol
```diff
@@ -3,7 +3,6 @@
 pragma solidity >=0.8.0;
 
 import {IP256Verifier, P256VerifierLib} from "./IP256Verifier.sol";
-import {Base64Url} from "../vendor/FCL/utils/Base64Url.sol";
```
nlordell, line 6: We can remove the vendored file now that we don't use it.

modules/4337/contracts/experimental/verifiers/WebAuthnVerifier.sol
```diff
@@ -102,17 +104,47 @@ contract WebAuthnVerifier is IWebAuthnVerifier, P256Wrapper {
...
+            {
+                i := sub(i, 6)
+            } {
+                sixBitGroup := and(shr(i, challenge), 0x3F)
+                mstore8(resultPtr, mload(add(tablePtr, sixBitGroup)))
+                resultPtr := add(resultPtr, 1)
+            }
+
+            // Load the remaining last 4 bits of challenge that are yet to be encoded and then shift left to add 2 bits at the end to make it a group of 6 bits.
+            sixBitGroup := shl(2, and(challenge, 0x0F))
```
mmv08, line 141: I wouldn't mix decimals and hexadecimal in assembly and would stick to one numbering system to ease processing and remove switching contexts

modules/passkey/contracts/verifiers/WebAuthnVerifier.sol
```diff
@@ -106,15 +107,44 @@ contract WebAuthnVerifier is IWebAuthnVerifier {
...
             clientDataFields,
             "}"
         );
+
+        string memory table = ENCODING_TABLE;
+
+        // solhint-disable-next-line no-inline-assembly
+        assembly {
+            // Skip first 32 bytes of the table containing the length
+            let tablePtr := add(table, 1)
```
mmv08, line 125: Why does it add 1 here, and the comment is about skipping 32 bytes?

mmv08:
For some reason, tests fail when run through the coverage command. When run normally, tests pass. I will investigate tomorrow.

modules/passkey/src/types/solc.d.ts
```diff
@@ -0,0 +1,41 @@
...
+    sources: {
+      [fileName: string]: {
+        ast: any
+      }
+    }
+  }
+
+  export function compile(input: string): string
+
+  export function loadRemoteVersion(version: string, callback: (err: Error, solc: any) => void): void
```
nlordell, line 40: nit:lints

modules/passkey/hardhat.config.ts
```diff
@@ -58,7 +59,7 @@ const config: HardhatUserConfig = {
             enabled: true,
             runs: 10_000_000,
           },
-          viaIR: false,
+          viaIR: true,
```
nlordell, line 62: This has some nasty implications for the FCL lib AFAIR. I would not sneak in this change in this PR.

modules/passkey/tsconfig.json
```diff
@@ -8,5 +8,6 @@
     "strict": true,
     "skipLibCheck": true,
     "resolveJsonModule": true
-  }
+  },
+  "include": ["src/**/*.ts", "hardhat.config.ts"]
```
nlordell, line 12: Shouldn't tests be included as well?

modules/passkey/src/types/solc.d.ts
```diff
@@ -0,0 +1,41 @@
+declare module 'solc' {
+  export interface CompilerInput {
```
nlordell, line 2: Fairly sure these types aren't used anywhere - and I think are inaccurate (in particular `CompilerOutput` specifically depends on the `outputSelection` of the input). I would remove them if they aren't used.

modules/passkey/hardhat.config.ts
mmv08: Oops, sorry. Somehow, it made its way after my benchmarking.

modules/passkey/src/types/solc.d.ts
mmv08: They're used in the codesize task; without type declaration for `solc`, the project won't compile. They might be inaccurate since I generated them using the perplexity AI tool - however, I believe they're still safer than just declaring everything as any, like [here](https://github.com/safe-global/safe-modules/blob/main/modules/4337/types/solc.d.ts). I'll take one more look and fix the inaccuracies

modules/passkey/tsconfig.json
mmv08: Have included it now

modules/passkey/src/types/solc.d.ts
mmv08: Ahh, now I see what you mean, you're right

modules/passkey/src/types/solc.d.ts
mmv08: What do you mean by lints? I see no linting errors

akshay-ap:
LGTM

modules/passkey/contracts/libraries/Base64Url.sol
```diff
@@ -0,0 +1,91 @@
...
+                if iszero(sgt(i, 0)) {
+                    break
+                }
+            }
+
+            // Encode the final 4-bit group.
+            // 0x0F is a mask to get the last 4 bits.
+            // The encoded character is stored at memory location 74 (result + 74):
+            // <32byte string length><43byte encoded string>. 74 is the penultimate byte.
+            mstore8(add(result, 74), mload(add(0, shl(2, and(data, 0x0F)))))
```
nlordell, line 83:
```suggestion
            mstore8(add(result, 74), mload(shl(2, and(data, 0x0F))))
```

nlordell (approved):

PR #319, opened 2024-03-11
Refactor Contract Layout

nlordell:
This PR changes the contract layout in the `passkey` package, as well as some small renames. It also changes constants to be used as `library`-ies instead of `contract`s that are inherited (to be more consistent, as both were used previously).

There are almost no logic changes, some notable examples that came with the refactor:
1. Since `SignatureData` needed to be shared by the `WebAuthnSigner` and `WebAuthnSignerFactory`, I created a `WebAuthnSignature` library that declares the stuck and includes the casting logic. This lead to some small changes in the `WebAuthnSignerFactory` notably to work around some "stack too deep" error that didn't happen before.
2. The `WebAuthnConstants` is now `WebAuthnFlags` and I **removed** the `AUTH_DATA_FLAGS_BE` and `AUTH_DATA_FLAGS_BS` because they aren't actual flags :shrug:. Don't know how they got there, but the bits in those flags are officially [reserved for future use](https://www.w3.org/TR/webauthn-2/#sctn-authenticator-data). I also didn't bother adding the `AT` and `ET` flags, since those can't really be verified on-chain anyway and not really relevant from a signature verification perspective (they indicate what additional data is available in the authenticator data blob, but we don't make use of it anyway).


akshay-ap (approved):
lgtm

PR #320, opened 2024-03-12
Refactor WebAuthn Implementation

nlordell:
This PR refactors the WebAuthn implementation into a library instead of split into multiple contracts. This removes the amount of `CALL`s required for verifying a signature.

I did some analysis on the deployment vs. signature verification costs comparing this code, `main` and this change incorporating optimizations from #289 and found this:

| setup                                | deployment | verification (dummy) | break even |
| ------------------------------------ | ---------- | -------------------- | ---------- |
| main (viaIR = false)                 | 373543     | 17789                | 0          |
| main (viaIR = true)                  | 393494     | 16443                | 15         |
| this branch (viaIR = false)          | 612123     | 13835                | 61         |
| this branch (viaIR = true)           | 594058     | 12551                | 43         |
| base64 optimizations (viaIR = false) | 442545     | 12965                | 15         |
| base64 optimizations (viaIR = true)  | 414941     | 11102                | 7          |

Meaning that after 7 signatures, even with the slightly larger deployment costs, we would break even by avoiding the extra `CALL` (noting that ABI-encoding the call to the `WebAuthnVerifier` is not particularly efficient in the first place).


modules/passkey/contracts/WebAuthnSigner.sol
```diff
@@ -14,36 +13,25 @@ import {WebAuthnSignature} from "./libraries/WebAuthnSignature.sol";
...
-                data.authenticatorData,
-                WebAuthnFlags.USER_VERIFICATION,
-                message,
-                data.clientDataFields,
-                data.r,
-                data.s,
-                X,
-                Y
-            );
+    function _verifySignature(bytes32 message, bytes calldata signature) internal view virtual override returns (bool success) {
```
remedcu, line 34: Subjective nit: I resonate more with `isValid` than `success` in this case.

A similar change in `SignatureValidator` as well.

remedcu (approved):
Should we add the bounty email before this is audited? Or adding the email doesn't signify anything until we explcitly say somewhere that these files are also available for bounty hunting?

Above and small nit, rest LGTM 👍🏾

akshay-ap (approved):

nlordell:
> Should we add the bounty email before this is audited?

Nice catch, yes we should.

mmv08 (approved):

PR #321, opened 2024-03-12
Using new `v0.7` for 4337 Gas Metering

remedcu:
This PR intends to use the latest `EntryPoint v0.7` along with new `paymaster`'s from providers like Pimlico, Alchemy, etc.

## Current Blockers
- Alchemy only supports `v0.6`
- Pimlico Paymasters are still not ready for Production

## Changes completed/in-progress
- Removing `goerli` as it being deprecated.
- Using `base-sepolia` as well (`sepolia` sometimes have high gas cost)
- Using new packed Safe User Op with the latest 4337 Safe Module
- Update of `permissionless` of Pimlico

### DIFF
--- a/examples/4337-gas-metering/.env.example
+++ b/examples/4337-gas-metering/.env.example
@@ -2,20 +2,20 @@
 PRIVATE_KEY = "0x..." # Add "0x" to Private Key if not already added.
 
 # Pimlico Values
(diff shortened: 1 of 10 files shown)


remedcu:
A new PR will be created for Pimlico Gas Metering Task.

PR #324, opened 2024-03-12
Add Gas Benchmarking Tests for WebAuthn Signer

nlordell:
This PR adds gas benchmarking tests for the WebAuthn signer. Note that we got rid of the `MultipleVerifiers.spec.ts` file, as it is essentially the new gas benchmarking tests, and there was no need to duplicate the work.

The benchmarks look like this:
```
  Gas Benchmarking
    WebAuthnSigner
      ⛽ deployment: 612123
      ✔ Benchmark signer deployment cost (694ms)
      ⛽ verification (FreshCryptoLib): 219365
      ✔ Benchmark signer verification cost with FreshCryptoLib verifier (172ms)
      ⛽ verification (daimo-eth): 351273
      ✔ Benchmark signer verification cost with daimo-eth verifier (201ms)
      ⛽ verification (Dummy): 13835
```

Note that we include a "dummy" benchmark, this is because gas consumption of P-256 signature verification is unstable, and this allows us to better compare gas characteristics of the `WebAuthn` signing message computation overhead.

Furthermore, we remove the check on whether or not the P-256 verifier has code. This is important as precompiles have no code, so our WebAuthn stuff wouldn't work with precompiles without this change.


modules/passkey/contracts/WebAuthnSignerFactory.sol
```diff
@@ -26,7 +26,7 @@ contract WebAuthnSignerFactory is ICustomECDSASignerFactory {
     function createSigner(uint256 x, uint256 y, address verifier) external returns (address signer) {
         signer = getSigner(x, y, verifier);
 
-        if (_hasNoCode(signer) && _validVerifier(verifier)) {
+        if (_hasNoCode(signer)) {
```
remedcu, line 29: This removal was because if it didn't have any code, the `signer` wouldn't have been created. Right?

remedcu (approved):
One doubt, rest LGTM 👍🏾

modules/passkey/contracts/WebAuthnSignerFactory.sol
nlordell: Updating the PR description.

akshay-ap (approved):

PR #343, opened 2024-03-27
User story: Swap passkey owner

akshay-ap:
Fixes #279 

Changes in PR:

- Add a test case to showcase that a Passkey signer for a Safe can be replaced with new signer. The signer is replaced by an EOA which is also a Safe owner.

Tasks:

- Add test case
- Add comments

PR #344, opened 2024-03-27
User story: Execute userOp with Passkey signer

### LINKED ISSUE #277: [USER STORY] Passkey ERC-4337 User Operation
> As a developer, I want all transactions to be sent by 4337 so I can benefit from the larger 4337 ecosystem

This issue capture the above user story.

The expected outcome is an E2E test that verifies that a ERC-4337 user operation can be signed by a passkey owner and then executed.

akshay-ap:
Fixes #277 

Changes in PR:

- Add a test case to showcase that a Passkey signer can sign a `userOp` and execute transactions through a Safe. 

Tasks:

- Add test case
- Add comments

### DIFF
--- a/modules/passkey/test/userStories/executeUserOpFromPasskeySigner.ts
+++ b/modules/passkey/test/userStories/executeUserOpFromPasskeySigner.ts
@@ -1,0 +1,212 @@
+import { expect } from 'chai'
+import { deployments, ethers } from 'hardhat'
+import { WebAuthnCredentials, decodePublicKey, encodeWebAuthnSignature } from '../utils/webauthn'
+import { Safe4337Module } from '@safe-global/safe-4337/typechain-types/contracts/Safe4337Module'
+import { buildSafeUserOpTransaction, buildPackedUserOperationFromSafeUserOperation } from '@safe-global/safe-4337/src/utils/userOp'
+import { buildSignatureBytes } from '@safe-global/safe-4337/src/utils/execution'
+
+describe('Execute UserOp from Passkey signer [@User story]', () => {
+  const setupTests = deployments.createFixture(async ({ deployments }) => {
+    const { EntryPoint, Safe4337Module, SafeProxyFactory, SafeModuleSetup, SafeL2, FCLP256Verifier } = await deployments.run()
+
+    const [user] = await ethers.getSigners()
+
+    const entryPoint = await ethers.getContractAt('IEntryPoint', EntryPoint.address)
+    const module = (await ethers.getContractAt(Safe4337Module.abi, Safe4337Module.address)) as unknown as Safe4337Module
+    const proxyFactory = await ethers.getContractAt(SafeProxyFactory.abi, SafeProxyFactory.address)
+    const safeModuleSetup = await ethers.getContractAt(SafeModuleSetup.abi, SafeModuleSetup.address)
+    const singleton = await ethers.getContractAt(SafeL2.abi, SafeL2.address)
+    const verifier = await ethers.getContractAt('IP256Verifier', FCLP256Verifier.address)
+
+    const WebAuthnSignerFactory = await ethers.getContractFactory('WebAuthnSignerFactory')
+    const signerFactory = await WebAuthnSignerFactory.deploy()
+
+    const navigator = {
+      credentials: new WebAuthnCredentials(),
+    }
+
+    const credential = navigator.credentials.create({
+      publicKey: {
+        rp: {
+          name: 'Safe',
+          id: 'safe.global',
+        },
+        user: {
+          id: ethers.getBytes(ethers.id('chucknorris')),
+          name: 'chucknorris',
+          displayName: 'Chuck Norris',
+        },
+        challenge: ethers.toBeArray(Date.now()),
+        pubKeyCredParams: [{ type: 'public-key', alg: -7 }],
+      },
+    })
+
+    const publicKey = decodePublicKey(credential.response)
+    await (await signerFactory.createSigner(publicKey.x, publicKey.y, await verifier.getAddress())).wait()
+    const signer = await signerFactory.getSigner(publicKey.x, publicKey.y, await verifier.getAddress())
+
+    return {
+      user,
+      proxyFactory,
+      safeModuleSetup,
+      module,
+      entryPoint,
+      singleton,
+      signerFactory,
+      navigator,
+      verifier,
+      SafeL2,
+      signer,
+      credential,
+    }
+  })
+
+  it('should execute a userOp with WebAuthn signer as owner', async () => {
+    const { user, proxyFactory, safeModuleSetup, module, entryPoint, singleton, navigator, SafeL2, signer, credential } = await setupTests()
+
+    const safeSalt = 1n
+    const initializer = safeModuleSetup.interface.encodeFunctionData('enableModules', [[module.target]])
+
+    const setupData = singleton.interface.encodeFunctionData('setup', [
+      [signer],
+      1n,
+      await safeModuleSetup.getAddress(),
+      initializer,
+      await module.getAddress(),
+      ethers.ZeroAddress,
+      0,
+      ethers.ZeroAddress,
+    ])
+
+    const safe = await proxyFactory.createProxyWithNonce.staticCall(singleton, setupData, safeSalt)
+
+    const deployData = proxyFactory.interface.encodeFunctionData('createProxyWithNonce', [singleton.target, setupData, safeSalt])
+    const initCode = ethers.solidityPacked(['address', 'bytes'], [proxyFactory.target, deployData])
+
+    const safeOp = buildSafeUserOpTransaction(
+      safe,
+      user.address,
+      ethers.parseEther('0.0'),
+      '0x',
+      await entryPoint.getNonce(safe, 0),
+      await entryPoint.getAddress(),
+      false,
+      true,
+      {
+        initCode,
+        verificationGasLimit: 700000,
+        callGasLimit: 2000000,
+        maxFeePerGas: 10000000000,
+        maxPriorityFeePerGas: 10000000000,
+      },
+    )
+
+    const packedUserOp = buildPackedUserOperationFromSafeUserOperation({
+      safeOp,
+      signature: '0x',
+    })
+
+    const opHash = await module.getOperationHash(packedUserOp)
+
+    const assertion = navigator.credentials.get({
+      publicKey: {
+        challenge: ethers.getBytes(opHash),
+        rpId: 'safe.global',
+        allowCredentials: [{ type: 'public-key', id: new Uint8Array(credential.rawId) }],
+        userVerification: 'required',
+      },
+    })
+
+    const signature = buildSignatureBytes([
+      {
+        signer: signer as string,
+        data: encodeWebAuthnSignature(assertion.response),
+        dynamic: true,
+      },
+    ])
+
+    expect(await ethers.provider.getCode(entryPoint)).to.not.equal('0x')
+
+    await user.sendTransaction({ to: safe, value: ethers.parseEther('1') }).then((tx) => tx.wait())
+    expect(await ethers.provider.getBalance(safe)).to.equal(ethers.parseEther('1'))
+    expect(await ethers.provider.getCode(safe)).to.equal('0x')
+
+    await (
+      await entryPoint.handleOps(
+        [
+          {
+            ...packedUserOp,
+            signature: ethers.solidityPacked(['uint48', 'uint48', 'bytes'], [safeOp.validAfter, safeOp.validUntil, signature]),
+          },
+        ],
+        user.address,
+      )
+    ).wait()
+
+    expect(await ethers.provider.getCode(safe)).to.not.equal('0x')
+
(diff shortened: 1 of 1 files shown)


modules/passkey/test/userStories/executeUserOpFromPasskeySigner.ts
```diff
@@ -0,0 +1,224 @@
...
+      await entryPoint.handleOps(
+        [
+          {
+            ...packedUserOp,
+            signature: ethers.solidityPacked(['uint48', 'uint48', 'bytes'], [safeOp.validAfter, safeOp.validUntil, signature]),
+          },
+        ],
+        user.address,
+      )
+    ).wait()
```
nlordell, line 159: Same comment as #343 (pre-req to user story).

akshay-ap: done: [2093cd0](https://github.com/safe-global/safe-modules/pull/344/commits/2093cd0ef983b45eb2877a62f2e304b86e950a7b) [af0d2e7](https://github.com/safe-global/safe-modules/pull/344/commits/af0d2e758112d1d12baa2b2235fead3e997c5503)

modules/passkey/test/userStories/executeUserOpFromPasskeySigner.ts
```diff
@@ -0,0 +1,157 @@
...
+ * Step 3: Execute the userOp.
+ */
+describe('Execute userOp from Passkey signer [@User story]', () => {
+  const setupTests = deployments.createFixture(async ({ deployments }) => {
+    const { EntryPoint, Safe4337Module, SafeProxyFactory, SafeModuleSetup, SafeL2, FCLP256Verifier } = await deployments.run()
+
+    const [user] = await ethers.getSigners()
+
+    const entryPoint = await ethers.getContractAt('IEntryPoint', EntryPoint.address)
+    const module = (await ethers.getContractAt(Safe4337Module.abi, Safe4337Module.address)) as unknown as Safe4337Module
```
nlordell, line 27: Since we have to do this dance in the other 3 PRs, can we instead add this to the `Dependencies.sol` and:
```suggestion
    const module = await ethers.getContractAt('Safe4337Module', Safe4337Module.address)
```

modules/passkey/test/userStories/executeUserOpFromPasskeySigner.ts
```diff
@@ -0,0 +1,157 @@
...
+      ethers.ZeroAddress,
+      0,
+      ethers.ZeroAddress,
+    ])
+
+    // Deploy a Safe with EOA and passkey signer as owners
+    const safeSalt = Date.now()
+    const txReceipt = await (await proxyFactory.createProxyWithNonce(singleton, setupData, safeSalt)).wait()
+    // Get Safe address from the event logs
+    const safeAddress = txReceipt.logs.filter((log: Log) => log.address === proxyFactory.target)[0].args[0]
```
nlordell, line 82: Applies to other PRs as well, but it would be less brittle if we read the address from the proxy factory instead of parsing the logs. In general the logs method here is nice, but it requires some slightly more intimate knowledge of how the proxy factory works.

```suggestion
    const safeAddress = await proxyFactory.createProxyWithNonce.staticCall(singleton, setupData, safeSalt)
    await proxyFactory.createProxyWithNonce(singleton, setupData, safeSalt)
```

modules/passkey/test/userStories/executeUserOpFromPasskeySigner.ts
```diff
@@ -0,0 +1,157 @@
...
+ * The flow can be summarized as follows:
+ * Step 1: Setup the contracts.
+ * Step 2: Create a userOp and sign it using passkey credential.
+ * Step 3: Execute the userOp.
+ */
+describe('Execute userOp from Passkey signer [@User story]', () => {
+  const setupTests = deployments.createFixture(async ({ deployments }) => {
+    const { EntryPoint, Safe4337Module, SafeProxyFactory, SafeModuleSetup, SafeL2, FCLP256Verifier } = await deployments.run()
+
+    const [user] = await ethers.getSigners()
```
nlordell, line 24: nit: In this test, this account acts as a `relayer` and not a `user`. I would rename it.

nlordell (commented):
Some more small nits.

modules/passkey/test/userStories/executeUserOpFromPasskeySigner.ts
```diff
@@ -0,0 +1,153 @@
...
+        dynamic: true,
+      },
+    ])
+
+    packedUserOp.signature = ethers.solidityPacked(['uint48', 'uint48', 'bytes'], [safeOp.validAfter, safeOp.validUntil, signature2])
+
+    // Step 5: Execute the userOp.
+
+    // Send 1 ETH to the Safe
+    await relayer.sendTransaction({ to: safeAddress, value: ethers.parseEther('1') }).then((tx) => tx.wait())
```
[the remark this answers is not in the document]
nlordell: Just noticed the same thing in `modules/passkey/test/userStories/createSafeWithPasskeySignerAsOwner.ts`, do you mind fixing it there too?

modules/passkey/test/userStories/executeUserOpFromPasskeySigner.ts
```diff
@@ -0,0 +1,153 @@
...
+
+    const balanceBefore = await ethers.provider.getBalance(ethers.ZeroAddress)
+
+    await (await entryPoint.handleOps([packedUserOp], relayer.address)).wait()
```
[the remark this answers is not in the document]
nlordell: Just noticed the same thing in `modules/passkey/test/userStories/createSafeWithPasskeySignerAsOwner.ts`, do you mind fixing it there too?

modules/passkey/test/userStories/executeUserOpFromPasskeySigner.ts
```diff
@@ -0,0 +1,153 @@
...
+      0,
+      ethers.ZeroAddress,
+    ])
+
+    // Deploy a Safe with EOA and passkey signer as owners
+    const safeSalt = Date.now()
+    const safeAddress = await proxyFactory.createProxyWithNonce.staticCall(singleton, setupData, safeSalt)
+    await proxyFactory.createProxyWithNonce(singleton, setupData, safeSalt)
+
+    return {
```
remedcu, line 80: nit: We can only pass the required ones, and remove the rest.

remedcu (commented):
Small nit, rest LGTM 👍🏾

modules/passkey/test/userStories/executeUserOpFromPasskeySigner.ts
nlordell: Nice catch

modules/passkey/test/userStories/executeUserOpFromPasskeySigner.ts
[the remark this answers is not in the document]
nlordell: Unresolving as I think that the same issue is still in `createSafeWithPasskeySignerAsOwner.ts`.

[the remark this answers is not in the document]
akshay-ap: Already merged to main in another PR: https://github.com/safe-global/safe-modules/pull/354/files#diff-5e8ff80b8d12563f7724faace05753e5dd17a944c3c01716356044648f32634dR151

modules/passkey/test/userStories/executeUserOpFromPasskeySigner.ts
```diff
@@ -0,0 +1,147 @@
+// Import necessary dependencies from chai, hardhat, @safe-global/safe-4337, webauthn
```
mmv08, line 1: This comment doesn't add any value. In general, a good practice to determine if the comment is useful is the comment should answer the question "why" and not "what"

modules/passkey/test/userStories/executeUserOpFromPasskeySigner.ts
```diff
@@ -0,0 +1,147 @@
...
+      initializer,
+      module.target,
+      ethers.ZeroAddress,
+      0,
+      ethers.ZeroAddress,
+    ])
+
+    // Deploy a Safe with EOA and passkey signer as owners
+    const safeSalt = Date.now()
+    const safeAddress = await proxyFactory.createProxyWithNonce.staticCall(singleton, setupData, safeSalt)
```
mmv08, line 77: nit: if something is doable off-chain, it's better to do that off-chain. We should have a utility function for address calculation somewhere

modules/passkey/test/userStories/executeUserOpFromPasskeySigner.ts
```diff
@@ -0,0 +1,147 @@
...
+      entryPoint,
+      navigator,
+      signer,
+      credential,
+      safeAddress,
+    }
+  })
+
+  it('should execute a userOp with WebAuthn signer as owner', async () => {
+    // Step 1: Setup the contracts
```
mmv08, line 92: The comment doesn't add any value

modules/passkey/test/userStories/executeUserOpFromPasskeySigner.ts
```diff
@@ -0,0 +1,147 @@
...
+      false,
+      true,
+    )
+
+    const packedUserOp = buildPackedUserOperationFromSafeUserOperation({
+      safeOp: safeOp,
+      signature: '0x',
+    })
+
+    const opHash = await module.getOperationHash(packedUserOp)
```
mmv08, line 112: why not offchain?

modules/passkey/test/userStories/executeUserOpFromPasskeySigner.ts
```diff
@@ -0,0 +1,147 @@
...
+        rpId: 'safe.global',
+        allowCredentials: [{ type: 'public-key', id: new Uint8Array(credential.rawId) }],
+        userVerification: 'required',
+      },
+    })
+
+    // Build contract signature that the Safe will forward to the signer contract
+    const signature2 = buildSignatureBytes([
+      {
+        signer: signer as string,
```
mmv08, line 126: why is the casting to string required? isn't it a string already?

modules/passkey/test/userStories/executeUserOpFromPasskeySigner.ts
```diff
@@ -0,0 +1,147 @@
...
+        dynamic: true,
+      },
+    ])
+
+    packedUserOp.signature = ethers.solidityPacked(['uint48', 'uint48', 'bytes'], [safeOp.validAfter, safeOp.validUntil, signature2])
+
+    // Step 3: Execute the userOp.
+
+    // Send 1 ETH to the Safe
+    await relayer.sendTransaction({ to: safeAddress, value: ethers.parseEther('1') })
```
mmv08, line 137:
```suggestion
    // Send 1 ETH to the Safe for the transaction and the prefund
    await relayer.sendTransaction({ to: safeAddress, value: ethers.parseEther('1') })
```

mmv08 (changes requested):

modules/passkey/test/userStories/executeUserOpFromPasskeySigner.ts
nlordell: In general, I agree, but with the user story I think it is nice to use view methods to simplify the code and cognitive load when reading the code here. Implementing this off-chain would just add noise IMO.

modules/passkey/test/userStories/executeUserOpFromPasskeySigner.ts
akshay-ap: it was residue from the initial code. Its not required now. removed in https://github.com/safe-global/safe-modules/pull/344/commits/f84c241a2faf9000dbebff38d522d446e42d3d8c

modules/passkey/test/userStories/executeUserOpFromPasskeySigner.ts
akshay-ap: Alternatively, safeAddress can be retrieved from event logs as it was done in this test previously

modules/passkey/test/userStories/executeUserOpFromPasskeySigner.ts
mmv08: I get this, but I don't think there's a developer that won't understand the motivation behind calling the function named `setupTests`

modules/passkey/test/userStories/executeUserOpFromPasskeySigner.ts
mmv08: Shall we perhaps mention in a comment that an alternative off-chain address calculation via create2 is available?

It's not a dealbreaker for me, but IMO, if we write an example for something, we can expect this example to be followed

modules/passkey/test/userStories/executeUserOpFromPasskeySigner.ts
akshay-ap: Perhaps we should do it for all the tests and not just here. I think its better to cover it in another PR.

mmv08 (approved):

nlordell (approved):

Issue #360, opened 2024-04-04
Enhance Passkey example app


mmv08:
We need to enhance our example application in the examples/ directory to include passkey support similar to the base wallet example application available at [smart-wallet.xyz](https://www.smart-wallet.xyz/). The expected outcome is either adding a new package in the examples/ directory or modifying the existing examples/4337-passkey package to create a demo application with passkey support. The demo application should have features, documentation, and user-friendly elements like the Coinbase wallet demo app. Critical areas for improvement include:

1. Incorporating passkey support to interact with the Wallet once deployed.
2. Implementing features like a "mint" button to enable users to perform account-related actions.
3. Ensuring the application's documentation is comprehensive and user-friendly.

PR #364, opened 2024-04-04
Passkey example app: refactors and dependency upgrades

mmv08:
This PR is a small step towards #360. What it does:
- Refactors the signing part with the passkey. I was rushing a bit when writing it and putting it directly in the `userOp.ts` file while a separate `passkeys.ts` file existed
- Adds `.idea` to the `.gitignore` file
- Upgrades all the dependencies for the app
- **(non-app refactor, I can move it to a separate PR if there are concerns)** Add comments to the `getInitHash` function in the launchpad contract

### DIFF
--- a/examples/4337-passkeys/package.json
+++ b/examples/4337-passkeys/package.json
@@ -7,6 +7,7 @@
     "dev": "vite",
     "build": "tsc && vite build",
     "lint": "eslint . --ext ts,tsx --report-unused-disable-directives --max-warnings 0",
+    "lint:fix": "npm run lint -- --fix",
     "preview": "vite preview"
   },
   "dependencies": {
--- a/examples/4337-passkeys/package.json
+++ b/examples/4337-passkeys/package.json
@@ -10,17 +10,18 @@
     "preview": "vite preview"
   },
   "dependencies": {
-    "@web3modal/ethers": "^4.1.1",
+    "@web3modal/ethers": "^4.1.5",
     "@safe-global/safe-4337": "^0.3.0",
+    "@safe-global/safe-passkey": "*",
     "ethers": "^6.11.1",
     "react": "^18.2.0",
     "react-dom": "^18.2.0"
   },
   "devDependencies": {
-    "@types/react": "^18.2.72",
-    "@types/react-dom": "^18.2.22",
+    "@types/react": "^18.2.74",
+    "@types/react-dom": "^18.2.24",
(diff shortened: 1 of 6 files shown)


examples/4337-passkeys/package.json
```diff
@@ -6,21 +6,22 @@
   "scripts": {
     "dev": "vite",
     "build": "tsc && vite build",
-    "lint": "eslint . --ext ts,tsx --report-unused-disable-directives --max-warnings 0",
+    "lint": "eslint . --ext ts,tsx --report-unused-disable-directives --max-warnings 0 --fix",
```
mmv08, line 9: now this command will lint and fix fixable warnings/errors

nlordell: I removed this a while back, because this causes CI to fix lints instead of error on them.

To fix, I run `npm run lint -- --fix`.

examples/4337-passkeys/package.json
```diff
@@ -6,21 +6,22 @@
...
     "preview": "vite preview"
   },
   "dependencies": {
-    "@web3modal/ethers": "^4.1.1",
+    "@web3modal/ethers": "^4.1.5",
     "@safe-global/safe-4337": "^0.3.0",
+    "@safe-global/safe-passkey": "*",
```
nlordell, line 15: Should both dependencies look like this?

```suggestion
    "@safe-global/safe-4337": "*",
    "@safe-global/safe-passkey": "*",
```

nlordell (commented):
Looks good. One small nit and one question.

nlordell (approved):

examples/4337-passkeys/package.json
mmv08: I pushed a separate script "lint:fix" that runs the above command.

On a side note, I think there's something incorrectly configured, because I never had to run these commands when working on a project. Usually my IDE would pick up all the formatting settings and execute them automatically.

From prettier logs:
```
["INFO" - 11:00:13 AM] Extension Name: esbenp.prettier-vscode.
["INFO" - 11:00:13 AM] Extension Version: 10.4.0.
["INFO" - 11:00:13 AM] Using config file at /home/mmv/Projects/Safe/safe-modules/.prettierrc
["INFO" - 11:00:23 AM] Using config file at /home/mmv/Projects/Safe/safe-modules/.prettierrc
```

Looks like it uses the global prettier file and not the nested one.

PR #370, opened 2024-04-08
Create SafeWebAuthnSignerProxy

### LINKED ISSUE #312: Evaluate Overhead Of Deploying WebAuthn Logic In Signer
One reason to have the `WebAuthn` signer contract "proxy" the signature verification logic to a `WebAuthnVerifier` (or the WebAuthn contract with #311) is to save on deployment gas.

However, given a more optimal Base64 implementation in #289, it may not save that much deployment bytecode to merrit the separation.

This issue captures the work to analyze including WebAuthn logic directly in the `WebAuthnSigner` contract and evaluate the gas tradeoffs of doing so.

akshay-ap:
Fixes #312 

Changes in PR:
- [x] Create `SafeWebAuthnSignerSingleton`. This contract has no storage, no immutables, and no constants.
- [x] Create `SafeWebAuthnSignerProxy` and corresponding proxy factory i.e. `SafeWebAuthnSignerProxyFactory`

- Create immutable `singleton` contract when deploying proxy factory
- `SafeWebAuthnSignerProxyFactory.isValidSignatureForSigner(...)` forwards to singleton contract instead directly calling `WebAuthn` library
- `SafeWebAuthnSignerProxy` uses no storage to avoid 4337 storage access restriction violation.
- Add unit test to verify that `SafeWebAuthnSignerProxy` forwards every call to singleton address
- [x] Report gas usage metrics
- [x] Update github workflow to execute benchmarking tests separately. (This is required to get accurate gas usage numbers)
- [x] Update documentation
- [x] Replace `SafeWebAuthnSigner` with `SafeWebAuthnSignerProxy`, same for factory contract
- [x] Update all tests to use `SafeWebAuthnSignerProxy`. Hence, the large diff in this PR.

Benchmark number from this PR:

```
Gas Benchmarking Proxy [@bench]
    SafeWebAuthnSignerProxy
      ⛽ deployment: 164012
      ✔ Benchmark signer deployment cost (10[38](https://github.com/safe-global/safe-modules/actions/runs/8739938588/job/23982409583?pr=370#step:4:39)ms)
      ⛽ verification (FreshCryptoLib): 216248
      ✔ Benchmark signer verification cost with FreshCryptoLib verifier (201ms)
      ⛽ verification (daimo-eth): 346151
      ✔ Benchmark signer verification cost with daimo-eth verifier (186ms)
      ⛽ verification (Dummy): 14[40](https://github.com/safe-global/safe-modules/actions/runs/8739938588/job/23982409583?pr=370#step:4:41)2
```

Benchmark number from main:
```
  Gas Benchmarking [@bench]
    SafeWebAuthnSigner
      ⛽ deployment: 491288
      ✔ Benchmark signer deployment cost (654ms)
      ⛽ verification (FreshCryptoLib): 209630
      ✔ Benchmark signer verification cost with FreshCryptoLib verifier (138ms)
      ⛽ verification (daimo-eth): 336881
      ✔ Benchmark signer verification cost with daimo-eth verifier (108ms)
      ⛽ verification (Dummy): 11266
```


modules/passkey/contracts/SafeWebAuthnSignerSingleton.sol
```diff
@@ -0,0 +1,41 @@
...
+/**
+ * @title WebAuthn Safe Signature Validator Singleton
+ * @dev A contract that represents a WebAuthn signer.
+ * @custom:security-contact bounty@safe.global
+ */
+contract SafeWebAuthnSignerSingleton is SignatureValidator {
+    address public singleton;
+    uint256 public x;
+    uint256 public y;
+    IP256Verifier public verifier;
```
nlordell, line 17: Unfortunately, this will not work with 4337 :(

akshay-ap: This should resolve the problem https://github.com/safe-global/safe-modules/pull/370/commits/764804b9a8d0d4783b05533d8763faecf8cb3cea

modules/passkey/contracts/SafeWebAuthnSignerProxy.sol
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+ * @title SafeWebAuthnSignerProxy
+ * @dev A proxy contract that points to SafeWebAuthnSigner.
+ */
+contract SafeWebAuthnSignerProxy {
+    uint256 internal immutable X;
+    uint256 internal immutable Y;
+    IP256Verifier internal immutable VERIFIER;
+    address internal singleton;
```
nlordell, line 13: This also needs to be immutable for 4337.

modules/passkey/contracts/SafeWebAuthnSignerProxy.sol
```diff
@@ -0,0 +1,45 @@
...
+        bytes memory data = abi.encodePacked(msg.data, X, Y, VERIFIER);
+
+        // solhint-disable-next-line no-inline-assembly
+        assembly {
+            let _singleton := and(sload(0), 0xffffffffffffffffffffffffffffffffffffffff)
+            // 0xa619486e == keccak("masterCopy()"). The value is right padded to 32-bytes with 0s
+            if eq(calldataload(0), 0xa619486e00000000000000000000000000000000000000000000000000000000) {
+                mstore(0, _singleton)
+                return(0, 0x20)
+            }
```
nlordell, line 33: Because we have an immutable `singleton`, I don't think that we need to support a `masterCopy` method.

akshay-ap: https://github.com/safe-global/safe-modules/pull/370/commits/8f613e69f3309c6ae7c0b5772906f9fa1a0735e0

modules/passkey/contracts/SafeWebAuthnSignerProxy.sol
akshay-ap: https://github.com/safe-global/safe-modules/pull/370/commits/8f613e69f3309c6ae7c0b5772906f9fa1a0735e0

modules/passkey/contracts/SafeWebAuthnSignerProxyFactory.sol
```diff
@@ -0,0 +1,64 @@
...
+import {ICustomECDSASignerProxyFactory} from "./interfaces/ICustomECDSASignerProxyFactory.sol";
+import {IP256Verifier} from "./interfaces/IP256Verifier.sol";
+import {ERC1271} from "./libraries/ERC1271.sol";
+import {WebAuthn} from "./libraries/WebAuthn.sol";
+import {SafeWebAuthnSignerProxy} from "./SafeWebAuthnSignerProxy.sol";
+/**
+ * @title SafeWebAuthnSignerProxyFactory
+ * @dev A factory contract for creating and managing WebAuthn proxy signers.
+ */
+contract SafeWebAuthnSignerProxyFactory is ICustomECDSASignerProxyFactory {
```
nlordell, line 13: This should not require an interface change. I think the singleton should be chosen per factory and set in the constructor.

modules/passkey/contracts/SafeWebAuthnSignerProxyFactory.sol
```diff
@@ -0,0 +1,64 @@
...
+     * @inheritdoc ICustomECDSASignerProxyFactory
+     */
+    function isValidSignatureForSigner(
+        bytes32 message,
+        bytes calldata signature,
+        uint256 x,
+        uint256 y,
+        address verifier
+    ) external view override returns (bytes4 magicValue) {
+        if (WebAuthn.verifySignature(message, signature, WebAuthn.USER_VERIFICATION, x, y, IP256Verifier(verifier))) {
```
nlordell, line 46: This should use the singleton for verification. It ensures that the implementation is the same.

modules/passkey/contracts/SafeWebAuthnSignerProxyFactory.sol
```diff
@@ -10,31 +10,36 @@ import {SafeWebAuthnSignerProxy} from "./SafeWebAuthnSignerProxy.sol";
  * @title SafeWebAuthnSignerProxyFactory
  * @dev A factory contract for creating and managing WebAuthn proxy signers.
  */
-contract SafeWebAuthnSignerProxyFactory is ICustomECDSASignerProxyFactory {
+contract SafeWebAuthnSignerProxyFactory is ICustomECDSASignerFactory {
+    address public immutable SINGLETON;
+    constructor(address _singleton) {
+        SINGLETON = _singleton;
```
nlordell, line 16: You can just deploy the `SINGLETON` here right?

```suggestion
    address public immutable SINGLETON;

    constructor() {
        SINGLETON = new SafeWebAuthnSignerSingleton;
```

modules/passkey/contracts/SafeWebAuthnSignerProxyFactory.sol
akshay-ap: done

nlordell:
@akshay-ap - can you rebase the PR, it would help with the review process.

examples/4337-passkeys/src/logic/safe.ts
```diff
@@ -28,7 +28,7 @@ const SafeProxyBytecode =
 function getSignerAddressFromPubkeyCoords(x: string, y: string): string {
   const deploymentCode = ethers.solidityPacked(
     ['bytes', 'uint256', 'uint256', 'uint256'],
-    [SafeWebAuthSignerBytecode, x, y, P256_VERIFIER_ADDRESS],
+    [SafeWebAuthnSignerSingletonBytecode, x, y, P256_VERIFIER_ADDRESS],
```
nlordell, line 31: This seems wrong to me - shouldn't it use the proxy code for computing the `CREATE2` address? Might be worth waiting for #389 to rebase and test the example application works.

In general, I think we need to find a good way to make the DevEx in this repo better for incorporating changes to the `passkey` and `4337` code into the example application, because it is becoming a bit of a maintenance burden ATM.

modules/passkey/contracts/SafeWebAuthnSignerProxyFactory.sol
```diff
@@ -0,0 +1,92 @@
+// SPDX-License-Identifier: LGPL-3.0-only
```
nlordell, line 1: Can we please not rename this file? I think it is still a `SignerFactory`. The fact it uses proxy contracts is an internal detail that does not need to be reflected in a change of contract name IMO.

modules/passkey/contracts/base/SignatureValidator.sol
```diff
@@ -28,7 +38,17 @@ abstract contract SignatureValidator {
...
-        if (_verifySignature(message, signature)) {
+        uint256 x;
+        uint256 y;
+        P256.Verifiers verifiers;
+        // solhint-disable-next-line no-inline-assembly
+        assembly {
+            x := calldataload(sub(calldatasize(), 88))
+            y := calldataload(sub(calldatasize(), 56))
+            verifiers := shr(64, calldataload(sub(calldatasize(), 24)))
+        }
```
nlordell, line 49: Hmm, to me this feels like it should be part of `_verifySignature` as it is duplicated and not part of the base `SignatureValidator`.

The `SignatureValidator` base class is a *general* implementation of a smart contract owner for Safes, and the fact that we are using this trusted forwarder patterns is an implementation detail of the actual signer, and not of the base class.

modules/passkey/contracts/test/TestWebAuthnSingletonSigner.sol
```diff
@@ -1,53 +0,0 @@
-// SPDX-License-Identifier: LGPL-3.0-only
```
nlordell, line 1: This was used for 4337 integration tests, and I think we should keep it around.

examples/4337-passkeys/src/logic/safe.ts
nlordell: Also, I think it is missing the `singletonAddress` for computing the `CREATE2` address.

modules/passkey/contracts/SafeWebAuthnSignerProxy.sol
```diff
@@ -0,0 +1,50 @@
...
+    address internal immutable SINGLETON;
+    constructor(address implementation, uint256 x, uint256 y, P256.Verifiers verifiers) {
+        SINGLETON = implementation;
+        X = x;
+        Y = y;
+        VERIFIERS = verifiers;
+    }
+
+    /// @dev Fallback function forwards all transactions and returns all received return data.
+    // solhint-disable-next-line no-complex-fallback
```
nlordell, line 32: nit: We use `/** */` style NatSpec in the project, I would prefer for it to remain consistent.

modules/passkey/contracts/SafeWebAuthnSignerProxy.sol
```diff
@@ -0,0 +1,50 @@
...
+    uint256 public immutable X;
+    /**
+     * @notice The Y coordinate of the P-256 public key of the WebAuthn credential.
+     */
+    uint256 public immutable Y;
+    /**
+     * @notice The P-256 verifiers used for ECDSA signature validation.
+     */
+    P256.Verifiers public immutable VERIFIERS;
+    address internal immutable SINGLETON;
```
nlordell, line 23: This is missing `NatSpec`

modules/passkey/contracts/SafeWebAuthnSignerProxyFactory.sol
akshay-ap: I did it because we have `SafeProxyFactory` for safe creating safe accounts. Will revert name change though.

akshay-ap: https://github.com/safe-global/safe-modules/pull/370/commits/fa594dc1fd83e353ae7c362e4f221c963c717dbb

examples/4337-passkeys/src/logic/safe.ts
mmv08: Just FYI,  in https://github.com/safe-global/safe-modules/pull/389 I replaced this with an RPC call to the factory's `getSigner` function

modules/passkey/contracts/base/SignatureValidator.sol
akshay-ap: https://github.com/safe-global/safe-modules/pull/370/commits/954078b6179e8fea2c8b7da9774a779281d2428d

modules/passkey/contracts/SafeWebAuthnSignerProxy.sol
akshay-ap: https://github.com/safe-global/safe-modules/pull/370/commits/954078b6179e8fea2c8b7da9774a779281d2428d

modules/passkey/contracts/SafeWebAuthnSignerFactory.sol
```diff
@@ -43,8 +55,24 @@ contract SafeWebAuthnSignerFactory is ISafeSignerFactory {
...
+        );
+
+        // solhint-disable-next-line no-inline-assembly
+        assembly {
+            let dataSize := mload(data)
+            let dataLocation := add(data, 0x20)
+
+            let success := staticcall(gas(), _singleton, dataLocation, dataSize, 0, 0)
+            if eq(success, 1) {
+                returndatacopy(magicValue, 0, returndatasize())
```
nlordell, line 73: I think this is incorrect. It will copy to offset `0`, the default value for a `bytes4` variable.

modules/passkey/contracts/SafeWebAuthnSignerProxy.sol
```diff
@@ -0,0 +1,56 @@
...
+/**
+ * @title WebAuthn Safe Signature Validator
+ * @dev A proxy contracy that points to Safe signature validator implementation for a WebAuthn P-256 credential.
+ * @custom:security-contact bounty@safe.global
+ */
+contract SafeWebAuthnSignerProxy {
+    /**
+     * @notice The X coordinate of the P-256 public key of the WebAuthn credential.
+     */
+    uint256 internal immutable X;
```
nlordell, line 14: I would try and see if using `private` makes a difference (as Solidity will not have to generate functions for these).

nlordell: Also, in terms of naming convention, we prefix `internal` and `private` members with an underscore `_`.

modules/passkey/contracts/SafeWebAuthnSignerProxy.sol
```diff
@@ -0,0 +1,56 @@
...
+        VERIFIERS = verifiers;
+    }
+
+    /**
+     * @dev Fallback function forwards all transactions and returns all received return data.
+     */
+    // solhint-disable-next-line no-complex-fallback
+    fallback() external payable {
+        bytes memory data = abi.encodePacked(msg.data, X, Y, VERIFIERS);
+        address _singleton = SINGLETON;
```
nlordell, line 41:
```suggestion
        address singleton = SINGLETON;
```

modules/passkey/contracts/SafeWebAuthnSignerSingleton.sol
```diff
@@ -0,0 +1,48 @@
...
+            y := calldataload(sub(calldatasize(), 56))
+        }
+    }
+
+    function getVerifiers() external view returns (P256.Verifiers verifiers) {
+        // solhint-disable-next-line no-inline-assembly
+        assembly {
+            verifiers := shr(64, calldataload(sub(calldatasize(), 24)))
+        }
+    }
```
nlordell, line 47: If you instead write this as:

```
     function getConfiguration() public view returns (uint256 x, uint256 y, P256.Verifiers verifiers) {
```

You can reuse it in `_verifySignature`.

modules/passkey/test/GasBenchmarkingProxy.spec.ts
```diff
@@ -2,11 +2,12 @@ import { expect } from 'chai'
...
 
 import * as ERC1271 from './utils/erc1271'
-import { WebAuthnCredentials } from '../test/utils/webauthnShim'
+import { WebAuthnCredentials } from './utils/webauthnShim'
 import { decodePublicKey, encodeWebAuthnSignature } from '../src/utils/webauthn'
 import { IP256Verifier } from '../typechain-types'
+import { setCode } from '@nomicfoundation/hardhat-toolbox/network-helpers'
 
-describe('Gas Benchmarking [@bench]', function () {
+describe('Gas Benchmarking Proxy [@bench]', function () {
```
nlordell, line 10:
```suggestion
describe('Gas Benchmarking [@bench]', function () {
```

modules/passkey/test/GasBenchmarkingProxy.spec.ts
```diff
@@ -32,7 +33,7 @@ describe('Gas Benchmarking [@bench]', function () {
     const Benchmarker = await ethers.getContractFactory('Benchmarker')
     const benchmarker = await Benchmarker.deploy()
 
-    const factory = await ethers.getContractAt('SafeWebAuthnSignerFactory', SafeWebAuthnSignerFactory.address)
+    const proxyFactory = await ethers.getContractAt('SafeWebAuthnSignerFactory', SafeWebAuthnSignerFactory.address)
```
nlordell, line 36:
```suggestion
    const factory = await ethers.getContractAt('SafeWebAuthnSignerFactory', SafeWebAuthnSignerFactory.address)
```

modules/passkey/test/GasBenchmarkingProxy.spec.ts
```diff
@@ -75,15 +79,22 @@ describe('Gas Benchmarking [@bench]', function () {
...
-        const signer = await ethers.getContractAt('SafeWebAuthnSigner', await factory.getSigner(x, y, verifier))
+        const isValidSignatureInterface = new ethers.Interface(['function isValidSignature(bytes32,bytes) external view returns (bytes4)'])
+
+        const precompileAddress = `0x${'00'.repeat(18)}0100`
+        await setCode(precompileAddress, await ethers.provider.getCode(verifier))
+
+        const verifiersData = BigInt(ethers.solidityPacked(['uint32', 'address'], [Number(precompileAddress), verifier]))
+
+        await proxyFactory.createSigner(x, y, verifiersData)
+        const signerProxy = await ethers.getContractAt('SafeWebAuthnSignerProxy', await proxyFactory.getSigner(x, y, verifiersData))
```
nlordell, line 92: I'm not sure I understand the rationale behind configuring a `precompile` address for benchmarking. I would revert this change.

modules/passkey/test/SafeWebAuthnSignerProxyFactory.spec.ts
```diff
@@ -9,16 +9,18 @@ import { encodeWebAuthnSigningMessage } from './utils/webauthnShim'
 describe('SafeWebAuthnSignerFactory', () => {
   const setupTests = deployments.createFixture(async () => {
     const { SafeWebAuthnSignerFactory } = await deployments.fixture()
+
     const factory = await ethers.getContractAt('SafeWebAuthnSignerFactory', SafeWebAuthnSignerFactory.address)
 
+    const singletonAddress = await factory.SINGLETON()
```
nlordell, line 15: This is only needed for a test, I wouldn't consider it as being part of the "Setup".

modules/passkey/test/userstories/PasskeyCredentialCreation.spec.ts
```diff
@@ -51,7 +51,7 @@ describe('Passkey Credential Creation for Safe Ownership [@userstory]', () => {
     const publicKey = decodePublicKey(credential.response)
     await signerFactory.createSigner(publicKey.x, publicKey.y, verifierAddress)
     const signerAddress = await signerFactory.getSigner(publicKey.x, publicKey.y, verifierAddress)
-    const signer = await ethers.getContractAt('SafeWebAuthnSigner', signerAddress)
+    const signer = await ethers.getContractAt('SafeWebAuthnSignerProxy', signerAddress)
```
nlordell, line 54: Same here.

modules/passkey/contracts/SafeWebAuthnSignerProxy.sol
akshay-ap: > I would try and see if using private makes a difference (as Solidity will not have to generate functions for these).
The byte codesize remains same after marking private. As per documentation, getter function is generated only for `public` variables.
https://docs.soliditylang.org/en/latest/contracts.html#state-variable-visibility

modules/passkey/test/userstories/PasskeyCredentialCreation.spec.ts
akshay-ap: https://github.com/safe-global/safe-modules/pull/370/commits/577142e61bd4f5128fda13d8bc315ded43410805

modules/passkey/contracts/SafeWebAuthnSignerFactory.sol
akshay-ap: good catch!

modules/passkey/contracts/SafeWebAuthnSignerFactory.sol
```diff
@@ -2,21 +2,33 @@
...
+import {P256} from "./libraries/WebAuthn.sol";
 
 /**
- * @title WebAuthnSignerFactory
- * @dev A factory contract for creating and managing WebAuthn signers.
+ * @title SafeWebAuthnSignerFactory
+ * @dev A factory contract for creating and managing WebAuthn proxy signers.
  */
 contract SafeWebAuthnSignerFactory is ISafeSignerFactory {
+    address public immutable SINGLETON;
```
nlordell, line 14: ubernit: Since here we know **for sure** that `SINGLETON` is a `SafeWebAuthnSignerSingleton`, I wouldn't use the `address` type.

modules/passkey/contracts/SafeWebAuthnSignerFactory.sol
```diff
@@ -42,9 +54,25 @@ contract SafeWebAuthnSignerFactory is ISafeSignerFactory {
         uint256 x,
         uint256 y,
         P256.Verifiers verifiers
-    ) external view override returns (bytes4 magicValue) {
-        if (WebAuthn.verifySignature(message, signature, WebAuthn.USER_VERIFICATION, x, y, verifiers)) {
-            magicValue = ERC1271.MAGIC_VALUE;
+    ) external view override returns (bytes4 /*magicValue*/) {
+        address _singleton = SINGLETON;
```
nlordell, line 58:
```suggestion
        address singleton = SINGLETON;
```

modules/passkey/contracts/SafeWebAuthnSignerSingleton.sol
```diff
@@ -0,0 +1,39 @@
...
+contract SafeWebAuthnSignerSingleton is SignatureValidator {
+    /**
+     * @inheritdoc SignatureValidator
+     */
+    function _verifySignature(bytes32 message, bytes calldata signature) internal view virtual override returns (bool success) {
+        uint256 x;
+        uint256 y;
+        P256.Verifiers verifiers;
+        // solhint-disable-next-line no-inline-assembly
+        (x, y, verifiers) = getConfiguration();
```
nlordell, line 20:
```suggestion
        (uint256 x, uint256 y, P256.Verifiers verifiers) = getConfiguration();
```

modules/passkey/contracts/SafeWebAuthnSignerSingleton.sol
```diff
@@ -0,0 +1,39 @@
...
+
+    /**
+     * @notice Returns the x coordinate, y coordinate, and P-256 verifiers used for ECDSA signature validation. The values are expected to be passed by the SafeWebAuthnSignerProxy contract in msg.data.
+     * @return x The x coordinate of the P-256 public key.
+     * @return y The y coordinate of the P-256 public key.
+     * @return verifiers The P-256 verifiers.
+     */
+    function getConfiguration() public view returns (uint256 x, uint256 y, P256.Verifiers verifiers) {
+        // solhint-disable-next-line no-inline-assembly
+        assembly {
```
nlordell, line 33:
```suggestion
        assembly ("memory-safe") {
```

modules/passkey/contracts/SafeWebAuthnSignerFactory.sol
akshay-ap: I kept it address because the code does not use Singleton related function anywhere and only need address in assembly call. If we use `SafeWebAuthnSignerSingleton` then we need to cast it to `address` which incurs 2 gas cost.

nlordell (commented):
I don't know if you saw my comment, but the following files were removed and I don't think they should be:

- modules/passkey/contracts/test/TestWebAuthnSingletonSigner.sol
- modules/passkey/test/4337/WebAuthnSingletonSigner.spec.ts

akshay-ap:
> I don't know if you saw my comment, but the following files were removed and I don't think they should be:
> 
> * modules/passkey/contracts/test/TestWebAuthnSingletonSigner.sol
> * modules/passkey/test/4337/WebAuthnSingletonSigner.spec.ts

yes. restoring them

modules/passkey/contracts/SafeWebAuthnSignerFactory.sol
nlordell: The reason I suggested this was that this is the only occurrence of a function local variable starting with `_` in the passkey code. The conversation was marked as resolved without additional context, so I just want to understand why you think we should keep it this way.

akshay-ap: I added the change to code here: https://github.com/safe-global/safe-modules/pull/370/files#diff-ce13df21c642f8e2ad855020959aff44c4035a27feb2c9e478cb7f96be9d620dR58

modules/passkey/contracts/SafeWebAuthnSignerFactory.sol
nlordell: In theory, `SafeWebAuthnSignerSingleton` is already an address. I'm surprised that the compiler generates different code. I guess this only affects the `getSigner` call.

akshay-ap: also affects `isValidSignatureForSigner` and `createSigner`

examples/4337-passkeys/src/logic/safe.ts
akshay-ap: marking as resolved after rebase

modules/passkey/contracts/SafeWebAuthnSignerFactory.sol
```diff
@@ -2,21 +2,33 @@
 pragma solidity >=0.8.0;
 
 import {ISafeSignerFactory} from "./interfaces/ISafeSignerFactory.sol";
-import {ERC1271} from "./libraries/ERC1271.sol";
-import {P256, WebAuthn} from "./libraries/WebAuthn.sol";
-import {SafeWebAuthnSigner} from "./SafeWebAuthnSigner.sol";
+import {SafeWebAuthnSignerProxy} from "./SafeWebAuthnSignerProxy.sol";
+import {SafeWebAuthnSignerSingleton} from "./SafeWebAuthnSignerSingleton.sol";
+import {P256} from "./libraries/WebAuthn.sol";
```
nlordell, line 7:
```suggestion
import {P256} from "./libraries/P256.sol";
```

modules/passkey/contracts/SafeWebAuthnSignerFactory.sol
```diff
@@ -42,9 +54,25 @@ contract SafeWebAuthnSignerFactory is ISafeSignerFactory {
...
+            verifiers
+        );
+
+        // solhint-disable-next-line no-inline-assembly
+        assembly {
+            let dataSize := mload(data)
+            let dataLocation := add(data, 0x20)
+
+            let success := staticcall(gas(), singleton, dataLocation, dataSize, 0, 0)
+            if eq(success, 1) {
```
nlordell, line 72: Shouldn't this revert when the inner call reverts?

modules/passkey/contracts/SafeWebAuthnSignerProxy.sol
```diff
@@ -0,0 +1,64 @@
...
+        address singleton = _SINGLETON;
+
+        // solhint-disable-next-line no-inline-assembly
+        assembly {
+            let dataSize := mload(data)
+            let dataLocation := add(data, 0x20)
+
+            let success := delegatecall(gas(), singleton, dataLocation, dataSize, 0, 0)
+            returndatacopy(0, 0, returndatasize())
+            if eq(success, 0) {
```
nlordell, line 58:
```suggestion
            if iszero(success) {
```

modules/passkey/contracts/SafeWebAuthnSignerFactory.sol
nlordell: I was curious as to what 2 gas opcode instruction the compiler was adding for no reason, however, when checking locally there was no change between using address or the actual contract type:

**Before**
```
  Gas Benchmarking [@bench]
    SafeWebAuthnSignerProxy
      ⛽ deployment: 108099
      ✔ Benchmark signer deployment cost (763ms)
...
      ⛽ verification (Dummy): 14420
```

**After**:
```
  Gas Benchmarking [@bench]
    SafeWebAuthnSignerProxy
      ⛽ deployment: 108099
      ✔ Benchmark signer deployment cost (934ms)
...
      ⛽ verification (Dummy): 14420
```

(note that there is natural variance in the costs of actually recovering signatures with Daimo.eth and FCL)

FWIW, here is my patch

```diff
diff --git a/modules/passkey/contracts/SafeWebAuthnSignerFactory.sol b/modules/passkey/contracts/SafeWebAuthnSignerFactory.sol
index 7a6a682..3fb4dca 100644
--- a/modules/passkey/contracts/SafeWebAuthnSignerFactory.sol
+++ b/modules/passkey/contracts/SafeWebAuthnSignerFactory.sol
@@ -11,10 +11,10 @@ import {P256} from "./libraries/WebAuthn.sol";
  * @dev A factory contract for creating and managing WebAuthn proxy signers.
  */
 contract SafeWebAuthnSignerFactory is ISafeSignerFactory {
-    address public immutable SINGLETON;
+    SafeWebAuthnSignerSingleton public immutable SINGLETON;
 
     constructor() {
-        SINGLETON = address(new SafeWebAuthnSignerSingleton());
+        SINGLETON = new SafeWebAuthnSignerSingleton();
     }
 
     /**
@@ -24,7 +24,7 @@ contract SafeWebAuthnSignerFactory is ISafeSignerFactory {
         bytes32 codeHash = keccak256(
             abi.encodePacked(
                 type(SafeWebAuthnSignerProxy).creationCode,
-                uint256(uint160(SINGLETON)),
+                uint256(uint160(address(SINGLETON))),
                 x,
                 y,
                 uint256(P256.Verifiers.unwrap(verifiers))
@@ -40,7 +40,7 @@ contract SafeWebAuthnSignerFactory is ISafeSignerFactory {
         signer = getSigner(x, y, verifiers);
 
         if (_hasNoCode(signer)) {
-            SafeWebAuthnSignerProxy created = new SafeWebAuthnSignerProxy{salt: bytes32(0)}(SINGLETON, x, y, verifiers);
+            SafeWebAuthnSignerProxy created = new SafeWebAuthnSignerProxy{salt: bytes32(0)}(address(SINGLETON), x, y, verifiers);
             require(address(created) == signer);
         }
     }
@@ -55,7 +55,7 @@ contract SafeWebAuthnSignerFactory is ISafeSignerFactory {
         uint256 y,
         P256.Verifiers verifiers
     ) external view override returns (bytes4 /*magicValue*/) {
-        address singleton = SINGLETON;
+        SafeWebAuthnSignerSingleton singleton = SINGLETON;
         bytes memory data = abi.encodePacked(
             abi.encodeWithSignature("isValidSignature(bytes32,bytes)", message, signature),
             x,

```

nlordell (commented):
I noticed when running locally that there are compiler warning that can be resolved.

modules/passkey/contracts/SafeWebAuthnSignerFactory.sol
akshay-ap: ohh.. i see... I was getting the prompt in vscode as below and thought prompt text is buggy but casting would incur additional 2 gas.

Screenshot 2024-04-23 at 10 11 13

I would cast to address as mentioned by you.

akshay-ap: ooh.. i see.. Vscode was giving buggy incorrect prompt when casting to address. I would follow your suggesstion.

akshay-ap: https://github.com/safe-global/safe-modules/pull/370/commits/f35a2eb452100fa33f5250391bea457fd980b89f

modules/passkey/contracts/SafeWebAuthnSignerFactory.sol
akshay-ap: I did not add revert because as per specs in `ISafeSignerFactory`. So, thought not reverting would also work.
```
  * @return magicValue Returns the ERC-1271 magic value when the signature is valid. Reverting or
     * returning any other value implies an invalid signature.
```

nlordell: Good point. In this case then, instead of `return`-ing we can copy the return data to offset 0 and with length 32 and then `magicValue := mload(0)`.

This way we don't have separate returning paths from the function, but a single one. Also, from a conceptual level it is nicer in that if the singleton does follow the same return ABI as the factory function, things would still work.

modules/passkey/contracts/SafeWebAuthnSignerFactory.sol
```diff
@@ -42,9 +54,25 @@ contract SafeWebAuthnSignerFactory is ISafeSignerFactory {
...
+                returndatacopy(0, 0, returndatasize())
+                return(0, returndatasize())
+            }
```
nlordell, line 75:
```suggestion
            if staticcall(gas(), singleton, dataLocation, dataSize, 0, 32)
 {
                magicValue := mload(0)
            }
```

modules/passkey/contracts/SafeWebAuthnSignerFactory.sol
akshay-ap: done. 
https://github.com/safe-global/safe-modules/pull/370/commits/3138afc05b321142cf4fe22263edd0f94b7bac6e
https://github.com/safe-global/safe-modules/pull/370/commits/7c6bb5755b3be37c5093e85abcd6676fc284d767
https://github.com/safe-global/safe-modules/pull/370/commits/a35771a9833ef023303fe0599587b96246e90b3e

nlordell (approved):

PR #371, opened 2024-04-08
NatSpec and Unit Test Coverage for `SafeWebAuthnSigner`

nlordell:
This PR adds some missing NatSpec documentation to the `SafeWebAuthnSigner` contract (notably the `immutable`s) and implements missing coverage for the contract. In the unit and E2E tests, we only ever call the `isValidSignature(bytes,bytes)` function, and never the `isValidSignature(bytes32,bytes)`.

Related to #365

### DIFF
--- a/modules/passkey/test/utils/webauthn.ts
+++ b/modules/passkey/test/utils/webauthn.ts
@@ -195,7 +195,7 @@ export class WebAuthnCredentials {
 
     // <https://w3c.github.io/webauthn/#dictionary-client-data>
     const clientData = {
-      type: 'webauthn.get',
+      type: 'webauthn.get' as const,
       challenge: base64UrlEncode(publicKey.challenge),
       origin: `https://${publicKey.rpId}`,
     }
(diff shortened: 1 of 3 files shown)


modules/passkey/test/utils/webauthn.ts
```diff
@@ -195,7 +195,7 @@ export class WebAuthnCredentials {
 
     // <https://w3c.github.io/webauthn/#dictionary-client-data>
     const clientData = {
-      type: 'webauthn.get',
+      type: 'webauthn.get' as const,
```
remedcu, line 198: TIL why use `as const`. More info [here](https://stackoverflow.com/a/66993654/7520013).

remedcu (approved):
nit: Use the 1271 magic value as a constant in the test for easier comprehension while reading the code.

nlordell:
> Use the 1271 magic value as a constant in the test for easier comprehension while reading the code.

Addressed in #374, and updated other tests to also use constants.

PR #373, opened 2024-04-09
Full Coverage of `SafeWebAuthnSignerFactory` Contract

### LINKED ISSUE #365: Add Full Test Coverage for `WebAuthnSigner` and `WebAuthnSignerFactory`

nlordell:
Fixes #365 

This PR adds missing branch coverage for the `SafeWebAuthnSignerFactory` contract so that it is fully covered.

Note that we changed a `require` to an `assert`. This was prompted by the Solidity coverage tool complaining that all paths weren't taken. However, having that path not be taken would imply that the address derivation for `CREATE2` to not function as expected, so should never happen on EVM chains (note to self, do we require a slightly different contract for zkSync?). Looking through the Solidity documentation, I found that `assert` was more applicable here, and luckily the coverage tool does not expect the negative path to be explored. From the docs:

> Assert should only be used to test for internal errors, and to check invariants. Properly functioning code should never create a Panic, not even on invalid external input. If this happens, then there is a bug in your contract which you should fix. Language analysis tools can evaluate your contract to identify the conditions and function calls which will cause a Panic.

According to this description, this is a exemplar use of the `assert` builtin.

### DIFF
--- a/modules/passkey/test/SafeWebAuthnSignerFactory.spec.ts
+++ b/modules/passkey/test/SafeWebAuthnSignerFactory.spec.ts
@@ -29,7 +29,7 @@ describe('SafeWebAuthnSignerFactory', () => {
       expect(ethers.dataLength(await ethers.provider.getCode(signer))).to.not.equal(0)
     })
 
-    it('Should return the address that a signer will be created on', async () => {
+    it('Should return different signer for different inputs', async () => {
       const { factory, mockVerifier } = await setupTests()
 
       const x = ethers.id('publicKey.x')
--- a/modules/passkey/test/SafeWebAuthnSignerFactory.spec.ts
+++ b/modules/passkey/test/SafeWebAuthnSignerFactory.spec.ts
@@ -1,0 +1,178 @@
+import { expect } from 'chai'
+import { deployments, ethers } from 'hardhat'
+import { DUMMY_AUTHENTICATOR_DATA, base64UrlEncode, encodeWebAuthnSigningMessage, getSignatureBytes } from './utils/webauthn'
+
+describe('SafeWebAuthnSignerFactory', () => {
+  const setupTests = deployments.createFixture(async () => {
+    const { SafeWebAuthnSignerFactory } = await deployments.fixture()
+    const factory = await ethers.getContractAt('SafeWebAuthnSignerFactory', SafeWebAuthnSignerFactory.address)
+
+    const MockContract = await ethers.getContractFactory('MockContract')
+    const mockVerifier = await MockContract.deploy()
+
+    return { factory, mockVerifier }
+  })
+
+  describe('getSigner', function () {
+    it('Should return the address that a signer will be created on', async () => {
+      const { factory, mockVerifier } = await setupTests()
+
+      const x = ethers.id('publicKey.x')
+      const y = ethers.id('publicKey.y')
+
+      const signer = await factory.getSigner(x, y, mockVerifier)
+
+      expect(ethers.dataLength(await ethers.provider.getCode(signer))).to.equal(0)
+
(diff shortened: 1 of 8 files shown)


modules/passkey/test/SafeWebAuthnSignerFactory.spec.ts
```diff
@@ -0,0 +1,178 @@
...
+    const factory = await ethers.getContractAt('SafeWebAuthnSignerFactory', SafeWebAuthnSignerFactory.address)
+
+    const MockContract = await ethers.getContractFactory('MockContract')
+    const mockVerifier = await MockContract.deploy()
+
+    return { factory, mockVerifier }
+  })
+
+  describe('getSigner', function () {
+    it('Should return the address that a signer will be created on', async () => {
```
remedcu, line 17: The test description feels a bit ambiguous based on the `expect` we check in this test.

modules/passkey/test/SafeWebAuthnSignerFactory.spec.ts
```diff
@@ -0,0 +1,178 @@
...
+        [x, ethers.id('publicKey.otherY'), mockVerifier],
+        [x, y, '0x0000000000000000000000000000000000000100'],
+      ] as const) {
+        expect(await factory.getSigner(...params)).to.not.equal(signer)
+      }
+    })
+  })
+
+  describe('createSigner', function () {
+    it('Should create a signer and return its address', async () => {
```
remedcu, line 51: Similar to the previous comment, we ensure the creation is successful, doesn't check the `return` value.

modules/passkey/test/SafeWebAuthnSignerFactory.spec.ts
```diff
@@ -0,0 +1,178 @@
...
+
+      await mockVerifier.givenCalldataReturnBool(
+        ethers.solidityPacked(
+          ['bytes32', 'uint256', 'uint256', 'uint256', 'uint256'],
+          [ethers.sha256(encodeWebAuthnSigningMessage(clientData, DUMMY_AUTHENTICATOR_DATA)), r, s, x, y],
+        ),
+        true,
+      )
+
+      expect(await factory.isValidSignatureForSigner(dataHash, signature, x, y, mockVerifier)).to.equal('0x1626ba7e')
```
remedcu, line 112: Can we use the magic value here as well?

remedcu (commented):
Small nits and can we move the `r, s, x, y` and possibly `signer` also into the `setupTests`?

modules/passkey/test/SafeWebAuthnSignerFactory.spec.ts
nlordell: Hmm, to me it makes sense. We are checking that the address returned to by `getSigner` is the address where the signer will be deployed to when creating it.

So we get the address, deploy the signer, and check there is code there, i.e.:
> [`getSigner` ...] returns the address that a signer will be created on

modules/passkey/test/SafeWebAuthnSignerFactory.spec.ts
nlordell: This gets addressed in #374 already

modules/passkey/test/SafeWebAuthnSignerFactory.spec.ts
nlordell: Changed test to check deterministic deployment address.

nlordell:
> Small nits and can we move the r, s, x, y and possibly signer also into the setupTests?

the reason I didn't do this is that it is part of input selection for the methods, and not really related to setup logic.

remedcu (approved):

Issue #351, opened 2024-04-03
Publish NPM Package for 4337 and Passkey Module


nlordell:
Currently, user stories depend on unpublished code in the `modules/4337` and `modules/passkey` packages in order to work.

We should publish these packages to the NPM registry in order to facilitate the use of the 4337 and passkey modules.

## Expected Outcome

- [ ] The code required to make the user story demos work (notably, extracting public key from a WebAuthn response and encoding the WebAuthn signer signature) should move to the `src/` tree (out of the `test/` tree)
- [ ] Both `modules/4337` and `modules/passkey` packages should be published to the public NPM registry (npmjs.org).

rmeissner:
Can this issue be closed @nlordell

PR #375, opened 2024-04-09
Moving `utils` from `test -> src`

remedcu:
Part of #351

> The code required to make the user story demos work (notably, extracting the public key from a WebAuthn response and encoding the WebAuthn signer signature) should move to the src/ tree (out of the test/ tree)

This PR moves the files in the `test/utils` folder to `src/utils`. Corresponding changes to the test files also are made. This is a pre-requisite for the NPM module publishing.


modules/passkey/test/4337/WebAuthn.spec.ts
```diff
@@ -9,7 +9,7 @@ import {
 } from '@safe-global/safe-4337/src/utils/userOp'
 import { chainId } from '@safe-global/safe-4337/test/utils/encoding'
 import { Safe4337 } from '@safe-global/safe-4337/src/utils/safe'
-import { WebAuthnCredentials, decodePublicKey, encodeWebAuthnSignature } from '../utils/webauthn'
+import { WebAuthnCredentials, decodePublicKey, encodeWebAuthnSignature } from '../../src/utils/webauthn'
```
nlordell, line 12: Hmm, `webauthn.ts` has two parts:
1. a shim that should **only** be used for testing
2. utilities for interacting with WebAuthn API (extracting signatures, public key, etc.)

These need to be split and we cannot just move the whole module.

remedcu: So, it has to be mainly divided as?

`test/utils`:
- Dummy values
- `WebAuthnCredentials`

`src/utils`:
- `decodeSignature, getSignatureBytes, encodeWebAuthnSignature`
- `decodeClientDataFields, decodePublicKey, base64UrlEncode, encodeWebAuthnSigningMessage`

nlordell: Dummy values should be in `src`, they are used for signature gas estimation.

remedcu: Ok, so just `WebAuthnCredentials` should be in `test/utils`, the rest in `src/utils`, right?

nlordell: Yup, just the credential shim (and related types). I think `src` should even use the builtin TypeScript types for the WebAuthn API.

remedcu: The first part is done, the second part might need more clarification. Did you mean about using `bigint` instead of `BigNumberish`, or something else?

modules/passkey/src/utils/webauthn.ts
```diff
@@ -0,0 +1,185 @@
...
+    case 'required':
+      return 0x04
+    default:
+      throw new Error(`user verification requirement ${userVerification} not supported`)
+  }
+}
+
+export function b2ab(buf: Uint8Array): ArrayBuffer {
+  return buf.buffer.slice(buf.byteOffset, buf.byteOffset + buf.byteLength)
+}
```
nlordell, line 18: This is only used internally in the WebAuthn shim, so I would move it there.

modules/passkey/src/utils/webauthn.ts
```diff
@@ -0,0 +1,185 @@
...
+/**
+ * Encodes the given WebAuthn signature into a string. Used for testing purposes.
+ *
+ * @param authenticatorData - The authenticator data as a Uint8Array.
+ * @param clientDataFields - The client data fields as a string.
+ * @param r - The value of r as a bigint.
+ * @param s - The value of s as a bigint.
+ * @returns The encoded string.
+ */
+export function getSignatureBytes({
```
nlordell, line 134: Unrelated to this PR, but from the documentation, shouldn't this method be called 
```suggestion
export function encodeSignatureBytes({
```

Don't have to change it here, just a small detail I noticed in passing.

modules/passkey/src/utils/webauthn.ts
remedcu: Done 👍🏾

modules/passkey/src/utils/webauthn.ts
remedcu: Will keep it unresolved to change in the next PR 👍🏾

modules/passkey/src/utils/webauthn.ts
```diff
@@ -0,0 +1,185 @@
+import { ethers } from 'ethers'
+import type { BigNumberish, BytesLike } from 'ethers'
+import CBOR from 'cbor'
+
+export function userVerificationFlag(userVerification: UserVerificationRequirement = 'preferred'): number {
```
nlordell, line 5: This is missing documentation comments. It used to be only used internally, but now that it is not we should document the function.

modules/passkey/test/verifiers/FCLP256Verifier.spec.ts
```diff
@@ -2,7 +2,7 @@ import { expect } from 'chai'
 import { BigNumberish, BytesLike } from 'ethers'
 import { deployments, ethers } from 'hardhat'
 
-import { Account } from '../utils/p256'
+import { Account } from '../../src/utils/p256'
```
nlordell, line 5: The `p256` account implementation is only used for testing, I don't think it should move.

modules/passkey/src/utils/webauthn.ts
```diff
@@ -0,0 +1,185 @@
+import { ethers } from 'ethers'
+import type { BigNumberish, BytesLike } from 'ethers'
+import CBOR from 'cbor'
```
nlordell, line 3: This isn't great because now the package depends on `ethers` which ideally would be great to avoid. Is there a way to remove this dependency; i.e. can we implement some helper functions to replace the `ethers` functionality we use?

If my memory serves me well, we only use it for byte manipulation which doesn't merit pulling in the entire `ethers` dependency for.

`cobr` makes sense though.

modules/passkey/src/utils/webauthn.ts
remedcu: Done 👍🏾

modules/passkey/test/userstories/PasskeyCredentialCreation.spec.ts
```diff
@@ -1,8 +1,9 @@
 import { expect } from 'chai'
 import { deployments, ethers } from 'hardhat'
-import { WebAuthnCredentials, decodePublicKey, encodeWebAuthnSignature } from '../utils/webauthn'
+import { WebAuthnCredentials } from '../../test/utils/webauthn'
+import { decodePublicKey, encodeWebAuthnSignature } from '../../src/utils/webauthn'
 import { buildSignatureBytes } from '@safe-global/safe-4337/src/utils/execution'
-import { buildSafeTransaction, buildSafeTransactionData, SafeDomain } from '../utils/safe'
+import { buildSafeTransaction, buildSafeTransactionData, SafeDomain } from '../../src/utils/safe'
```
nlordell, line 6: The `safe` account implementation is only used for testing, I don't think it should move. Also, conceptually, tools for interacting with Safes should be implemented by the `safe-smart-account` package and not here.

modules/passkey/src/utils/webauthn.ts
```diff
@@ -0,0 +1,186 @@
+import { ethers } from 'ethers'
```
nlordell, line 1: I think the `package.json` needs to be updated in order to include the correct dependencies used by src. Note that dependencies used here cannot be `devDependencies` but need to be regular `dependencies`

modules/passkey/test/verifiers/FCLP256Verifier.spec.ts
remedcu: Done 👍🏾

modules/passkey/test/userstories/PasskeyCredentialCreation.spec.ts
remedcu: Moved back 👍🏾 

For using `safe-smart-account` (or `safe-contracts` older version), we don't have a `buildSafeTransactionData` in Safe Contracts I think on a quick look.

modules/passkey/src/utils/webauthn.ts
remedcu: `ethers` is used for:
- `getBytes`
- `concat`
- `sha256`
- `toUtf8Bytes`
- `isBytesLike`
- `encodeBase64`
- `hexlify`
- `MaxUint256`
- `AbiCoder`
- `solidityPacked`
- `toBeHex`

modules/passkey/src/utils/webauthn.ts
mmv08: > I think the `package.json` needs to be updated in order to include the correct dependencies used by src. Note that dependencies used here cannot be `devDependencies` but need to be regular `dependencies`

`peerDependencies` should work too, right? https://nodejs.org/en/blog/npm/peer-dependencies

modules/passkey/src/utils/webauthn.ts
mmv08: Maybe this will be helpful, but in the example app I implemented public key extraction without the `cbor`, using only built-in `crypto` module: https://github.com/safe-global/safe-modules/blob/89c46b2c9a27fec6ffbaefc6391742e6b76e41c3/examples/4337-passkeys/src/logic/passkeys.ts#L57-L68

(This would require double-checking that the module exists in the node.js)

modules/passkey/src/utils/webauthn.ts
```diff
@@ -0,0 +1,186 @@
...
+ *
+ * See <https://www.rfc-editor.org/rfc/rfc4648#section-5>
+ *
+ * @param data data to encode to `base64url`
+ * @returns the `base64url` encoded data as a string.
+ */
+export function base64UrlEncode(data: BytesLike | ArrayBufferLike): string {
+  const bytes = ethers.isBytesLike(data) ? data : new Uint8Array(data)
+  return ethers.encodeBase64(bytes).replace(/\+/g, '-').replace(/\//g, '_').replace(/=*$/, '')
+}
```
mmv08, line 44: can be encoded without ethers:
https://github.com/safe-global/safe-modules/blob/89c46b2c9a27fec6ffbaefc6391742e6b76e41c3/modules/passkey/test/libraries/WebAuthn.spec.ts#L5-L11

modules/passkey/test/4337/WebAuthn.spec.ts
nlordell: No, TypeScript has built-in types for the `navigator.credential...` API. I think you already did this since there are no type errors and you didn't copy over the interface definitions.

modules/passkey/src/utils/webauthn.ts
```diff
@@ -0,0 +1,186 @@
...
+ * Returns the message that gets signed by the WebAuthn credentials.
+ *
+ * See <https://w3c.github.io/webauthn/#fig-signature>
+ */
+export function encodeWebAuthnSigningMessage(
+  clientData: { type: 'webauthn.get'; challenge: string; [key: string]: unknown },
+  authenticatorData: BytesLike,
+) {
+  return ethers.getBytes(ethers.concat([authenticatorData, ethers.sha256(ethers.toUtf8Bytes(JSON.stringify(clientData)))]))
+}
```
nlordell, line 31: This is a testing function, since the signing message should only ever be computed by the `WebAuthn` implementation in the browser.

This can move to the `test/` folder and make the `src/` no longer depend on `sha256` hashing.

modules/passkey/src/utils/webauthn.ts
nlordell: `Buffer` is not a browser API AFAIK.

modules/passkey/src/utils/webauthn.ts
nlordell: Appears to be the case: https://nodejs.org/docs/latest/api/crypto.html#cryptosubtle and https://nodejs.org/docs/latest/api/webcrypto.html#class-subtlecrypto

modules/passkey/src/utils/webauthn.ts
remedcu: Moved 👍🏾

modules/passkey/src/utils/webauthn.ts
nlordell: - `getBytes`: not needed if we use `Uint8Array` everywhere
- `concat`: easy to implement
- `sha256`: shouldn't be needed, added comment :point_down: 
- `toUtf8Bytes`: Trivial with `TextEncoder` API
- `isBytesLike`: shouldn't be needed if base-64 URL encoding is implemented without `ethers`
- `encodeBase64`: shouldn't be hard, there are lots of implementations floating around the internet.
- `hexlify`: should be trivial to implement
- `MaxUint256`: trivial
- `AbiCoder`: Should be implemented in function of concatenation, we don't need general ABI encoding support
- `solidityPacked`: idem
- `toBeHex`: should be easy to implement.

I also have a feeling that many of these methods will no longer be needed once you start de-ethers-ifying things.

modules/passkey/package.json
```diff
@@ -64,6 +63,7 @@
     }
   },
   "dependencies": {
+    "cbor": "^9.0.2",
     "@safe-global/safe-contracts": "^1.4.1-build.0"
```
nlordell, line 67:
```suggestion
    "@safe-global/safe-contracts": "^1.4.1-build.0",
    "cbor": "^9.0.2"
```

NPM automatically orders dependencies when adding/removing them with the tool, so otherwise this will generate noise.

Also make sure to NPM install after updating the package JSON, otherwise the lock file will be out of date.

remedcu: Done 👍🏾

modules/passkey/src/utils/webauthn.ts
mmv08: Right! I just found this, seems like there are browser's built-in functions: https://www.digitalocean.com/community/tutorials/how-to-encode-and-decode-strings-with-base64-in-javascript#encoding-and-decoding-strings-with-base64

modules/passkey/src/utils/webauthn.ts
remedcu: Done 👍🏾

modules/passkey/src/utils/helper.ts
```diff
@@ -0,0 +1,207 @@
+/**
```
nlordell, line 1: I don't think we need to just copy-paste the `ethers` code in here:
- We can use stricter types (like `bigint` and `Uint8Array`) in the helper methods, instead of converting things to and from hex for example.
- I don't think we should implement `encodeABI`. We should just manually encode the things in the one place we use it (we have deterministic types, and do not need a general ABI encoding function).
- Buffer isn't supported in the browser, so maybe we shouldn't use it here - and if we do, I don't think we should encode as base64 just to manually convert it to base64url when it supports base64url encoding directly.

modules/passkey/src/utils/webauthn.ts
nlordell: It would... but I wanted to avoid using `ethers` if we don't need it because:
- it pulls in a lot of stuff
- v5 vs v6 shenanigans

modules/passkey/src/utils/webauthn.ts
```diff
@@ -0,0 +1,168 @@
...
+].join(',')
+
+/**
+ * Dummy authenticator data. This can be used for gas estimations, as it ensures that the correct
+ * authenticator flags are set.
+ */
+export const DUMMY_AUTHENTICATOR_DATA = fromHexString(
+  '0x' + MaxUint256.toString(16) + '0' + userVerificationFlag('required') + 'ffffffff',
+  // `0x` for Hex, added a `0` for userVerification as it returns `number` and removed the `0x` from the sign count.
+)
```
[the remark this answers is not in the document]
remedcu: Done 👍🏾

modules/passkey/src/utils/helper.ts
remedcu: Done 👍🏾

modules/passkey/src/utils/webauthn.ts
```diff
@@ -0,0 +1,200 @@
...
+  }
+}
+
+/**
+ * Encode bytes using the Base64 URL encoding.
+ *
+ * See <https://www.rfc-editor.org/rfc/rfc4648#section-5>
+ *
+ * @param data data to encode to `base64url`
+ * @returns the `base64url` encoded data as a string.
```
nlordell, line 25: I think other methods are missing the `@param` stuff.

modules/passkey/src/utils/webauthn.ts
```diff
@@ -0,0 +1,200 @@
...
+// Authenticator data is the concatenation of:
+// - 32 byte SHA-256 hash of the relying party ID
+// - 1 byte for the user verification flag
+// - 4 bytes for the signature count
+// We fill it all with `0xfe` and set the appropriate user verification flag.
+DUMMY_AUTHENTICATOR_DATA.fill(0xfe)
+DUMMY_AUTHENTICATOR_DATA[32] = userVerificationFlag('required')
+
+/**
+ *  Returns a Hex from the %%bytes%% (Uint8Array).
```
nlordell, line 196:
```suggestion
 * Returns the hexadecimal encoding of the specified {@link Uint8Array}.
```

nlordell (commented):
Everything looks great, just doc comments missing some consistency.

modules/passkey/src/utils/webauthn.ts
remedcu: Done 👍🏾

modules/passkey/src/utils/webauthn.ts
remedcu: Updated 👍🏾

modules/passkey/test/utils/webauthn.ts
```diff
@@ -9,8 +9,25 @@
 
 import { p256 } from '@noble/curves/p256'
```
nlordell, line 10: Last nit, I promise :sweat_smile: - now that this only has the WebAuthn shim code, I think we should `git mv modules/passkey/test/utils/webauthn.ts modules/passkey/test/utils/webauthnShim.ts`

nlordell (approved):

modules/passkey/test/utils/webauthn.ts
remedcu: Done 😅

Issue #308, opened 2024-03-06
Add a proper documentation to the `Safe256BitECSignerLaunchpad` contract


mmv08:
## Context / issue

The migration of the experimental `Safe256BitECSignerLaunchpad` contract in https://github.com/safe-global/safe-modules/pull/306 wasn't complete. It missed `NatSpec` and a proper testing suite.

## Proposed solution

- Complete the `NatSpec` documentation

## Alternatives

- Since the contract is not "mandatory" for the release of the passkey signer, we can also make it explicitly experimental and not include it in the release/audit scope

nlordell:
The work to add 100% test coverage to the launchpad contract was moved to #380 as it is blocked on #372 and #377

rmeissner:
Should this issue be closed @nlordell ?

nlordell:
Yes it can

PR #376, opened 2024-04-10
Add NatSpec Documentation To The Safe Launchpad

nlordell:
Related to #308

This PR adds NatSpec documentation to the `SafeSignerLaunchpad` contract for all methods. This is done in preparation for an auditing these contracts and as a precursor to getting 100% test coverage.

In writing this PR, I found some issues with the current launchpad implementation that need to be addressed pre-audit:
- We allow accounts that haven't been properly initialized, which can be a security risk. We should do checks similar to the Safe `setup` method to prevent double initialization and the validation should revert if the account is not initialized.
- We are using revert messages instead of errors.

Additionally, I would like to suggest that we change the implementation to use a **single** setup instead of doing a pre- and post- setup delegatecall. This would greatly simplify the implementation and security assumptions of the launchpad. The downside is that the setup cannot break 4337 rules, but we do not have a use-case for this, so I wouldn't implement it.

### DIFF
--- a/modules/passkey/contracts/4337/SafeECDSASignerLaunchpad.sol
+++ b/modules/passkey/contracts/4337/SafeECDSASignerLaunchpad.sol
@@ -12,29 +12,42 @@ import {ERC1271} from "../libraries/ERC1271.sol";
 
 /**
  * @title Safe Launchpad for Custom ECDSA Signing Schemes.
- * @dev A launchpad account implementation that enables the creation of Safes that use custom ECDSA
- * signing schemes that require additional contract deployments over ERC-4337.
+ * @dev A launchpad account implementation that enables the creation of Safes that use custom ECDSA signing schemes that
+ * require additional contract deployments over ERC-4337. Note that it is not safe to rely on this launchpad for
+ * deploying Safes that has an initial threshold greater than 1. This is because the first user operation (which can
+ * freely change the owner structure) will only ever require a single signature to execute, so effectively the initial
+ * owner will always have ultimate control over the Safe during the first user operation and can undo any changes to the
+ * `threshold` during the `setup` phase.
  * @custom:security-contact bounty@safe.global
  */
 contract SafeECDSASignerLaunchpad is IAccount, SafeStorage {
-    bytes32 private constant DOMAIN_SEPARATOR_TYPEHASH = keccak256("EIP712Domain(uint256 chainId,address verifyingContract)");
+    /**
+     * @notice The EIP-712 type-hash for the domain separator used for verifying Safe initialization signatures.
+     * @custom:computed-as keccak256("EIP712Domain(uint256 chainId,address verifyingContract)")
+     */
+    bytes32 private constant DOMAIN_SEPARATOR_TYPEHASH = 0x47e79534a245952e8b16893a336b85a3d9ea9fa8c573f3d803afb92a79469218;
 
-    // keccak256("SafeSignerLaunchpad.initHash") - 1
-    uint256 private constant INIT_HASH_SLOT = 0x1d2f0b9dbb6ed3f829c9614e6c5d2ea2285238801394dc57e8500e0e306d8f80;
+    /**
+     * @notice The storage slot used for the initialization hash of account.
+     * @custom:computed-as keccak256("SafeSignerLaunchpad.initHash") - 1
+     * @dev This value is intentionally computed to be a hash -1 as a precaution to avoid any potential issues from
+     * unintended hash collisions.
+     */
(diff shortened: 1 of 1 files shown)


modules/passkey/contracts/4337/SafeECDSASignerLaunchpad.sol
```diff
@@ -164,55 +224,31 @@ contract SafeECDSASignerLaunchpad is IAccount, SafeStorage {
...
+     * @param signerFactory The custom ECDSA signer factory to use for creating an owner.
+     * @param signerX The X coordinate of the signer's public key.
+     * @param signerY The Y coordinate of the signer's public key.
+     * @param signerVerifier The address of the contract that verifies the signer's signature.
+     * @param setupTo The contract to `DELEGATECALL` during setup.
+     * @param setupData The calldata for the setup `DELEGATECALL`.
+     * @param fallbackHandler The fallback handler to initialize the Safe with.
+     * @param callData The calldata to `DELEGATECALL` self with in order to actually execute the user operation.
      */
-    function _validateSignatures(
```
nlordell, line 186: function was moved down with the other internal methods.

modules/passkey/contracts/4337/SafeECDSASignerLaunchpad.sol
```diff
@@ -243,37 +279,77 @@ contract SafeECDSASignerLaunchpad is IAccount, SafeStorage {
...
+            // The timestamps are validated by the entry point, therefore we will not check them again
+            validationData = _packValidationData(magicValue != ERC1271.MAGIC_VALUE, validUntil, validAfter);
+        } catch {
+            validationData = _packValidationData(true, validUntil, validAfter);
+        }
     }
 
+    /**
+     * @notice Reads the configured initialization hash from storage.
+     */
```
mmv08, line 330: nit: should this specify the return value type? similar question about `isContract`

nlordell: 82ba1ef19f95fbbd8a382fc96a08b611ba27bc08

modules/passkey/contracts/4337/SafeECDSASignerLaunchpad.sol
```diff
@@ -12,90 +12,129 @@ import {ERC1271} from "../libraries/ERC1271.sol";
...
+     * @custom:computed-as keccak256("EIP712Domain(uint256 chainId,address verifyingContract)")
+     */
+    bytes32 private constant DOMAIN_SEPARATOR_TYPEHASH = 0x47e79534a245952e8b16893a336b85a3d9ea9fa8c573f3d803afb92a79469218;
 
-    // keccak256("SafeSignerLaunchpad.initHash") - 1
-    uint256 private constant INIT_HASH_SLOT = 0x1d2f0b9dbb6ed3f829c9614e6c5d2ea2285238801394dc57e8500e0e306d8f80;
+    /**
+     * @notice The storage slot used for the initialization hash of account.
+     * @custom:computed-as keccak256("SafeSignerLaunchpad.initHash") - 1
+     * @dev This value is intentionally computed to be a hash -1 as a precaution to avoid any potential issues from
```
remedcu, line 33: Subjective-nit:

```suggestion
     * @dev This value is intentionally computed to be a hash - 1 as a precaution to avoid any potential issues from
```

modules/passkey/contracts/4337/SafeECDSASignerLaunchpad.sol
```diff
@@ -12,90 +12,129 @@ import {ERC1271} from "../libraries/ERC1271.sol";
...
      *  {uint48} validUntil - A timestamp representing until when the user operation is valid, or 0 to indicated "forever".
      *  {address} entryPoint - The address of the entry point that will execute the user operation.
+     * @custom:computed-as keccak256("SafeInitOp(bytes32 userOpHash,uint48 validAfter,uint48 validUntil,address entryPoint)")
      */
-    bytes32 private constant SAFE_INIT_OP_TYPEHASH =
-        keccak256("SafeInitOp(bytes32 userOpHash,uint48 validAfter,uint48 validUntil,address entryPoint)");
+    bytes32 private constant SAFE_INIT_OP_TYPEHASH = 0x25838d3914a61e3531f21f12b8cd3110a5f9d478292d07dd197859a5c4eaacb2;
 
+    /**
+     * @dev Address of the launchpad contract itself. it is used for determining whether or not the contract is being
```
remedcu, line 63:
```suggestion
     * @dev Address of the launchpad contract itself. It is used for determining whether or not the contract is being
```

remedcu (approved):

mmv08 (approved):

Issue #372, opened 2024-04-09
`SafeECDSASignerLaunchpad` has unclear support for multi-owner accounts


mmv08:
## Description

When working on https://github.com/safe-global/safe-modules/issues/360, we wanted to add two signers to the newly created Safe account: the user's connected wallet and the created passkey signer. However, this turned out to be non-trivial because `SafeECDSASignerLaunchpad` has unclear/limited support for multi-owner accounts.

1. There's no straightforward way to specify multiple owners for the Safe configuration.
When setting up the Safe, the contract hardcodes the signer/threshold parameters, leaving the only option to extend the configuration in the call executed later in the function or the optional delegatecall in the Safe setup phase.
https://github.com/safe-global/safe-modules/blob/5a5f7f6984a21e42e28d24bf525547be5c1068fa/modules/passkey/contracts/4337/SafeECDSASignerLaunchpad.sol#L228-L233
2. During the validation phase, it only validates the signature from one (Passkey) owner. This is a potential security risk since a single signer has full control over a Safe.

## Additional context

This issue was created to explore potential solutions and decide where we want to improve this situation. While briefly discussing it in a standup meeting, we agreed that solving it would be complex because it would require duplicating the whole  Safe signature check logic in the contract (one of the solutions).

nlordell:
Correct me if I'm wrong, but would you say that the expected outcome would be to:
1. evaluate the security implications of the current setup? (Basically, what you said in the standup where we a single owner has full control of the multisig)
2. propose changes to the API to adjust for potential multi-owner use cases?

mmv08:
> Correct me if I'm wrong, but would you say that the expected outcome would be to:
> 
> 1. evaluate the security implications of the current setup? (Basically, what you said in the standup where we a single owner has full control of the multisig)
> 2. propose changes to the API to adjust for potential multi-owner use cases?

Bang on! 

Something that came to me while writing the issue: IIRC, the current launchpad contract was your experiment to make Passkey signers work with 4337, and we never defined any formal requirements (product, technical, etc) for the contract. Perhaps this is the moment to do this, sync with Niharika and formalize the requirements, etc

nlordell:
This is blocked on product decision on what must be supported for SDK passkey support.

nlordell:
We found a general solution with the shared signer, which would work for arbitrary ownership structures, so this is once again unblocked.

PR #377, opened 2024-04-10
Rework Safe Signer Launchpad

nlordell:
This PR does a rework of some of the implementation details of the `SafeSignerLaunchpad` contract in light of some observations from the previous PR #376.

Namely, this changes the initialization process to work in a slightly different way:
1. Set the target singleton to a special slot when the entry point executes the `initCode` for the account. Safe `setup` also happens at this point, meaning that any `DELEGATECALL` to the Safe singleton should work and be valid.
2. Signature verification checks that the account is an owner. This has the side-effect that you can initialize an account with multiple custom ECDSA owners and use any of them to sign the first user operation.
3. Promote the Safe to the singleton that was previously in storage.

The main difference with the previous flow is that we no longer have two separate `setup` initializers that we `DELEGATECALL` to. Additionally, we added checks that prevent double initialization as well as reentrency issues in the execution function.

In addition, this also opens up a pretty clear path for supporting multiple owners with the launchpad as the account has already undergone "regular" Safe setup. This is relevant for #372.

Unit tests in order to reach 100% coverage will be introduced in a follow up.


modules/passkey/contracts/4337/SafeECDSASignerLaunchpad.sol
```diff
@@ -201,85 +348,65 @@ contract SafeECDSASignerLaunchpad is IAccount, SafeStorage {
...
-    function _initHash() public view returns (bytes32 value) {
+    /**
+     * @notice Reads the configured target Safe singleton address to promote to from storage.
+     * @return value The target Safe singleton address.
+     */
+    function _targetSingleton() internal view returns (address value) {
         // solhint-disable-next-line no-inline-assembly
         assembly ("memory-safe") {
-            value := sload(INIT_HASH_SLOT)
+            value := and(sload(TARGET_SINGLETON_SLOT), 0xffffffffffffffffffffffffffffffffffffffff)
```
mmv08, line 383: does it need `and` op here? remember https://github.com/safe-global/safe-smart-account/pull/737?

modules/passkey/contracts/4337/SafeECDSASignerLaunchpad.sol
```diff
@@ -12,168 +12,315 @@ import {ERC1271} from "../libraries/ERC1271.sol";
...
+     * @param initializerData The calldata for the setup `DELEGATECALL`.
+     * @param fallbackHandler Handler for fallback calls to the Safe.
      */
-    function getInitHash(
+    function setup(
         address singleton,
         address signerFactory,
         uint256 signerX,
         uint256 signerY,
         address signerVerifier,
```
mmv08, line 169: idea for multiple owners: make this an array of structs and add a separate ecdsa owner array (just realised that the main bottleneck is still the signature verification but still leacing this here)

In general I like the direction of where this is going.

modules/passkey/contracts/4337/SafeECDSASignerLaunchpad.sol
nlordell: I did this because I don't know if its "legal" to have the `address value` field hold a value that isn't sanitized.

mmv08: So I played with the following contract in remix:
```
// SPDX-License-Identifier: Apache-2.0
pragma solidity ^0.8.0;

contract Test {
    address internal singleton;

    constructor(address _singleton) {
        singleton = _singleton;
    }

    function _targetSingleton() public view returns (address value) {
        // solhint-disable-next-line no-inline-assembly
        assembly ("memory-safe") {
            value := sload(0)
        }
    }
}
```
compiler version 0.8.25, 200 optimizer runs. 

I think the compiler still generates the sanitizing code, and if you have the `and` op, it will execute sanitizing twice (in this particular case, I tested with no optimiser enabled)

Here's the screenshot from debugging the call where the sanitising is executed:
Screenshot 2024-04-22 at 14 48 31

nlordell: Ah, right. It looks like it masks the value on access. Thanks for checking.

modules/passkey/contracts/4337/SafeECDSASignerLaunchpad.sol
nlordell: Captured in issue #399

nlordell:
@remedcu and @akshay-ap - this is ready for review.

akshay-ap (approved):

akshay-ap:
Approved. The documentation needs to be updated as per new implementation which will be handled in a separate PR.

nlordell:
> The documentation needs to be updated as per new implementation which will be handled in a separate PR.

Documentation updated in #415

Issue #361, opened 2024-04-04
Add FAQ Section to Example Application


mmv08:
We require the addition of a FAQ section to our example application in the examples/ directory. The FAQ section should provide comprehensive documentation to assist users in understanding and utilizing the application effectively. The expected outcome is to enhance user experience and clarify common queries. Key objectives include:

1. Creating a dedicated FAQ section within the application.
2. Offering detailed explanations and solutions to frequently asked questions.
3. Ensuring the FAQ section is easily accessible and navigable for users seeking assistance.

Example: https://smart-wallet.xyz

Exact FAQ contents TBD

Alternative:
- Add a faq section to the GitHub documentation

mmv08:
I’m currently looking at the example (https://smart-wallet.xyz/) and their FAQ section covers three main topics:
1. Basics of smart wallets
2. Integration of smart wallets into dapps
3. Passkeys, passkeys as a single point of failure and recovery

I’d say that for our example application topics 1 and 2 are out of scope.
For topic 3, they’re discussing the importance of having alternative recovery options, which we currently do not support at deployment time and do not have a way to add a recovery account.
Therefore, I’d propose either dropping this issue OR waiting until the multisig flow is supported in the launchpad OR adding a way to add a recovery signer to the app instead. WDYT?

PR #378, opened 2024-04-10
Passkey example application: A Major refactoring

mmv08:
This PR lays the foundation for https://github.com/safe-global/safe-modules/issues/360 and https://github.com/safe-global/safe-modules/issues/361.

**Motivation**
The example application was built as a mere example, and all of the state/screen management was done in a single file—` src/App.ts`—through tens of different boolean and state variables. Since we have now decided to add more features to the app, the approach is no longer scalable.

**Solution**
Isolate every step in the deployment flow - in other words, make every step in a process its route. This should also help the developers to comprehend the example since now, to understand what is needed to create a passkey, they only have to review the `CreatePasskey` route, not mixed with anything else.

**This PR**
- Add `react-router-dom` to manage the routing in the application
- Create a route for every step in the process
- Introduce a new Safe Dashboard page with a link to the Safe{Wallet} App (This page will be used for #360, the new action will be added here)

**Demo**
https://www.loom.com/share/b5dcec589cc44fd8be8cefe1ef9cb306


examples/4337-passkeys/src/routes/CreatePasskey.tsx
```diff
@@ -0,0 +1,55 @@
...
+    }
+  }
+
+  return (
+    <div className="card">
+      <p>First, create a passkey to sign transactions</p>
+      <p style={{ opacity: '0.75' }}>
+        Passkey is a secure authentication method that replaces traditional passwords. It uses public key cryptography and unique
+        cryptographic keys to log users into websites and apps securely. Users can unlock their device using biometrics like a fingerprint,
+        Face ID, or a device PIN. This makes passkeys much more secure than passwords and resistant to phishing attacks.
```
nlordell, line 42: Maybe add a blurb about how the credential will be used (i.e. it will be used to verify user operations on chain, blah, blah, etc,) - in other words give more context as to _why_ we are creating the passkey.

examples/4337-passkeys/src/main.tsx
```diff
@@ -1,10 +1,30 @@
...
+import { Root } from './routes/Root.tsx'
+import { Home, loader as homeLoader } from './routes/Home.tsx'
+import { DeploySafe, loader as deploySafeLoader } from './routes/DeploySafe.tsx'
+import { CreatePasskey } from './routes/CreatePasskey.tsx'
+import { CREATE_PASSKEY, DEPLOY_SAFE, SAFE } from './routes/constants.ts'
+import { Safe, loader as safeLoader } from './routes/Safe.tsx'
+
+const router = createBrowserRouter([
+  {
+    path: '/',
```
nlordell, line 18: Shouldn't this be:
```suggestion
    path: HOME,
```

nlordell: (disclaimer: I'm react noob)

examples/4337-passkeys/src/routes/Home.tsx
```diff
@@ -0,0 +1,71 @@
...
+}
+
+// This page doesn't have a UI, it just determines where to redirect the user based on the state.
+function Home() {
+  const { safeAddress, passkey } = useLoaderData() as LoaderData
+  const { walletProvider } = useOutletContext<OutletContext>()
+  const [safeCode, requestStatus] = useCodeAtAddress(walletProvider, safeAddress)
+
+  if (requestStatus === RequestStatus.SUCCESS && safeCode !== '0x') {
+    return <Navigate to={getSafeRoute(safeAddress)} />
```
nlordell, line 61: Why do you sometimes redirect in the loader, but for other cases redirect in the component?

nlordell: Is it because this comes from the `useCodeAtAddress(walletProvider, safeAddress)` which is `async`?

nlordell (approved):
LGTM. Just one question.

examples/4337-passkeys/src/routes/Home.tsx
mmv08: That's correct, but probably not for the reason you think. For the async data, we need to use the user's RPC provider because we don't have our own one, but the `web3modal` doesn't expose a reliable method to get it except a react hook. I tried to use the `web3modal` object directly, but it was not reliable and would fail in cases where the user doesn't have a wallet connected.

There might be a workaround where we retrieve the provider higher up in the component hierarchy and pass it to the router itself. However, I decided not to waste time on trying that since it's still primarily an example app :smile:

examples/4337-passkeys/src/routes/CreatePasskey.tsx
mmv08: Check the new copy :)

mmv08:
Just so you know, I requested a review from a React expert, @iamacook. He said he would have time to check tomorrow, so let's hold off on the merge until he approves it.

examples/4337-passkeys/src/routes/DeploySafe.tsx
```diff
@@ -20,12 +21,42 @@ import {
...
+  }
+  const initHash = getInitHash(initializer, APP_CHAIN_ID)
+  const launchpadInitializer = getLaunchpadInitializer(initHash)
+  const safeAddress = getSafeAddress(launchpadInitializer)
+
+  return { passkey, safeAddress }
+}
+
+function DeploySafe() {
+  const { passkey, safeAddress } = useLoaderData() as { passkey: PasskeyLocalStorageFormat; safeAddress: string }
```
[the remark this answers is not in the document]
mmv08: Unfortunately, this doesn't work because the loader/useLoaderData doesn't accept any type arguments, and the return type of the loader function is either the data or a response type (redirect, for example)

examples/4337-passkeys/src/routes/constants.ts
```diff
@@ -0,0 +1,8 @@
+const HOME = '/'
+const CREATE_PASSKEY = '/create-passkey'
+const DEPLOY_SAFE = '/deploy-safe'
+const SAFE = '/safe/:safeAddress'
```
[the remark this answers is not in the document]
mmv08: done (although i think it was enough cos it was reported from the routes "package")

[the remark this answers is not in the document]
nlordell: > although i think it was enough cos it was reported from the routes "package")

I tend to agree.

You could express it as:

```ts
import * as routes from "path/to/routes.ts"

routes.HOME
routes.SAFE
// etc.
```

PR #379, opened 2024-04-11
Support Precompile With Fallback For Signers

nlordell:
This PR changes the internal workings of the signers to accept a uint192 instead of an address as the `verifier` (well, now `verifiers`) parameter. This allows a signer to encode BOTH a precompile address as well as a Solidity implementation fallback for the P-256 curve.

Additionally, this required changing the interface of the `ICustomECDSASignerFactory`. It is now very much tied to the P-256 implementation that we have, and as such I renamed things to just `ISafeSignerFactory` and `SafeSignerLaunchpad`, since it is not a general ECDSA solution, but one that is particular to the P-256 implementation that we have and in particular supporting P-256 precompile with Solidity fallback.

The PR changes **a lot** of files, but this is mostly small modifications to tests. Unfortunately, I don't really see a way around this without pushing a PR with a broken test suite which I did not want to do.

Overall, I think this is definitely a nice feature to have - as it adds a small overhead on top of the regular verification but allows accounts to seamlessly start using precompile on networks that don't have it available yet.

### DIFF
--- a/modules/passkey/contracts/4337/SafeSignerLaunchpad.sol
+++ b/modules/passkey/contracts/4337/SafeSignerLaunchpad.sol
@@ -41,7 +41,7 @@ contract SafeSignerLaunchpad is IAccount, SafeStorage {
      *  {address} signerFactory - The custom ECDSA signer factory to use for creating an owner.
      *  {uint256} signerX - The X coordinate of the public key of the custom ECDSA signing scheme.
      *  {uint256} signerY - The Y coordinate of the public key of the custom ECDSA signing scheme.
-     *  {uint256} signerVerifiers - The ECDSA verifier contract used by the custom signing scheme.
+     *  {uint256} signerVerifiers - The P-256 verifiers to use for signature validation.
      *  {address} setupTo - The contract to `DELEGATECALL` during setup.
      *  {bytes} setupData - The calldata for the setup `DELEGATECALL`.
      *  {address} fallbackHandler - The fallback handler to initialize the Safe with.
@@ -132,7 +132,7 @@ contract SafeSignerLaunchpad is IAccount, SafeStorage {
      * @param signerFactory The custom ECDSA signer factory to use for creating an owner.
      * @param signerX The X coordinate of the signer's public key.
      * @param signerY The Y coordinate of the signer's public key.
-     * @param signerVerifiers The address of the contract that verifies the signer's signature.
+     * @param signerVerifiers The P-256 verifiers to use for signature validation.
      * @param setupTo The contract to `DELEGATECALL` during setup.
      * @param setupData The calldata for the setup `DELEGATECALL`.
      * @param fallbackHandler The fallback handler to initialize the Safe with.
@@ -252,7 +252,7 @@ contract SafeSignerLaunchpad is IAccount, SafeStorage {
      * @param signerFactory The custom ECDSA signer factory to use for creating an owner.
      * @param signerX The X coordinate of the signer's public key.
      * @param signerY The Y coordinate of the signer's public key.
-     * @param signerVerifiers The address of the contract that verifies the signer's signature.
+     * @param signerVerifiers The P-256 verifiers to use for signature validation.
      * @param setupTo The contract to `DELEGATECALL` during setup.
      * @param setupData The calldata for the setup `DELEGATECALL`.
      * @param fallbackHandler The fallback handler to initialize the Safe with.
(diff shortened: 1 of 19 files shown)


modules/passkey/contracts/4337/SafeSignerLaunchpad.sol
```diff
@@ -243,7 +252,7 @@ contract SafeECDSASignerLaunchpad is IAccount, SafeStorage {
      * @param signerFactory The custom ECDSA signer factory to use for creating an owner.
      * @param signerX The X coordinate of the signer's public key.
      * @param signerY The Y coordinate of the signer's public key.
-     * @param signerVerifier The address of the contract that verifies the signer's signature.
+     * @param signerVerifiers The address of the contract that verifies the signer's signature.
```
akshay-ap, line 255:
```suggestion
     * @param signerVerifiers The packed value containing precompile and fallback addresses.
```

modules/passkey/contracts/SafeWebAuthnSigner.sol
```diff
@@ -22,27 +21,29 @@ contract SafeWebAuthnSigner is SignatureValidator {
...
 
     /**
      * @dev Constructor function.
      * @param x The X coordinate of the P-256 public key of the WebAuthn credential.
      * @param y The Y coordinate of the P-256 public key of the WebAuthn credential.
-     * @param verifier The P-256 verifier to use for signature validation. It MUST implement the
-     * same interface as the EIP-7212 precompile.
+     * @param verifiers The P-256 verifiers to use for signature validation. This is the
+     * concatenation of `uint32(precompile) || address(fallback)` specifying the `precompile` to
+     * use as well as the `fallback` Solidity P-256 verifier implementation in case the precompile
```
remedcu, line 34:
```suggestion
     * use the `fallback` Solidity P-256 verifier implementation in case the precompile
```

modules/passkey/contracts/libraries/P256.sol
```diff
@@ -94,4 +130,33 @@ library P256 {
...
+        uint256 s,
+        uint256 x,
+        uint256 y
+    ) internal view returns (bool success) {
+        address precompileVerifier = address(uint160(uint256(Verifiers.unwrap(verifiers)) >> 160));
+        address fallbackVerifier = address(uint160(Verifiers.unwrap(verifiers)));
+        if (precompileVerifier != address(0)) {
+            success = verifySignatureAllowMalleability(IP256Verifier(precompileVerifier), message, r, s, x, y);
+        }
+        if (!success && fallbackVerifier != address(0)) {
```
remedcu, line 158: To confirm: Won't this check the signature twice in case of a wrong signature even if the precompile is now available?

Though don't think it could be a problem, as in case of signature being correct, it only checks once 👍🏾

modules/passkey/test/SafeWebAuthnSignerFactory.spec.ts
```diff
@@ -12,38 +13,42 @@ describe('SafeWebAuthnSignerFactory', () => {
...
-      const signer = await factory.getSigner(x, y, mockVerifier)
+      const signer = await factory.getSigner(x, y, verifiers)
 
       for (const params of [
-        [ethers.id('publicKey.otherX'), y, mockVerifier],
-        [x, ethers.id('publicKey.otherY'), mockVerifier],
-        [x, y, '0x0000000000000000000000000000000000000100'],
+        [ethers.id('publicKey.otherX'), y, verifiers],
+        [x, ethers.id('publicKey.otherY'), verifiers],
+        [x, y, '0xfefefefefefefefefefefefefefefefefefefefe'],
```
remedcu, line 51: nit:
```suggestion
        [x, y, `0x${'fe'.repeat(20)}`],
```

remedcu (approved):

mmv08 (approved):

modules/passkey/contracts/libraries/P256.sol
nlordell: Yes, this is correct. I will clarify why in a comment, but basically, it is because the P-256 precompile does not differentiate between invalid signature and not being available (notably, in both cases the empty bytes `bytes("")` are returned).

modules/passkey/contracts/4337/SafeSignerLaunchpad.sol
nlordell: Nice catch thanks. Fixed in multiple places in the file.

modules/passkey/contracts/SafeWebAuthnSigner.sol
nlordell: Hmm, I think it is correct as I wrote it. When I read it out loud it made sense as I wrote it at least.

modules/passkey/test/SafeWebAuthnSignerFactory.spec.ts
nlordell: 261cd37159fe86d805dc21dd774cc8375ec0c831

modules/passkey/contracts/4337/SafeSignerLaunchpad.sol
nlordell: 261cd37159fe86d805dc21dd774cc8375ec0c831

PR #297, opened 2024-03-04
Create Passkey NPM Project

nlordell:
Fixes #282 

This PR adds an empty NPM package for the Passkey signer "module", including:

- CI configuration
- Default Hardhat template
- `hardhat-deploy` configuration
- Linting

Issue #348, opened 2024-04-03
Initial Design for Safe Delay-Based Recovery Module


nlordell:

mmv08:
Do you think existing options like Zodiac should be considered? I think the Wallet team uses the delay module for the recovery

Sednaoui:
If you would like to consider the safe recovery module that candide has built. The module supports: 

- A Recovery Delay Period
- An Owner can add or remove guardians at anytime
- Guardians can remian private with fresh addresses
- Each recovery contact can submit their signature transactions separately. Once the threshold is meant, anyone can call execute recovery to start the grace period.
- Anyone can execute the guardians aggregated signatures. The last guardian can execute the transaction directly and pay gas fees. This is similar to how Safe's interface works when multiple owners for a multisig sign transactions before submitting them.

Here is the relevant links: 

- [walkthrough](https://docs.candide.dev/blog/making-accounts-recoverable/)
- [smart contracts](https://github.com/candidelabs/CandideWalletContracts/blob/main/contracts/modules/social_recovery/SocialRecoveryModule.sol)

As part of SGP, we're integrating this module into _abstractionkit_. We've also built a version for Safe{core} protocol that we have yet to publish.

nlordell:
@remedcu - can you clarify what our conclusion here was so it is documented in GitHub?

remedcu:
The Safe Team have created an MVP for designing the Safe Recovery Module for this task on the Internal Notion Page. But based on the comments above, we are also exploring other viable options like Candide Recovery Module, Delay Based Module, etc. A final decision will be shared in #352 once things are finalized based on the requirements and the currently available solutions.

As this issue was intended to create a general design and additionally check the current ecosystem products, this issue is considered complete.

PR #383, opened 2024-04-16
Safe Recovery Initial Setup

### LINKED ISSUE #350: Setup NPM Package For Recovery Module
This issue captures the work for setting up a new NPM workspace package for the Passkey contracts.

The new package should include basic CI items:
- Empty `hardhat` setup
- Formatting and linting
- Testing (although, this can be a no-op at first)

remedcu:
This PR adds an empty NPM package for the Recovery Module, including:

- CI configuration
- Default Hardhat template
- `hardhat-deploy` configuration
- Linting

Inspired from #297, this PR fixes #350

### DIFF
--- a/.github/workflows/ci_recovery.yml
+++ b/.github/workflows/ci_recovery.yml
@@ -1,0 +1,35 @@
+name: safe-modules-recovery
+on:
+  push:
+    paths:
+      - 'modules/recovery/**'
+
+jobs:
+  tests:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 20.x
+          cache: npm
+          cache-dependency-path: package-lock.json
+      - run: npm ci
+      - run: npm run coverage -w modules/recovery
+      - uses: coverallsapp/github-action@master
+        with:
+          path-to-lcov: modules/recovery/coverage/lcov.info
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 20.x
+          cache: npm
+          cache-dependency-path: package-lock.json
+      - run: npm ci
+      - run: npm run lint -w modules/recovery
+      - run: npm run fmt:check -w modules/recovery
+      - run: npm run build -w modules/recovery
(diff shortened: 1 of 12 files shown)


mmv08 (changes requested):
Did we decide that we need a new module instead of choosing plenty of existing options? I cannot see any decision made in https://github.com/safe-global/safe-modules/issues/348

remedcu:
> Did we decide that we need a new module instead of choosing plenty of existing options? I cannot see any decision made in #348

Even though not as the first milestone, we want to support the 4337 flow sooner or later, which I believe the existing options do not support.

CC @nlordell

mmv08:
> > Did we decide that we need a new module instead of choosing plenty of existing options? I cannot see any decision made in #348
> 
> Even though not as the first milestone, we want to support the 4337 flow sooner or later, which I believe the existing options do not support.
> 
> CC @nlordell

I took a look at candide contracts and they seem to fit our requirements, except for the 4337. Theoretically, we may just inherit our contract from it and add a `validateUserOp` function - was that option explored?

cc @sednaoui

Sednaoui:
> > > Did we decide that we need a new module instead of choosing plenty of existing options? I cannot see any decision made in #348
> > 
> > 
> > Even though not as the first milestone, we want to support the 4337 flow sooner or later, which I believe the existing options do not support.
> > CC @nlordell
> 
> I took a look at candide contracts and they seem to fit our requirements, except for the 4337. Theoretically, we may just inherit our contract from it and add a `validateUserOp` function - was that option explored?
> 
> cc @Sednaoui

I am not sure I understood the 4337 requirements needed for the Recovery Module. Can you please further clarify? From our knowledge: 

- Candide recovery module works for any safe, including ones with the 4337Module. 
- When the owner interacts with the recovery module, if the owner is a 4337 account, then the account normally `validatesUserOp` before they can call the recovery module. The recovery module only checks that the call is coming from the account itself.
- When a guardian interacts with the module, they interact with it just as any other contract on chain. 

Having `validateUserOp` can work for Safe{Core} **Protocol** because they do have signature validators, if the goal for example for guardians to initiate and finalize the recovery using the owner's account (owner pays the gas and is the sender), then the recovery module can act both as a plugin and a signature validator. That is not possible with just 4337 Safe Module, so I don't understand the reasoning to include `validateUserOp` in the recovery module.

nlordell:
For some additional context, we are considering adding "signature validator" support to the 4337 module to allow modules to be executed over 4337 (which is currently not possible with Safe and the 4337 module).

Sednaoui:
> For some additional context, we are considering adding "signature validator" support to the 4337 module to allow modules to be executed over 4337 (which is currently not possible with Safe and the 4337 module).

Got it -- yes in that case the support for signature validators are quite beneficial. We can add support for it in the social recovery module if this is planned

remedcu:
Based on internal discussion, we have decided we will be exploring currently available module options instead of creating a new one from scratch. So, this PR will be considered as "won't implement".

PR #384, opened 2024-04-16
Passkey package: Add typechain types to the typescript build, fix types

mmv08:
I needed to import types for the example application type from the passkey package and I discovered that types were not included in the production build

This PR:
- Adds `typechain-types` folder to the build in typescript config
- Fixes the type errors that arose

Notes:
- The PR doesn't include other type improvements --> for example, there are many occurrences of `new ethers.Contract(abi, address, provider)` which would be of a high-level type `Contract`, but we have a type for the contract and could cast it to a more narrow type
- After the PR gets approved, I will create pump the package version number and push a beta release

### DIFF
--- a/modules/passkey/test/userstories/RotatePasskeyOwner.spec.ts
+++ b/modules/passkey/test/userstories/RotatePasskeyOwner.spec.ts
@@ -9,7 +9,7 @@ import {
   signSafeOp,
 } from '@safe-global/safe-4337/dist/src/utils/userOp'
 import { buildSignatureBytes } from '@safe-global/safe-4337/dist/src/utils/execution'
-import { chainId } from '@safe-global/safe-4337/dist/test/utils/encoding'
+import { chainId } from '../utils/hardhat'
 
 /**
  * User story: Rotate passkey owner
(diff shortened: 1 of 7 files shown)


nlordell (approved):
Can we also "pull" the package v0.1.0 from NPM (not sure what the technical term is, but mark it as "don't use"). I want to avoid any ambiguity about the code being "production ready and audite', which it is not.

modules/passkey/test/userstories/RotatePasskeyOwner.spec.ts
```diff
@@ -7,9 +7,9 @@ import {
   buildSafeUserOpTransaction,
   buildPackedUserOperationFromSafeUserOperation,
   signSafeOp,
-} from '@safe-global/safe-4337/src/utils/userOp'
-import { buildSignatureBytes } from '@safe-global/safe-4337/src/utils/execution'
-import { chainId } from '@safe-global/safe-4337/test/utils/encoding'
+} from '@safe-global/safe-4337/dist/src/utils/userOp'
+import { buildSignatureBytes } from '@safe-global/safe-4337/dist/src/utils/execution'
+import { chainId } from '@safe-global/safe-4337/dist/test/utils/encoding'
```
nlordell, line 12: I don't think this should be part of the user story code (I.e. we shouldn't depend on trivial test code).

Not related to this PR though.

mmv08: I can easily include it in this PR if you don't mind.

nlordell: Nope, don't mind as it is a very small change and somewhat related.

PR #388, opened 2024-04-17
Verify WebAuthn Signature Encoding

nlordell:
This PR adds logic to verify the WebAuthn signature encoding length. This is done so that encodings have a strict upper bound (determined by the 'standard' ABI encoding of the `WebAuthn.Signature` struct) on the length of the signatures.

This prevents a potential griefing attack where the signature that is sent to an ERC-4337 bundler could be arbitrarily padded with additional `0`s (which would have trivial increases to the `calldatasize` with onchain LZMA calldata decompression that already exists) while causing accounts to pay significantly more in gas costs for signature verification (bounded by the `verificationGasLimit`).

### DIFF
--- a/modules/passkey/contracts/libraries/WebAuthn.sol
+++ b/modules/passkey/contracts/libraries/WebAuthn.sol
@@ -68,24 +68,6 @@ library WebAuthn {
     AuthenticatorFlags internal constant USER_VERIFICATION = AuthenticatorFlags.wrap(0x04);
 
     /**
-     * @notice Casts calldata bytes to a WebAuthn signature data structure.
-     * @param signature The calldata bytes of the WebAuthn signature.
-     * @return data A pointer to the signature data in calldata.
-     * @dev This method casts the dynamic bytes array to a signature calldata pointer without
-     * additional verification. This is not a security issue for the WebAuthn implementation, as any
-     * signature data that would be represented from an invalid `signature` value, could also be
-     * encoded by a valid one. It does, however, mean that callers into the WebAuthn signature
-     * verification implementation might not validate as much of the data that they pass in as they
-     * would expect. With that in mind, callers should not rely on the encoding being verified.
-     */
-    function castSignature(bytes calldata signature) internal pure returns (Signature calldata data) {
-        // solhint-disable-next-line no-inline-assembly
-        assembly ("memory-safe") {
-            data := signature.offset
-        }
-    }
-
-    /**
      * @notice Encodes the client data JSON string from the specified challenge, and additional
      * client data fields.
      * @dev The client data JSON follows a very specific encoding process outlined in the Web
@@ -256,7 +238,11 @@ library WebAuthn {
         uint256 y,
         P256.Verifiers verifiers
     ) internal view returns (bool success) {
-        success = verifySignature(challenge, castSignature(signature), authenticatorFlags, x, y, verifiers);
+        Signature calldata signatureStruct;
+        (success, signatureStruct) = _castSignature(signature);
+        if (success) {
+            success = verifySignature(challenge, signatureStruct, authenticatorFlags, x, y, verifiers);
+        }
     }
 
     /**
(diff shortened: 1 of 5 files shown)


modules/passkey/contracts/libraries/WebAuthn.sol
```diff
@@ -70,18 +70,53 @@ library WebAuthn {
...
+        // Use of unchecked math as any overflows in dynamic length computations would cause
+        // out-of-gas reverts when computing the WebAuthn signing message.
+        unchecked {
+            // Allow for signature encodings where the dynamic bytes are aligned to 32-byte
+            // boundaries. This allows for high interoperability (as this is how Solidity and most
+            // tools `abi.encode`s the `Signature` struct) while setting a strict upper bound to how
+            // many additional padding bytes can be added to the signature, increasing gas costs.
+            uint256 alignmentMask = ~uint256(0x1f);
+            uint256 authenticatorDataAlignedLength = (authenticatorDataLength + 0x1f) & alignmentMask;
+            uint256 clientDataFieldsAlignedLength = (clientDataFieldsLength + 0x1f) & alignmentMask;
```
mmv08, line 105: I am not going to lie; I'm struggling to follow this 🙈 I tried to get [help from AI](https://www.perplexity.ai/search/Expain-this-code-5BdDJx.SQKiR0FVh87UNcQ#0) but I am still not getting it.

In particular, I don't get why adding `0x1f` is needed before doing the masking. Could you please explain?

nlordell: Ah, sorry. I will clarify in the comments. This is a combination of a few integer math tricks:

- Ceiling division can be implemented for integers like this: `let ceilDiv(a, b) = (a + (b-1)) / b`
- Multiplication and division by a number `x` which is a power of 2 can be done by logically shifting by `log2(x)`.

If we write padding more naturally, it would be like this:

```
let padded = ceilDiv(len, 32) * 32
```

Expanding this:

```
let padded = ((len + (32-1)) / 32) * 32
let padded = ((len + 31) >> log2(32)) << log2(32)
let padded = ((len + 0x1f) >> 5) << 5
```

Now, `(x >> y) << y` sets the least significant `y` bits to 0, which is the same as `x & 0b11...1100...00` with `y` trailing 0s, so continuing:

```
let padded = (len + 0x1f) & 0b11...1100000
let padded = (len + 0x1f) & ~0b00...0011111
let padded = (len + 0x1f) & ~0x1f
```

nlordell: Also, here is a stack overflow post about doing this in C: https://stackoverflow.com/questions/23927837/compute-size-of-object-if-it-has-to-be-aligned-as-a-specific-type

modules/passkey/contracts/test/TestWebAuthnLib.sol
```diff
@@ -4,6 +4,12 @@ pragma solidity ^0.8.0;
 import {P256, WebAuthn} from "../libraries/WebAuthn.sol";
 
 contract TestWebAuthnLib {
+    function castSignature(bytes calldata signature) external pure returns (WebAuthn.Signature calldata data) {
+        bool success;
+        (success, data) = WebAuthn.castSignature(signature);
+        require(success, "WebAuthnLib: invalid signature");
```
mmv08, line 10:
```suggestion
        require(success, "WebAuthnLib: invalid signature encoding");
```

Would that be more accurate?

modules/passkey/contracts/libraries/WebAuthn.sol
mmv08: Ahh, it was just required to get the size padded to a 32-byte constraint. I was not aware of the "aligned" terminology. Thank you very much for the explanation!

mmv08 (approved):

modules/passkey/contracts/libraries/WebAuthn.sol
mmv08: Leaving this for noobs like me https://www.perplexity.ai/search/What-does-it-4rF1dDWnTcm0wUoOzEpLsg#0

modules/passkey/contracts/test/TestWebAuthnLib.sol
nlordell: Yes, it would.

modules/passkey/contracts/libraries/WebAuthn.sol
nlordell: > aligned terminology

Ah yes, this is the typical term used in low level languages for memory allocations (i.e. to what multiple a memory allocation is padded to). For example: https://doc.rust-lang.org/std/alloc/struct.Layout.html

I know it from my low-level programming days.

modules/passkey/contracts/test/TestWebAuthnLib.sol
mmv08: Broke your PR by commiting the suggestion and not updating this in tests, fixed it

modules/passkey/contracts/libraries/WebAuthn.sol
nlordell: added short comment.

PR #396, opened 2024-04-22
Fix Precompile Address Width

nlordell:
When looking through the code, I realized that I made a mistake in #379. Specifically, we specify that precompiles fit into a `uint32`, when in fact they fit into a `uint16`:

https://eips.ethereum.org/EIPS/eip-1352

This PR changes the code to encode the `P256.Verifiers` type to be a `uint16 || address` instead of a slightly larger `uint32`.

Alternatively, we _could_ use a full 96 bits to represent the precompile address and make `P256.Verifiers` a `uint256` (this has a small positive effect on gas usage and code size, as it gets rid of compiler generated masking code).


remedcu (approved):

nlordell:
cc @akshay-ap and @mmv08 as this changes the function signature yet again 🙈. I can wait for you to merge your PRs so I can make sure that I don't break anything (I think there is a small needed adjustment in the proxy implementation for example).

akshay-ap:
@nlordell there are some merge conflicts.

PR #402, opened 2024-04-29
Emit signer created event

### LINKED ISSUE #397: WebAuthn Signer Creation Should Emit an Event
We should emit an event whenever a WebAuthn credential is created in order to be able to perform analytics for Passkey usage.

The event should look like:

```
event Created(address signer, uint256 x, uint256 y, P256.Verifiers verifiers);
```

Note that we should evaluate whether or not we want the signer address to be indexed or not. My intuition is "no".

akshay-ap:
Fixes #397

Changes in PR:
- [x] Emit signer created event in factory
- [x] Document
- [x] Test

The signer address is not indexed based on the justification below:

Using `indexed` keyword allows searching for events by filter (useful when searching events using libraries like web3.py, web3.js, etc.), Also, having indexed param would mean that `etherum.logs` table in Dune would have signer address available in `topic1` column. If not marked as indexed, then singer address need to be decoded from `data` column.

https://docs.soliditylang.org/en/latest/contracts.html#events

Indexing is accompanied by additional gas cost. So, indexed is recommended only if necessary. As we use dune and parsing data  column of `etherum.logs` would give signer address, I don't think indexing is needed.


modules/passkey/contracts/interfaces/ISafeSignerFactory.sol
```diff
@@ -10,6 +10,15 @@ import {P256} from "../libraries/P256.sol";
...
  */
 interface ISafeSignerFactory {
+    /**
+     * @notice Emitted when a new signer is created.
+     * @param signer The signer address.
+     * @param x The x-coordinate of the public key.
+     * @param y The y-coordinate of the public key.
+     * @param verifiers The P-256 verifiers to use.
+     */
+    event Created(address signer, uint256 x, uint256 y, P256.Verifiers verifiers);
```
nlordell, line 20: When we discussed in the sprint planning, what fields to index was still up for discussion. I would like to see some justification for the choice to not index the `signer` here.

nlordell (commented):
Code looks great. Missing reasoning/justification behind why we chose to index no fields.

modules/passkey/contracts/interfaces/ISafeSignerFactory.sol
akshay-ap: Here: https://github.com/safe-global/safe-modules/issues/397#issuecomment-2084695579

nlordell:
Please include in the PR description:

> Missing reasoning/justification behind why we chose to index no fields.

nlordell (approved):

PR #415, opened 2024-05-08
Document the Safe Signer Launchpad As Experimental

nlordell:
Fixes #409 

This PR moves the special 4337 integration contracts to an experimental folder and adds documentation specific to those contracts there (as well the rationale for the contracts being experimental in the first place).

To follow up this PR, I will add unit tests for how we intend passkeys to be used over ERC-4337 (i.e. the "non-experimental" way of using them).

PR #417, opened 2024-05-10
Add Additional ERC-4337 Related Passkey Tests

nlordell:
Follow up to #415

Now that we officially tagged the launchpad as "experimental", the module was missing unit and E2E tests demonstrating the "officially" supported way to use passkeys over ERC-4337. This PR adds both a unit test that a user operation with a passkey owner can be executed over the ERC-4337 reference entrypoint, as well as a local bundler test to ensure that no bundler rules are violated by the passkey signers.

### DIFF
--- a/modules/passkey/test/4337/experimental/Safe4337Module.spec.ts
+++ b/modules/passkey/test/4337/experimental/Safe4337Module.spec.ts
@@ -8,7 +8,6 @@ import {
   packGasParameters,
 } from '@safe-global/safe-4337/src/utils/userOp'
 import { chainId, encodeMultiSendTransactions } from '@safe-global/safe-4337/test/utils/encoding'
-import { Safe4337 } from '@safe-global/safe-4337/src/utils/safe'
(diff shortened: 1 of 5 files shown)


modules/passkey/test/4337/experimental/Safe4337Module.spec.ts
```diff
@@ -175,78 +174,6 @@ describe('Safe4337Module', () => {
         expect(await safeInstance.getOwners()).to.deep.equal([signerAddress])
       })
     })
-
-    describe('executeUserOp - existing account', () => {
```
nlordell, line 179: This test wasn't actually removed, just moved to `modules/passkey/test/4337/SafeWebAuthnSigner.spec.ts` as it is unrelated to the launchpad and was in the wrong place to begin with.

Conceptually, there are no "executeUserOp - existing account" tests for the launchpad, as it is only ever used for account creation.

mmv08 (approved):

PR #412, opened 2024-05-06
Use Account Storage for Shared WebAuthn Signer

nlordell:
Fixes #408 

This PR reworks the WebAuthn singleton signer to use the account storage instead of its own storage for signer configuration (x, y, and verifiers configuration values). This allows the shared signer to work **without** a staked factory :tada:!

We also rename the methods a bit to be more consistent with our signer factory implementation, and adjust the E2E test accordingly (noting that a staked factory is no longer needed).

This also required some small adjustments to the `4337-passkeys-singleton-signer` example for the new code (but overall simplification - no need for a staked factory anymore).

PR #413, opened 2024-05-06
Promote Shared Singleton Signer Implementation

nlordell:
This PR promotes the shared singleton signer implementation to the 4337 directory instead of being a "test" contract. This involves:
- Adding it to the deployments
- Moving and renaming the contract to remove the `Test` prefix

I also took this opportunity to rename it to a "shared" signer instead of "singleton" signer, so that it does not get confused with the other singleton contract (the `SafeWebAuthnSignerSingleton`) that already exists in this repository.

Additionally, I moved some of the test files around in order to clearly differentiate between E2E tests and unit tests in the `test/4337` directory.

Unit tests for 100% coverage of the `SafeWebAuthnSharedSigner` will be added in a follow up PR.

PR #419, opened 2024-05-13
Use Account Storage for Shared WebAuthn Signer

### LINKED ISSUE #408: Use Self-Storage for Passkey Singleton Signer
Instead of using storage in the `TestSingletonSigner` contract, delegate call from a Safe in order for the storage to be in the account itself. This allows the signer to work *without* a staked factory.

nlordell:
Fixes #408 

This PR reworks the WebAuthn singleton signer to use the account storage instead of its own storage for signer configuration (x, y, and verifiers configuration values). This allows the shared signer to work **without** a staked factory :tada:!

We also rename the methods a bit to be more consistent with our signer factory implementation, and adjust the E2E test accordingly (noting that a staked factory is no longer needed).

This also required some small adjustments to the `4337-passkeys-singleton-signer` example for the new code (but overall simplification - no need for a staked factory anymore).

---

This is a duplicate of #412, which was accidentally merged into the wrong branch :see_no_evil:


mmv08 (approved):

nlordell:
Merged with #413

PR #420, opened 2024-05-14
Pnpm experiment

### LINKED ISSUE #349: Evaluate Other Node Package Managers
Currently, we use NPM workspaces to some success. However there are some rough edges, for example:
- Workspace package dependencies don't correctly link `bin` when they are generated in the package's build step (workaround: https://github.com/safe-global/safe-modules/blob/cc9c764f72a1ab4580c19ace0a78461172a3c511/package.json#L14)
- Workspace packages don't have deterministic build order, which causes issues (workaround: https://github.com/safe-global/safe-modules/blob/cc9c764f72a1ab4580c19ace0a78461172a3c511/modules/passkey/package.json#L45)
- Issues with platform specific dependencies

All in all, it would be nice to evaluate alternatives and see if there are any advantages to switching to a different package management system.

mmv08:
This PR:
- Fixes #349 
- Upgrades the certora-cli version used on Github actions to `7.3.0` 

PNPM is more strict about transitive dependencies and sometimes strange things happen with module resolution unless the dependency is explicitly stated in `package.json` and imported. I added comments to highlight such cases.

@nlordell please test this change on linux :)


modules/4337/package.json
```diff
@@ -9,29 +9,27 @@
   "files": [
     "contracts",
     "dist",
-    "src",
-    "test",
```
mmv08, line 13: Somehow, I ran into an issue where it tried to build the source files inside the 4337 dependency. I don't think we need to publish source and test files to npm, so I removed them.

modules/4337/package.json
```diff
@@ -51,32 +49,29 @@
...
+    "solhint": "^5.0.1",
     "ts-node": "^10.9.2",
     "typescript": "^5.4.5",
     "yargs": "^17.7.2"
   },
-  "overrides": {
-    "@safe-global/safe-contracts": {
-      "ethers": "^6.12.0"
-    }
-  },
```
mmv08, line 79: doesn't seem to be needed anymore

modules/passkey/hardhat.config.ts
```diff
@@ -1,4 +1,5 @@
 import '@nomicfoundation/hardhat-toolbox'
+import '@nomicfoundation/hardhat-ethers'
```
mmv08, line 2: Somehow not adding this package explicitly to the `package.json` and importing here results in a type mess for `ethers.provider`

modules/passkey/package.json
```diff
@@ -37,34 +36,38 @@
...
+    "solhint": "^5.0.1"
   },
   "overrides": {
     "@safe-global/safe-contracts": {
       "ethers": "^6.11.1"
     }
   },
   "dependencies": {
+    "@openzeppelin/contracts": "^5.0.0",
+    "@account-abstraction/contracts": "^0.7.0",
```
mmv08, line 70: we were importing these dependencies but they were not installed

packages/4337-local-bundler/hardhat.config.ts
```diff
@@ -0,0 +1,8 @@
+import "@nomicfoundation/hardhat-toolbox"
```
mmv08, line 1: I don't know tsc worked without this

.github/workflows/certora_4337.yml
```diff
@@ -23,12 +23,17 @@ jobs:
...
+        with:
+          version: 9
+
       - name: Install python
         uses: actions/setup-python@v4
         with: { python-version: 3.11 }
 
       - name: Install certora cli
-        run: pip install -Iv certora-cli==6.1.3
+        run: pip install -Iv certora-cli==7.3.0
```
nlordell, line 36: Context: Unrelated but opportunistic change (was tested with this during development, and there are no additional Certora related changes because of the version bump).

.github/workflows/ci_4337_upstream_bundler.yml
```diff
@@ -15,12 +15,17 @@ jobs:
...
         with:
           node-version: 20.x
-          cache: npm
-          cache-dependency-path: package-lock.json
+          cache: pnpm
       - run: |
-          npm ci
+          pnpm install
           npm run test:4337:upstream -w modules/4337
           npm run test:4337:upstream -w modules/passkey
```
nlordell, line 31:
```suggestion
          pnpm run --filter "@safe-global/safe-4337" test:4337:upstream
          pnpm run --filter "@safe-global/safe-passkey" test:4337:upstream
```

.github/workflows/ci_passkey_example.yml
```diff
@@ -11,13 +11,16 @@ jobs:
...
         with:
           node-version: 20.x
-          cache: npm
-          cache-dependency-path: package-lock.json
+          cache: pnpm
       - run: |
-          npm ci
+          pnpm install
           npm run lint -w examples/4337-passkeys
           npm run build -w examples/4337-passkeys
```
nlordell, line 25: One more `pnpm -> npm` change.

.prettierignore
```diff
@@ -1,3 +1,5 @@
+# They're excluded on the global level because each of them defines its own linting and formatting rules
 examples/**
 modules/**
 packages/**
+pnpm-lock.yaml
```
nlordell, line 5: Nit: missing newline
```suggestion
pnpm-lock.yaml
```

examples/4337-passkeys-singleton-signer/README.md
```diff
@@ -33,7 +33,7 @@ Helpful links:
 ### Run the app in development mode
 
 ```bash
-npm run dev -w examples/4337-passkeys-singleton-signer
+pnpm run dev -w examples/4337-passkeys-singleton-signer
```
nlordell, line 36: Should use `--filter` and not `-w`

examples/4337-passkeys/README.md
```diff
@@ -33,7 +33,7 @@ Helpful links:
 ### Run the app in development mode
 
 ```bash
-npm run dev -w examples/4337-passkeys
+pnpm run dev -w examples/4337-passkeys
```
nlordell, line 36: Should use `--filter` instead fo `-w`

modules/4337/package.json
nlordell: I would argue that including `src` in the NPM package makes sense (it allows you to import TypeScript directly instead of the built JS artifacts).

modules/passkey/package.json
```diff
@@ -37,34 +36,38 @@
...
     "hardhat": "^2.22.3",
     "hardhat-deploy": "^0.12.4",
-    "solhint": "^4.5.4"
+    "solhint": "^5.0.1"
   },
   "overrides": {
     "@safe-global/safe-contracts": {
       "ethers": "^6.11.1"
     }
   },
```
nlordell, line 67: Is this still needed?

modules/passkey/package.json
nlordell: Why aren't these `devDependencies`? I don't think that contracts outside of `test` need them to compile.

modules/passkey/package.json
```diff
@@ -37,34 +36,33 @@
...
-  "overrides": {
-    "@safe-global/safe-contracts": {
-      "ethers": "^6.11.1"
-    }
+    "solhint": "^5.0.1"
   },
   "dependencies": {
+    "@openzeppelin/contracts": "^5.0.0",
+    "@account-abstraction/contracts": "^0.7.0",
     "@safe-global/safe-contracts": "^1.4.1-build.0",
```
nlordell, line 66: Hmm, why is this not a `devDependency` as well?

packages/4337-local-bundler/hardhat.config.ts
nlordell: Maybe it doesn't use some of the extensions that were causing problems in the other packages?

modules/passkey/package.json
mmv08: It is used in the launchpad contracts and required if someone will use the package and build the contracts from source.

modules/passkey/package.json
mmv08: Answered here https://github.com/safe-global/safe-modules/pull/420/files#r1601774914

nlordell (approved):

nlordell:
Created https://github.com/safe-global/safe-modules/pull/422 to make sure CI runs without issues.

remedcu:
Should we have a small section in the main `README.md` which says how to install the package for all the modules & examples?

mmv08:
> Should we have a small section in the main `README.md` which says how to install the package for all the modules & examples?

In my opinion, it's not rocket science and more than one way to install it exists - they all can be found on the pnpm website. I don't see value in maintaining pnpm installation docs for us

PR #422, opened 2024-05-16
Draft: PNPM Changes In Single Commit [DO NOT MERGE OR REVIEW]

nlordell:
It looks like GitHub CI "files changed" rules only check the *last* commit and not all commits. This means we may get false positive CI passes for PRs that have many changes where the last one only runs a small subset of CI.

This PR just just here to test the changes as a single PR, thus ensuring that **all** relevant CI is executed.


nlordell:
Yay! CI passes.

PR #423, opened 2024-05-17
Local Bundler Test with New Shared Signer Flow

nlordell:
This PR demonstrates that the launchpad is, in fact, kind of useless.

Specifically, a new local bundler end-to-end test is added where the shared signer is used to create a Safe that ends up with the same configuration as with the signer launchpad. It uses the shared signer as the initial owner which is used for signing **only** the first user operation. During execution of the first user operation, the Safe's owner is swapped for a freshly deployed `SafeWebAuthnSignerProxy` instead of the shared signer. This means that we end up with a Safe with a "regular" passkey owner without any of the downsides (complicated launchpad code, rigid ownership structure, only one passkey owner allowed, etc.).

If this PR is approved, I will **remove** the `SafeSignerLaunchpad` in follow up PRs as well as update the passkey example to use the same strategy so it ends up with a standard passkey owner.

### DIFF
--- a/modules/passkey/src/deploy/launchpad.ts
+++ b/modules/passkey/src/deploy/launchpad.ts
@@ -14,12 +14,6 @@ const deploy: DeployFunction = async ({ deployments, getNamedAccounts }) => {
     log: true,
     deterministicDeployment: true,
   })
-  await deploy('SafeWebAuthnSharedSigner', {
-    from: deployer,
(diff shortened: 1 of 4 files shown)


mmv08 (approved):

remedcu (approved):
Small nit, rest LGTM 👍🏾

PR #429, opened 2024-06-03
Shared Signer Mapping

nlordell:
This PR changes the behaviour of the shared signer to store its configuration as if it were behind a mapping where the key is the shared signer address. This allows multiple shared signer instances to be configured for a single account (in case you ever need to deploy a 2/2 Safe owned by passkeys with 4337).

I created an "accessor" contract to prove that the storage location is indeed the equivalent to a `mapping(address self => Signer)`.

### DIFF
--- a/modules/passkey/test/4337/experimental/SafeWebAuthnSharedSigner.spec.ts
+++ b/modules/passkey/test/4337/experimental/SafeWebAuthnSharedSigner.spec.ts
@@ -5,32 +5,41 @@ import * as ERC1271 from '../../utils/erc1271'
 import { DUMMY_AUTHENTICATOR_DATA, base64UrlEncode, getSignatureBytes } from '../../../src/utils/webauthn'
 import { encodeWebAuthnSigningMessage } from '../../utils/webauthnShim'
 
-const SIGNER_STORAGE_SLOT = ethers.toBeHex(BigInt(ethers.id('SafeWebAuthnSharedSigner.signer')) - 3n, 32)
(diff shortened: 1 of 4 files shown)


nlordell:
**Do not merge until the Certora audit is complete**.

modules/passkey/test/4337/experimental/SafeWebAuthnSharedSigner.spec.ts
```diff
@@ -94,11 +103,47 @@ describe('SafeWebAuthnSharedSigner', () => {
...
+        x: ethers.id('publicKey.x'),
+        y: ethers.id('publicKey.y'),
+        verifiers: ethers.toBeHex(await mockVerifier.getAddress(), 32),
+      }
+
+      const initializer = safeSingleton.interface.encodeFunctionData('setup', [
+        [sharedSigner.target],
+        1,
+        sharedSigner.target,
+        sharedSigner.interface.encodeFunctionData('configure', [config]),
```
remedcu, line 120: Doubt: For a 2/2 owner setup, won't it require something like a dispatcher (or some other better name) to call all shared signers when multiple instances are involved? Or am I getting it wrong?

Something like:
```
        [sharedSignerA.target, sharedSignerB.target],
        2,
        sharedSignerDispatcher.target,
        sharedSignerDispatcher.interface.encodeFunctionData('configureSharedSigners', [configSharedSigners]),
```

remedcu: Ah, that makes perfect sense. Thanks!

remedcu (approved):

PR #430, opened 2024-06-03
Add Test With Precompile

### LINKED ISSUE #317: E2E Test With P-256 Precompile
Follow up to #285, once running an execution client that supports the P-256 precompile is easier, an E2E test that the WebAuthn signer supports the precompile contract should be added. We expect to use `bor` when the dev mode issue is resolved. 

### Expected Outcome

And E2E test that spins up a local dev node with the P-256 precompile and a test that verifies that WebAuthn signature verification works as expected.

nlordell:
Fixes #317 

This PR adds a new `docker-compose` file for running a local Geth node (based on the Optimism Geth node) with RIP-7212 precompile support in order verify that the Safe WebAuthn verifier actually works with the [ER]IP-7212 precompile.

In particular, we just use the `@bench` tests which are already run in CI and would give us a rough idea on how much gas a Passkey signature verification will take with the precompile (spoiler - its just 15k gas, which is reasonable).

I did not create a dedicated E2E test, as I felt it didn't have much value.


modules/passkey/docker-compose.yaml
```diff
@@ -0,0 +1,18 @@
+version: '3.8'
```
remedcu, line 1: Could be removed from [docker version 25.05](https://docs.docker.com/compose/compose-file/04-version-and-name/#version-top-level-element-optional)

remedcu (approved):

mmv08 (approved):

modules/passkey/docker-compose.yaml
nlordell: https://github.com/safe-global/safe-modules/pull/431

PR #432, opened 2024-06-05
Add E2E Test For Multiple Passkey Owners With 4337

### LINKED ISSUE #372: `SafeECDSASignerLaunchpad` has unclear support for multi-owner accounts

nlordell:
Fixes #372

This PR adds a new E2E test that demonstrates the use of a Safe being deployed over 4337 with multiple passkey owners. This means that we can have arbitrary ownership structures with passkeys and 4337 :tada:.

There is an important caveat however, the tracing mechanism used by the bundler is timing out when verifying multiple passkey signatures, so this E2E test is only half complete. At least we verify that the signature verification logic would work, and eventually when precompiles become part of Geth, we will be able to switch the E2E test to use that instead which should greatly speed up tracing.

I still think it is worth merging this PR as is, as it also checks that multiple shared signer instances can be used in parallel, a new feature added in #429.

### DIFF
--- a/modules/passkey/test/4337/local-bundler/experimental/SafeWebAuthnSharedSigner.spec.ts
+++ b/modules/passkey/test/4337/local-bundler/experimental/SafeWebAuthnSharedSigner.spec.ts
@@ -298,6 +298,7 @@ describe('Safe WebAuthn Shared Signer [@4337]', () => {
     await user.sendTransaction({ to: safeAddress, value: ethers.parseEther('0.5') }).then((tx) => tx.wait())
 
     expect(ethers.dataLength(await ethers.provider.getCode(safeAddress))).to.equal(0)
+    expect(ethers.dataLength(await ethers.provider.getCode(signerAddress))).to.equal(0)
     expect(await sharedSigner.getConfiguration(safeAddress)).to.deep.equal([0n, 0n, 0n])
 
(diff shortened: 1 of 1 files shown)


mmv08 (approved):

nlordell:
Blocked on merge pending Certora audit (don't want to change any contracts).

PR #435, opened 2024-06-06
Fix `runOp` Script

nlordell:
I was running the 4337 `runOp.ts` script this morning and noticed it was subtly broken. This PR fixes it.

I also refactored the scripts a bit - specifically we no longer `--fix` on `lint:ts`. This was because it would cause CI to silently pass in cases where the linter could autofix the code (when we want it to fail instead).


nlordell:
I use it to send a user operation to an actual chain (Sepolia in my case). It is still convenient IMO.

mmv08 (approved):

Issue #426, opened 2024-05-28
Release a new `4337` module version with updated JS tooling.


mmv08:
## Context / issue

I needed to deploy & verify the 4337 contract to a new network, so I checked out to `[4337/v0.3.0](https://github.com/safe-global/safe-modules/tree/4337/v0.3.0)` tag.

And ran into an issue with installing dependencies:
```
~/projects/safe/safe-modules #4337/v0.3.0@5931a275 ❯ npm i                                                                                                                                               22s 17:22:01
npm error code 1
npm error path /Users/mmv/projects/safe/safe-modules/modules/4337
npm error command failed
npm error command sh -c npm run build
npm error > @safe-global/safe-4337@0.3.0 build
npm error > npm run build:sol && npm run build:ts
npm error
npm error
npm error > @safe-global/safe-4337@0.3.0 build:sol
npm error > rimraf build typechain-types && hardhat compile
npm error Error HH18: You installed Hardhat with a corrupted lockfile due to the NPM bug #4828.
npm error
npm error Please delete your node_modules, package-lock.json, reinstall your project, and try again.
npm error For more info go to https://hardhat.org/HH18 or run Hardhat with --show-stack-traces
npm error npm error Lifecycle script `build:sol` failed with error:
npm error npm error Error: command failed
npm error npm error   in workspace: @safe-global/safe-4337@0.3.0
npm error npm error   at location: /Users/mmv/projects/safe/safe-modules/modules/4337
npm error npm error Lifecycle script `build` failed with error:
npm error npm error Error: command failed
npm error npm error   in workspace: @safe-global/safe-4337@0.3.0
npm error npm error   at location: /Users/mmv/projects/safe/safe-modules/modules/4337

npm error A complete log of this run can be found in: /Users/mmv/.npm/_logs/2024-05-27T15_22_27_743Z-debug-0.log
```

In the end, I had to use the current main branch, which contains pnpm updates (luckily, we didn't do any code changes to the 4337 contract)

## Proposed solution

Release a new version with updated js tooling and a new tag (e.g., 4337/v0.3.0-1 per @nlordell suggestion)

akshay-ap:
My understanding is that as the script has a bug fix, new version should `v0.3.1`. Curious to know why there is `v0.3.0-1` as new version instead? I couldn't find it as a convention in npm docs.

mmv08:
[...]

PR #445, opened 2024-06-19
Feature: 4337 module `v0.3.0-1`

mmv08:
This PR:
- Implements https://github.com/safe-global/safe-modules/issues/426 by releasing the code we have now under a new tag
- There were no contract changes, so the expected addresses are the same. I confirmed it by trying to deploy to a network where the contracts are already deployed:
```
> hardhat deploy-contracts --network "mainnet"

Nothing to compile
No need to generate any newer typings.
reusing "SafeModuleSetup" at 0x2dd68b007B46fBe91B9A7c3EDa5A7a1063cB5b47
reusing "Safe4337Module" at 0x75cf11467937ce3F2f357CE24ffc3DBF8fD5c226
```

### DIFF
--- a/modules/4337/package.json
+++ b/modules/4337/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@safe-global/safe-4337",
-  "version": "0.3.0",
+  "version": "0.3.0-1",
   "description": "Safe Module for ERC-4337 support",
   "homepage": "https://github.com/safe-global/safe-modules/tree/main/modules/4337",
   "license": "GPL-3.0",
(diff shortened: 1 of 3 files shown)


akshay-ap:
My understanding is that as the script has a bug fix, new version would be v0.3.1. Curious to know why there is v0.3.0-1 as new version instead? I couldn't find it as a convention in npm docs.

akshay-ap (approved):
Approved with comment.

mmv08:
> My understanding is that as the script has a bug fix, new version would be v0.3.1. Curious to know why there is v0.3.0-1 as new version instead? I couldn't find it as a convention in npm docs.

There is no bug in the script or the contracts per se. The bug is in the surrounding tooling. We had problems with NPM where it would generate a platform-specific lockfile and refuse to regenerate it on other platforms. This led the `v0.3.0` to be only usable (to deploy or verify contracts) on the platform where the lockfile was generated, which happens to be Linux.

The version number suggestion came from Nick [here](https://5afe.slack.com/archives/C04G9CPMR3J/p1716879019086429?thread_ts=1716824302.324979&cid=C04G9CPMR3J)

PR #448, opened 2024-06-21
Optimize Compiler Settings

nlordell:
I noticed that the IR optimizer is better for all contracts **except** the FCL library. This PR optimizes the compiler configuration to take advantage of the IR optimizer, while adding an exception for the FCL P-256 verifier contract so that it does not suffer any regressions.

```
=== BEFORE ===
  Gas Benchmarking [@bench]
    SafeWebAuthnSignerProxy
      ⛽ deployment: 94366
      ✔ Benchmark signer deployment cost (476ms)
      ⛽ verification (FreshCryptoLib): 216907
      ✔ Benchmark signer verification cost with FreshCryptoLib verifier (116ms)
      ⛽ verification (Daimo): 346809
      ✔ Benchmark signer verification cost with Daimo verifier (103ms)
      ⛽ verification (Dummy): 14309
      ✔ Benchmark signer verification cost with Dummy verifier (100ms)
      ⛽ verification (Precompile): 15098
      ✔ Benchmark signer verification cost with Precompile verifier (104ms)

=== AFTER  ===
  Gas Benchmarking [@bench]
    SafeWebAuthnSignerProxy
      ⛽ deployment: 91042
      ✔ Benchmark signer deployment cost (457ms)
      ⛽ verification (FreshCryptoLib): 213919
      ✔ Benchmark signer verification cost with FreshCryptoLib verifier (121ms)
      ⛽ verification (Daimo): 342642
      ✔ Benchmark signer verification cost with Daimo verifier (110ms)
      ⛽ verification (Dummy): 12564
      ✔ Benchmark signer verification cost with Dummy verifier (104ms)
      ⛽ verification (Precompile): 13379
      ✔ Benchmark signer verification cost with Precompile verifier (98ms)
```

Furthermore, we bump the compiler version to v0.8.26 as it is the lastest stable release and brings some small gas improvements.

### DIFF
--- a/modules/passkey/hardhat.config.ts
+++ b/modules/passkey/hardhat.config.ts
@@ -31,6 +31,18 @@ const customNetwork = CUSTOM_NODE_URL
     }
   : {}
 
+const compilerSettings = {
+  version: '0.8.24',
+  settings: {
+    optimizer: {
+      enabled: true,
+      runs: 10_000_000,
+    },
+    viaIR: true,
+    evmVersion: 'paris',
+  },
+}
+
 const config: HardhatUserConfig = {
   paths: {
     artifacts: 'build/artifacts',
@@ -54,19 +66,19 @@ const config: HardhatUserConfig = {
     ...customNetwork,
   },
   solidity: {
-    compilers: [
-      {
-        version: '0.8.24',
+    compilers: [compilerSettings],
+    overrides: {
+      // FCL library does not optimize well via IR. In order to take advantage of the IR optimizer
+      // in the rest of the project without causing significant regressions to the FCL verifier, we
+      // add a compiler setting override for that specific contract.
+      'contracts/verifiers/FCLP256Verifier.sol': {
+        ...compilerSettings,
         settings: {
-          optimizer: {
-            enabled: true,
-            runs: 10_000_000,
-          },
+          ...compilerSettings.settings,
           viaIR: false,
-          evmVersion: 'paris',
         },
       },
-    ],
+    },
   },
   namedAccounts: {
     deployer: 0,
--- a/modules/passkey/hardhat.config.ts
+++ b/modules/passkey/hardhat.config.ts
@@ -32,7 +32,7 @@ const customNetwork = CUSTOM_NODE_URL
   : {}
 
 const compilerSettings = {
-  version: '0.8.24',
+  version: '0.8.26',
   settings: {
     optimizer: {
       enabled: true,
--- a/modules/passkey/contracts/4337/SafeWebAuthnSharedSigner.sol
+++ b/modules/passkey/contracts/4337/SafeWebAuthnSharedSigner.sol
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: LGPL-3.0-only
-pragma solidity >=0.8.0;
+pragma solidity 0.8.26;
 
 import {SignatureValidator} from "../base/SignatureValidator.sol";
 import {ISafe} from "../interfaces/ISafe.sol";
--- a/modules/passkey/contracts/SafeWebAuthnSignerFactory.sol
+++ b/modules/passkey/contracts/SafeWebAuthnSignerFactory.sol
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: LGPL-3.0-only
-pragma solidity >=0.8.0;
+pragma solidity 0.8.26;
 
 import {ISafeSignerFactory} from "./interfaces/ISafeSignerFactory.sol";
 import {SafeWebAuthnSignerProxy} from "./SafeWebAuthnSignerProxy.sol";
--- a/modules/passkey/contracts/verifiers/FCLP256Verifier.sol
+++ b/modules/passkey/contracts/verifiers/FCLP256Verifier.sol
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: LGPL-3.0-only
 /* solhint-disable no-complex-fallback */
 /* solhint-disable payable-fallback */
-pragma solidity 0.8.24;
+pragma solidity 0.8.26;
 
 import {IP256Verifier} from "../interfaces/IP256Verifier.sol";
 import {FCL_ecdsa} from "../vendor/FCL/FCL_ecdsa.sol";


modules/passkey/hardhat.config.ts
```diff
@@ -31,6 +31,18 @@ const customNetwork = CUSTOM_NODE_URL
     }
   : {}
 
+const compilerSettings = {
+  version: '0.8.24',
```
mmv08, line 35: if we're changing the compiler settings, what about using the latest version?

mmv08 (approved):

modules/passkey/hardhat.config.ts
```diff
@@ -31,6 +31,18 @@ const customNetwork = CUSTOM_NODE_URL
...
+  settings: {
+    optimizer: {
+      enabled: true,
+      runs: 10_000_000,
+    },
+    viaIR: true,
+    evmVersion: 'paris',
```
mmv08, line 42: paris is pre-push0, right? maybe we don't need this either?

modules/passkey/hardhat.config.ts
```diff
@@ -54,19 +66,19 @@ const config: HardhatUserConfig = {
...
-    compilers: [
-      {
-        version: '0.8.24',
+    compilers: [compilerSettings],
+    overrides: {
+      // FCL library does not optimize well via IR. In order to take advantage of the IR optimizer
+      // in the rest of the project without causing significant regressions to the FCL verifier, we
+      // add a compiler setting override for that specific contract.
+      'contracts/verifiers/FCLP256Verifier.sol': {
+        ...compilerSettings,
```
remedcu, line 75:
```suggestion
        version: compilerSettings.version,
```

nlordell: I did this on purpose to future proof in case new fields were added. Agree it isn't strictly necessary here, but wanted to future proof just in case. Also it feels to convey intent a bit more IMO: we want exactly `compilerSettings` except for this single value.

modules/passkey/hardhat.config.ts
nlordell: We do, so we are BNB compatible.

Also, note that `hardhat` currently defaults to `paris` anyway (at least it did when I experimented with switching this setting off), so this is more to explicitly document the EVM version that is being targetted.

modules/passkey/hardhat.config.ts
nlordell: My rationale for not doing it is as such:
1. from the changelog there are no significant features or fixes that were added since 0.8.24: [source](https://github.com/ethereum/solidity/blob/develop/Changelog.md)
2. Currently, the `FCL256Verifier` is byte-for-byte the same (and won't change with any of the planned changes :tada:, none of the audit competition changes affect it). Updating the compiler version would change that
3. The audit was done for compiler v0.8.24, I know we are changing the compiler settings, but it feels less extreme than a new compiler version.

I don't feel strongly either way, I did consider it when making this PR but opted to err on the side of caution.

mmv08: I ran the benchmarks with `0.8.26` and it seems it slightly improves runtime gas costs:
```
  Gas Benchmarking [@bench]
    SafeWebAuthnSignerProxy
      ⛽ deployment: 91042
      ✔ Benchmark signer deployment cost (568ms)
      ⛽ verification (FreshCryptoLib): 215607
      ✔ Benchmark signer verification cost with FreshCryptoLib verifier (107ms)
      ⛽ verification (Daimo): 345542
      ✔ Benchmark signer verification cost with Daimo verifier (100ms)
      ⛽ verification (Dummy): 12564
      ✔ Benchmark signer verification cost with Dummy verifier (107ms)
      ⛽ verification (Precompile): 13379
      ✔ Benchmark signer verification cost with Precompile verifier (117ms)
```

mmv08: With `0.8.25` they're almost identical except the Daimo verifier:
```
  Gas Benchmarking [@bench]
    SafeWebAuthnSignerProxy
      ⛽ deployment: 90399
      ✔ Benchmark signer deployment cost (508ms)
      ⛽ verification (FreshCryptoLib): 213844
      ✔ Benchmark signer verification cost with FreshCryptoLib verifier (145ms)
      ⛽ verification (Daimo): 342496
      ✔ Benchmark signer verification cost with Daimo verifier (121ms)
      ⛽ verification (Dummy): 12911
      ✔ Benchmark signer verification cost with Dummy verifier (110ms)
      ⛽ verification (Precompile): 13726
      ✔ Benchmark signer verification cost with Precompile verifier (111ms)
```

I guess those smile improvements (from the 0.8.26 changelog) made a slight difference, not a compiler engineer, though:
```
Compiler Features:
Yul IR Code Generation: Cheaper code for reverting with errors of a small static encoding size.
Yul Optimizer: New, faster default optimizer step sequence.
```

nlordell: Note that the code got slightly bigger from v0.8.25 -> v0.8.26

```
=== v0.8.25 ===
  Gas Benchmarking [@bench]
    SafeWebAuthnSignerProxy
      ⛽ deployment: 90399
      ✔ Benchmark signer deployment cost (508ms)

=== v0.8.26 ===
  Gas Benchmarking [@bench]
    SafeWebAuthnSignerProxy
      ⛽ deployment: 91042
      ✔ Benchmark signer deployment cost (568ms)
```

Overall, I think you are right, we can bump the compiler version here. My main argument was to reuse the FCLP256Verifier, but that is moot apparently :disappointed:

nlordell: > except the Daimo verifier:

This is slightly random because the verifiers have different gas characteristics depending on what public key/signature is used (which is random in the tests).

I would look specifically at `deployment`, `dummy` and `precompile` gas metrics only (unless you collect data from many runs and compute averages for FCL and Daimo).

Also, Daimo verifier is **not** compiled (we vendor bytecode directly), so compiler settings should not affect it at all.

nlordell: IMplemented suggestion in a6cd15a0c0c3e8b93a18e5ffa9975824c23c66ea

modules/passkey/hardhat.config.ts
mmv08: it seems like BNB already has push0? https://forum.bnbchain.org/t/bnb-chain-roadmap-mainnet/936

nlordell: This is super recent! It wasn't the case 4 days ago: https://github.com/hats-finance/Safe-0x2909fdefd24a1ced675cb1444918fa766d76bdac/issues/19#issuecomment-2180019350

Then `s/BNB/Linea/g`: I think the point remains the same that `PUSH0` isn't generally well supported across L2s:
```
$curl -s https://rpc.linea.build -H 'content-type: application/json' --data '{"jsonrpc":"2.0","id":0,"method":"et
h_call","params":[{"data":"0x5f"},"latest"]}'
{"jsonrpc":"2.0","id":0,"error":{"code":-32000,"message":"invalid opcode: PUSH0"}}
```

mmv08: interesting, the forum post says it was implemented in `Kepler: 2024/01/23` hardfork

akshay-ap (approved):

remedcu (approved):

PR #452, opened 2024-06-30
Add a `sherif` monorepo linter to the CI

mmv08:
The PR introduces the 'sherif' linter, which is an opinionated, zero-config linter designed specifically for JavaScript monorepos. This addition aims to improve code quality and standardize development practices across the safe-modules project.

**CI Integration:**
The PR adds 'sherif' to the CI workflow. This ensures that the linter runs automatically on pull requests, helping to catch potential issues early in the development process.
**Linter Features:**

'Sherif' offers several features:
- Zero-configuration setup, making it easy to integrate and use
- Fast performance, as it doesn't require node_modules to be installed
- Compatibility with various package managers (PNPM, NPM, Yarn)
- Ability to prevent regressions by enforcing consistent dependency versions across the monorepo

**Found and fixed issues:**
```
2 issues found in ./package.json:

 ⨯ error The root package.json should be private to prevent accidentaly publishing it to a registry. root-package-private-field
  │ {
  +   "private": "true"   ← missing private field.
  │ }

 ⨯ error The root package.json should specify the package manager and version to use. Useful for tools like corepack. root-package-manager-field
  │ {
  +   "packageManager": "..."   ← missing packageManager field.
  │ }

1 issue found in ./packages/4337-provider/package.json:

 ⨯ error Private packages shouldn't have @types/* in dependencies. types-in-dependencies
  │ {
  │   "private": "true",     ← package is private...
  │   ...
  -   "dependencies": {      ← but has @types/* in dependencies...
  -      "@types/node": "...",
  -   },
  │   ...
  +   "devDependencies": {   ← instead of devDependencies.
  +      "@types/node": "...",
  +   }
  │ }

13 issues found in ./:

 ⨯ error Dependency @account-abstraction/contracts has multiple versions defined in the workspace. multiple-dependency-versions
  ./examples
      4337-passkeys             0.7.0   ↓ lowest
  ./modules
      4337                      ^0.7.0   ↑ highest
      passkey                   ^0.7.0   ↑ highest
  ./packages
      4337-local-bundler        ^0.7.0   ↑ highest

 ⨯ error Dependency @safe-global/safe-contracts has multiple versions defined in the workspace. multiple-dependency-versions
  ./examples
      4337-passkeys             ^1.4.1-build.0   ↑ highest
  ./modules
      4337                      ^1.4.1-build.0   ↑ highest
      passkey                   ^1.4.1-build.0   ↑ highest
      recovery                  =1.4.1-build.0   ∼ between
  ./packages
      4337-local-bundler        ^1.4.1-build.0   ↑ highest

 ⨯ error Dependency ethers has multiple versions defined in the workspace. multiple-dependency-versions
  ./examples
      4337-gas-metering         ^6.12.1   ↓ lowest
  ./modules
      allowances                ^6.12.1   ↓ lowest
      passkey                   ^6.12.1   ↓ lowest
      recovery                  ^6.12.1   ↓ lowest
  ./packages
      4337-local-bundler        ^6.12.1   ↓ lowest
      4337-provider             ^6.12.1   ↓ lowest
  ./examples
      4337-passkeys             ^6.13.1   ↑ highest
  ./modules
      4337                      ^6.13.1   ↑ highest

 ⨯ error Dependency hardhat has multiple versions defined in the workspace. multiple-dependency-versions
  ./modules
      allowances                ^2.22.3   ↓ lowest
      passkey                   ^2.22.3   ↓ lowest
      recovery                  ^2.22.3   ↓ lowest
  ./packages
      4337-local-bundler        ^2.22.3   ↓ lowest
  ./modules
      4337                      ^2.22.5   ↑ highest

 ⨯ error Dependency typescript has multiple versions defined in the workspace. multiple-dependency-versions
  ./examples
      4337-gas-metering         ^5.4.5   ↓ lowest
  ./modules
      allowances                5.4.5   ∼ between
      passkey                   ^5.4.5   ↓ lowest
      recovery                  ^5.4.5   ↓ lowest
  ./packages
      4337-local-bundler        ^5.4.5   ↓ lowest
      4337-provider             ^5.4.5   ↓ lowest
  ./examples
      4337-passkeys             ^5.5.2   ↑ highest
  ./modules
      4337                      ^5.5.2   ↑ highest

 ⨯ error Dependency @types/node has multiple versions defined in the workspace. multiple-dependency-versions
  ./examples
      4337-gas-metering         20.14.0   ↓ lowest
  ./modules
      allowances                ^20.14.0   ∼ between
      passkey                   ^20.14.0   ∼ between
      recovery                  ^20.14.0   ∼ between
  ./packages
      4337-provider             ^20.14.0   ∼ between
  ./modules
      4337                      ^20.14.8   ↑ highest

 ⨯ error Dependency @openzeppelin/contracts has multiple versions defined in the workspace. multiple-dependency-versions
  ./modules
      recovery                  =4.9.6   ↓ lowest
      passkey                   ^5.0.0   ∼ between
      4337                      ^5.0.2   ↑ highest
      allowances                ^5.0.2   ↑ highest

 ⨯ error Dependency @safe-global/safe-deployments has multiple versions defined in the workspace. multiple-dependency-versions
  ./modules
      allowances                ^1.36.0   ↓ lowest
  ./examples
      4337-passkeys             ^1.37.0   ↑ highest

 ⨯ error Dependency @types/mocha has multiple versions defined in the workspace. multiple-dependency-versions
  ./modules
      allowances                ^10.0.6   ↓ lowest
      4337                      ^10.0.7   ↑ highest

 ⨯ error Dependency dotenv has multiple versions defined in the workspace. multiple-dependency-versions
  ./examples
      4337-gas-metering         16.4.5   ↓ lowest
  ./modules
      4337                      ^16.4.5   ↑ highest
      allowances                ^16.4.5   ↑ highest
      passkey                   ^16.4.5   ↑ highest
      recovery                  ^16.4.5   ↑ highest

 ⨯ error Dependency solhint has multiple versions defined in the workspace. multiple-dependency-versions
  ./modules
      4337                      ^5.0.1   ↑ highest
      allowances                5.0.1   ∼ between
      passkey                   ^5.0.1   ↑ highest

 ⨯ error Dependency @nomicfoundation/hardhat-network-helpers has multiple versions defined in the workspace. multiple-dependency-versions
  ./modules
      passkey                   ^1.0.10   ↓ lowest
      4337                      ^1.0.11   ↑ highest

 ⨯ error Dependency @simplewebauthn/server has multiple versions defined in the workspace. multiple-dependency-versions
  ./modules
      4337                      10.0.0   ↓ lowest
      passkey                   ^10.0.0   ↑ highest
```


modules/4337/package.json
```diff
@@ -47,15 +47,15 @@
     "url": "https://github.com/safe-global/safe-modules/issues"
   },
   "devDependencies": {
-    "@account-abstraction/contracts": "^0.7.0",
+    "@account-abstraction/contracts": "0.7.0",
```
nlordell, line 50: Should this be a dependency? I think we import types from it when building the contract:

https://github.com/safe-global/safe-modules/blob/bb5e1f7360633d5ecfb175c5d8667b2a3a1b746d/modules/4337/contracts/Safe4337Module.sol#L6-L9

modules/recovery/package.json
```diff
@@ -39,18 +39,18 @@
...
+    "ethers": "^6.13.1",
+    "hardhat": "^2.22.5",
     "hardhat-deploy": "^0.12.4",
-    "typescript": "^5.4.5",
+    "typescript": "^5.5.2",
     "yargs": "^17.7.2"
   },
   "dependencies": {
     "candide-contracts": "github:5afe/CandideWalletContracts#113d3c059e039e332637e8f686d9cbd505f1e738",
     "@openzeppelin/contracts": "=4.9.6",
```
nlordell, line 53: Should this be:
```suggestion
    "@openzeppelin/contracts": "4.9.6",
```

modules/recovery/package.json
```diff
@@ -39,18 +39,18 @@
...
-    "@safe-global/safe-contracts": "=1.4.1-build.0"
+    "@safe-global/safe-contracts": "^1.4.1-build.0"
```
nlordell, line 54: I wonder if we should use exact dependencies for these:
```suggestion
    "@safe-global/safe-contracts": "=1.4.1-build.0"
```

They affect the build and ultimately the bytecode - so not doing so could lead to issues where packages that depend on the contracts here build with different bytecode than what the contract build originally.

Same applies for other `@safe-global/safe-contracts` and `@account-abstraction/contracts` dependencies elsewhere.

.github/workflows/ci.yml
```diff
@@ -14,5 +14,8 @@ jobs:
         with:
           node-version: 20.x
           cache: pnpm
-      - run: pnpm install
-      - run: pnpm run fmt:global-check
+      - run: |
+          pnpm install
+          pnpm run fmt:global-check
+          # -i is to ignore packages that are expected to have multiple versions across the workspace
+          npx sherif@latest -i @openzeppelin/contracts
```
nlordell, line 21: I'm not a fan of this - it means that CI can break from one run to another because of new `sherif` rules.

Why not add it as a dependency to the root of the repository?

mmv08: This is their recommended way of running it: https://github.com/QuiiBz/sherif

I agree using `latest` is a mistake, I'll pin it to a specific version.

>  It can even be part of the fmt:global-check

It can be, but I am unsure if it's related to formatting 🤷

modules/recovery/package.json
mmv08: So I've done this in the recovery package (technically, we shouldn't have this dependency here, but it's listed as a dev dependency in their repo) and the 4337 one. I proposed a change to them in our shared Slack channel.

modules/4337/package.json
mmv08: Yes sir.

.github/workflows/ci.yml
mmv08: Pinned the version sir

examples/4337-gas-metering/package.json
```diff
@@ -53,14 +53,14 @@
...
-    "ethers": "^6.12.1",
+    "dotenv": "^16.4.5",
+    "ethers": "^6.13.1",
     "permissionless": "0.1.29",
     "viem": "2.12.5"
   },
   "devDependencies": {
-    "@types/node": "20.14.0",
+    "@types/node": "^20.14.8",
     "tsx": "4.11.0",
```
nlordell, line 63: Unrelated nit that I'm just noticing now: it feels weird that we write dependencies differently for different packages in our repository. Here, we are using "x.y.z", while in others we are using "^x.y.z", even though there is nothing special about these dependencies for expressing them differently from a semver perspective.

Not for this PR, just something that I wanted to bring up.

modules/passkey/package.json
```diff
@@ -48,24 +48,24 @@
...
     "ts-node": "^10.9.2",
-    "typescript": "^5.4.5"
+    "typescript": "^5.5.2"
   },
   "dependencies": {
-    "@account-abstraction/contracts": "^0.7.0",
-    "@openzeppelin/contracts": "^5.0.0",
+    "@account-abstraction/contracts": "0.7.0",
+    "@openzeppelin/contracts": "^5.0.2",
     "@safe-global/safe-contracts": "^1.4.1-build.0",
```
nlordell, line 69: I think this should also have strict dependencies.

modules/4337/package.json
```diff
@@ -74,6 +73,7 @@
     "yargs": "^17.7.2"
   },
   "dependencies": {
-    "@safe-global/safe-contracts": "^1.4.1-build.0"
+    "@account-abstraction/contracts": "0.7.0",
+    "@safe-global/safe-contracts": "1.4.1-build.0"
```
nlordell, line 77: And these too.

.github/workflows/ci.yml
mmv08: I migrated it to the package.json file, and one thing that I missed straight away is the comments 🙈 I have no way to explain what the `-i` flag does and why certain packages are excluded

nlordell: `git blame` :stuck_out_tongue:

True that JSON doesn't allow for comments which is annoying :disappointed:

modules/4337/package.json
mmv08: This is already strict, isn't it?

modules/passkey/package.json
mmv08: From what I found, passkeys only used safe contracts for testing and a version update wouldn't result in a new bytecode, so I reckon it's fine as is. I can pin it though

modules/4337/package.json
nlordell: Oops.

modules/passkey/package.json
nlordell: In that case, shouldn't these be `devDependencies`?

package.json
```diff
@@ -39,6 +42,7 @@
     "eslint-plugin-react-refresh": "^0.4.7",
     "prettier": "^3.3.0",
     "prettier-plugin-solidity": "^1.3.1",
-    "rimraf": "^5.0.7"
+    "rimraf": "^5.0.7",
+    "sherif": "0.9.0"
```
nlordell, line 46:
```suggestion
    "sherif": "^0.9.0"
```

nlordell (approved):

nlordell:
One more question here: https://github.com/safe-global/safe-modules/pull/452#discussion_r1668204779, but approving to reduce back-and-forth.

package.json
mmv08: They may not be following semver right?

examples/4337-gas-metering/package.json
mmv08: Maybe there's a purpose, the problem is that there are no comments allowed in the package.json, so that knowledge may be lost. There are some hacky solutions though: https://stackoverflow.com/questions/14221579/how-do-i-add-comments-to-package-json-for-npm-install

package.json
nlordell: Don't think that matters - we have a lock file, so the version gets pinned anyway. Also I _think_ this version _should_ follow semver (not super clear, but [the docs](https://docs.npmjs.com/cli/v10/configuring-npm/package-json#version) say that it "must be parseable by [node-semver](https://github.com/npm/node-semver)", which itself is a semantic versioning package for NodeJS).

examples/4337-gas-metering/package.json
nlordell: These packages are getting pinned by the lock file. The places where versioning matters more are for `dependencies`, because those get transitively installed.

I think that `git blame` might be able to include documentation for these.

modules/passkey/package.json
```diff
@@ -48,25 +48,25 @@
...
     "ts-node": "^10.9.2",
-    "typescript": "^5.4.5"
+    "typescript": "^5.5.2"
   },
   "dependencies": {
-    "@account-abstraction/contracts": "^0.7.0",
-    "@openzeppelin/contracts": "^5.0.0",
-    "@safe-global/safe-contracts": "^1.4.1-build.0",
+    "@account-abstraction/contracts": "0.7.0",
+    "@openzeppelin/contracts": "^5.0.2",
```
nlordell, line 69: This is a transitive dependency from account-abstraction (whose package has a bug and doesn't correctly include the contracts as a dependency) and uses version 5.0.0:

https://github.com/eth-infinitism/account-abstraction/blob/7af70c8993a6f42973f520ae0752386a5032abe7/yarn.lock#L901-L904

I would suggest marking this as:

```suggestion
    "@openzeppelin/contracts": "=5.0.0",
```

To indicate this is an exact version on purpose.

nlordell (approved):

modules/passkey/package.json
mmv08: Actually, they do include it as a dependency (I created a PR to fix it). We have it because hardhat can't resolve pnpm modules correctly: https://github.com/NomicFoundation/hardhat/issues/4292

nlordell: Ah, good find.

Should we use 5.0.0 though, so it matches the version in the repository?

mmv08: We only use the AA dependency in tests and do not depend on the same bytecode, so I think it's fine. Openzeppelin package is also following semver so I think it's in the allowed version range.

nlordell: Wait - if its only used in tests, then I think it should be a `devDependency`.

nlordell (approved):
Approving once again - I think some of the discussions around dependencies should happen, but they don't need to for this PR to merge since this is only about adding `sherif` for linting (so I'm sorry I keep on bringing them up :see_no_evil:).

PR #453, opened 2024-07-01
Signature length check

akshay-ap:
Related to https://github.com/safe-global/safe-smart-account/issues/754

## Changes in PR

- Add a test case that verifies that UserOp is reverted if signature contains additional bytes data
- Add `_checkSignatureLength` function  in `Safe4437Module` contract . This function retrieves threshold from Safe and iterates over signature data.
- Add a test contract which is used in testing the internal function:
- Add tests
- Summarise `_checkSignatureLength` in FV specs. The prover was reporting error without giving call traces.
- Unrelated to issue: [Fix script](https://github.com/safe-global/safe-modules/pull/453/files#diff-d6fef5446f8c9e62fd4180360854e7e830def1aa48a27dda677a99a74fcd6a44R24) in `package.json`

## Open for discussion

Should `_checkSignatureLength` also do more stricter checks that are already there in `checkNSignature` function of `Safe`? If `Safe4337Module` does not perform below mentioned checks then it will be implicitly assumed to be done in `Safe` contract and this creates a requirement that all future `Safe` versions to have these checks if used with this `Safe4337Module`.

1. `if (signatures.length < offset)` this check is currently done twice. 
    - In Safe4337Module: https://github.com/safe-global/safe-modules/pull/453/files#diff-8ec9433e2f98b74922f038206ac421a769cf3996997ad55ed210299abddb7908R223
    - In Safe: https://github.com/safe-global/safe-smart-account/blob/8f80a8372d193be121dcdb52e869a258824e5c0f/contracts/Safe.sol#L279
   
2.  Other potential check that can be repeated: https://github.com/safe-global/safe-smart-account/blob/8f80a8372d193be121dcdb52e869a258824e5c0f/contracts/Safe.sol#L233

### This PR

```
Safe4337Module 8910 bytes (limit is 24576)
```

### Main branch

```
Safe4337Module 8373 bytes (limit is 24576)
```

### This PR

```
  Gas Metering
    Safe Deployment + Enabling 4337 Module
           Used 420045 gas (Account or Paymaster) for >Safe with 4337 Module Deployment<
           Used 416973 gas (Transaction) for >Safe with 4337 Module Deployment<
      ✔ Safe with 4337 Module Deployment
    Safe Deployment + Enabling 4337 Module + Native Transfers
           Used 451408 gas (Account or Paymaster) for >Safe with 4337 Module Deployment + Native Transfer<
           Used 449392 gas (Transaction) for >Safe with 4337 Module Deployment + Native Transfer<
      ✔ Safe with 4337 Module Deployment + Native Transfer
           Used 193399 gas (Account or Paymaster) for >Safe with 4337 Module Native Transfer<
           Used 183844 gas (Transaction) for >Safe with 4337 Module Native Transfer<
      ✔ Safe with 4337 Module Native Transfer
    Safe Deployment + Enabling 4337 Module + Token Operations
           Used 436781 gas (Account or Paymaster) for >Safe with 4337 Module Deployment + ERC20 Transfer<
           Used 427911 gas (Transaction) for >Safe with 4337 Module Deployment + ERC20 Transfer<
      ✔ Safe with 4337 Module Deployment + ERC20 Token Transfer
           Used 471043 gas (Account or Paymaster) for >Safe with 4337 Module Deployment + ERC721 Transfer<
           Used 469710 gas (Transaction) for >Safe with 4337 Module Deployment + ERC721 Transfer<
      ✔ Safe with 4337 Module Deployment + ERC721 Token Minting
    Token Operations Only
           Used 177996 gas (Account or Paymaster) for >Safe with 4337 Module ERC20 Transfer<
           Used 162338 gas (Transaction) for >Safe with 4337 Module ERC20 Transfer<
      ✔ Safe with 4337 Module ERC20 Token Transfer
           Used 213064 gas (Account or Paymaster) for >Safe with 4337 Module ERC721 Token Minting<
           Used 204125 gas (Transaction) for >Safe with 4337 Module ERC721 Token Minting<
      ✔ Safe with 4337 Module ERC721 Token Minting
```

### Main branch
```
  Gas Metering
    Safe Deployment + Enabling 4337 Module
           Used 419096 gas (Account or Paymaster) for >Safe with 4337 Module Deployment<
           Used 414864 gas (Transaction) for >Safe with 4337 Module Deployment<
      ✔ Safe with 4337 Module Deployment
    Safe Deployment + Enabling 4337 Module + Native Transfers
           Used 450537 gas (Account or Paymaster) for >Safe with 4337 Module Deployment + Native Transfer<
           Used 447308 gas (Transaction) for >Safe with 4337 Module Deployment + Native Transfer<
      ✔ Safe with 4337 Module Deployment + Native Transfer
           Used 192424 gas (Account or Paymaster) for >Safe with 4337 Module Native Transfer<
           Used 182182 gas (Transaction) for >Safe with 4337 Module Native Transfer<
      ✔ Safe with 4337 Module Native Transfer
    Safe Deployment + Enabling 4337 Module + Token Operations
           Used 435685 gas (Account or Paymaster) for >Safe with 4337 Module Deployment + ERC20 Transfer<
           Used 425687 gas (Transaction) for >Safe with 4337 Module Deployment + ERC20 Transfer<
      ✔ Safe with 4337 Module Deployment + ERC20 Token Transfer
           Used 470086 gas (Account or Paymaster) for >Safe with 4337 Module Deployment + ERC721 Transfer<
           Used 467491 gas (Transaction) for >Safe with 4337 Module Deployment + ERC721 Transfer<
      ✔ Safe with 4337 Module Deployment + ERC721 Token Minting
    Token Operations Only
           Used 176846 gas (Account or Paymaster) for >Safe with 4337 Module ERC20 Transfer<
           Used 160584 gas (Transaction) for >Safe with 4337 Module ERC20 Transfer<
      ✔ Safe with 4337 Module ERC20 Token Transfer
           Used 212013 gas (Account or Paymaster) for >Safe with 4337 Module ERC721 Token Minting<
           Used 202376 gas (Transaction) for >Safe with 4337 Module ERC721 Token Minting<
      ✔ Safe with 4337 Module ERC721 Token Minting
```


modules/4337/certora/specs/Safe4337Module.spec
```diff
@@ -25,6 +25,8 @@ methods {
     function getOperationHash(
         Safe4337Module.PackedUserOperation userOp
     ) external returns(bytes32) envfree => PER_CALLEE_CONSTANT;
+    function _checkSignatureLength(bytes calldata, uint256) internal returns(bool) => ALWAYS(true);
+
```
nlordell, line 29:
```suggestion
```

modules/4337/certora/specs/ValidationDataLastBitOne.spec
```diff
@@ -11,6 +11,8 @@ methods {
     function getOperationHash(
         Safe4337Module.PackedUserOperation userOp
     ) external returns(bytes32) envfree => PER_CALLEE_CONSTANT;
+
```
nlordell, line 14:
```suggestion
```

modules/4337/contracts/Safe4337Module.sol
```diff
@@ -211,6 +211,42 @@ contract Safe4337Module is IAccount, HandlerContext, CompatibilityFallbackHandle
...
+                    let contractSignatureLen := calldataload(add(signatures.offset, s))
+                    // Update the required position of the next offset.
+                    offset := add(add(s, 0x20), contractSignatureLen)
+                }
+            }
+            /* solhint-enable no-inline-assembly */
+            if (!result) return result;
+        }
+        if (signatures.length > offset) return false;
+        return result;
```
nlordell, line 247: In this file at least, we use the style of returning by fallthrough:

```suggestion
        result = signatures.length <= offset;
```

modules/4337/contracts/Safe4337Module.sol
```diff
@@ -211,6 +211,42 @@ contract Safe4337Module is IAccount, HandlerContext, CompatibilityFallbackHandle
...
+     * @param signatures signatures data
+     * @param threshold Indicates the number of iterations to perform in the loop.
+     * @return result True if length check passes, false otherwise.
+     */
+    function _checkSignatureLength(bytes calldata signatures, uint256 threshold) internal pure returns (bool result) {
+        uint256 offset = threshold * 0x41;
+
+        if (signatures.length < offset) return false;
+
+        result = true;
```
nlordell, line 225: nit: Maybe we shouldn't reuse this variable, it makes the code harder to understand IMO.

modules/4337/contracts/Safe4337Module.sol
```diff
@@ -222,11 +258,19 @@ contract Safe4337Module is IAccount, HandlerContext, CompatibilityFallbackHandle
...
+
+        if (!success) {
             validationData = _packValidationData(true, validUntil, validAfter);
+        } else {
+            try ISafe(payable(userOp.sender)).checkSignatures(keccak256(operationData), operationData, signatures) {
+                // The timestamps are validated by the entry point, therefore we will not check them again
+                validationData = _packValidationData(false, validUntil, validAfter);
+            } catch {
+                validationData = _packValidationData(true, validUntil, validAfter);
+            }
```
nlordell, line 273: nit: Not a huge fan of having separate failure branches - why not:

```suggestion
        bool validSignature;
        try ISafe(payable(userOp.sender)).checkSignatures(keccak256(operationData), operationData, signatures) {
            validSignature = true;
        } catch {
            validSignature = false;
        }

        // TODO: Comment explaining why we validate signature length.
        validSignature = validSignature
            && SignatureLength(signatures, ISafe(payable(userOp.sender)).getThreshold());

        if (validSignature) {
            // The timestamps are validated by the entry point, therefore we will not check them again
            validationData = _packValidationData(false, validUntil, validAfter);
        } else {
            validationData = _packValidationData(true, validUntil, validAfter);
        }
```

modules/4337/contracts/Safe4337Module.sol
```diff
@@ -211,6 +211,42 @@ contract Safe4337Module is IAccount, HandlerContext, CompatibilityFallbackHandle
         operationHash = keccak256(operationData);
     }
 
+    /**
+     * @dev Checks if the signatures length is correct and does not contain addtional bytes.
+     * @param signatures signatures data
```
nlordell, line 216: nit: missing explicit documentation that this does not check integrity of encoding, as the Safe.checkSignatures already does this.

akshay-ap: https://github.com/safe-global/safe-modules/pull/453/commits/9e526d6de72915fcafc4fb4b0a9f1013821656fc

modules/4337/contracts/Safe4337Module.sol
```diff
@@ -258,16 +258,17 @@ contract Safe4337Module is IAccount, HandlerContext, CompatibilityFallbackHandle
     function _validateSignatures(PackedUserOperation calldata userOp) internal view returns (uint256 validationData) {
         (bytes memory operationData, uint48 validAfter, uint48 validUntil, bytes calldata signatures) = _getSafeOp(userOp);
 
+        // checkSignatures function in Safe contract does not force a fixed size on signature length. A malicious actor can append additional bytes in the signature than needed and force gas usage to reach verificationGasLimit. _checkSignatureLength ensures that there are no additional bytes in the signatures than required.
+        if (!_checkSignatureLength(signatures, ISafe(payable(userOp.sender)).getThreshold()))
+            return _packValidationData(true, validUntil, validAfter);
+
         bool validSignature;
         try ISafe(payable(userOp.sender)).checkSignatures(keccak256(operationData), operationData, signatures) {
             validSignature = true;
```
nlordell, line 267: My issue with this is the same as before, we now have multiple exit branches which feels a bit icky IMO. This also has signature verification gotcha's in that if you pass an invalid length signature for gas estimations, it will not estimate correctly (so same problem you were having, but other way around.

Maybe we can do:

```
        validSignature = _checkSignatureLength(signatures, ISafe(payable(userOp.sender)).getThreshold()));
        try ISafe(payable(userOp.sender)).checkSignatures(keccak256(operationData), operationData, signatures) {
        } catch {
            validSignature = false;
        }
```

modules/4337/contracts/Safe4337Module.sol
```diff
@@ -258,16 +258,17 @@ contract Safe4337Module is IAccount, HandlerContext, CompatibilityFallbackHandle
     function _validateSignatures(PackedUserOperation calldata userOp) internal view returns (uint256 validationData) {
         (bytes memory operationData, uint48 validAfter, uint48 validUntil, bytes calldata signatures) = _getSafeOp(userOp);
 
+        // checkSignatures function in Safe contract does not force a fixed size on signature length. A malicious actor can append additional bytes in the signature than needed and force gas usage to reach verificationGasLimit. _checkSignatureLength ensures that there are no additional bytes in the signatures than required.
```
nlordell, line 261: Nit: I think all other comments have been split at either 100 or 120 column mark, and we have some markdown-y ways for referencing identifiers (so `checkSignatures`). Can you keep this long comment format consistent here as well?

The reason for breaking comments into multiple lines is so that it is generally easier to read. For example, the GitHub code pane does not wrap long comment lines:

image

modules/4337/contracts/Safe4337Module.sol
mmv08: > This also has signature verification gotcha's in that if you pass an invalid length signature for gas estimations, it will not estimate correctly (so same problem you were having, but other way around.

I think this is fine. It still applies the other way around. If you pass an invalid length signature (but shorter), it will not estimate correctly because of the checks in the safe contract or the verifier if you use passkeys

nlordell: Right - but it is fairly trivial to always call `checkSignatures` (which is what I'm suggesting).

modules/4337/certora/specs/ValidationDataLastBitOne.spec
```diff
@@ -11,6 +11,7 @@ methods {
     function getOperationHash(
         Safe4337Module.PackedUserOperation userOp
     ) external returns(bytes32) envfree => PER_CALLEE_CONSTANT;
+    function _checkSignatureLength(bytes calldata, uint256) internal returns(bool)=> ALWAYS(true);
```
remedcu, line 14:
```suggestion
    function _checkSignatureLength(bytes calldata, uint256) internal returns(bool) => ALWAYS(true);
```

modules/4337/contracts/Safe4337Module.sol
```diff
@@ -211,6 +211,41 @@ contract Safe4337Module is IAccount, HandlerContext, CompatibilityFallbackHandle
         operationHash = keccak256(operationData);
     }
 
+    /**
+     * @dev Checks if the signatures length is correct and does not contain addtional bytes. The function does not check the integrity of the signature encoding.
```
remedcu, line 215:
```suggestion
     * @dev Checks if the signatures length is correct and does not contain additional bytes. The function does not check the integrity of the signature encoding.
```

modules/4337/test/erc4337/ERC4337ModuleNew.spec.ts
```diff
@@ -119,6 +119,37 @@ describe('Safe4337Module - Newly deployed safe', () => {
       expect(await ethers.provider.getBalance(safe.address)).to.be.eq(ethers.parseEther('0'))
     })
 
+    it('should revert when signature length contains additional bytes', async () => {
```
mmv08, line 122: This test uses an eoa signature, while the test with the reference entrypoint uses a smart contract signature - why?
Why not both in each of the test files?

modules/4337/contracts/Safe4337Module.sol
```diff
@@ -211,6 +211,41 @@ contract Safe4337Module is IAccount, HandlerContext, CompatibilityFallbackHandle
...
+        result = true;
+        for (uint256 i = 0; i < threshold; i++) {
+            /* solhint-disable no-inline-assembly */
+            /// @solidity memory-safe-assembly
+            assembly {
+                let signaturePos := mul(0x41, i)
+                let s := calldataload(add(signatures.offset, add(signaturePos, 0x20)))
+                let v := byte(0, calldataload(add(signatures.offset, add(signaturePos, 0x40))))
+
+                if iszero(v) {
```
mmv08, line 234: I'd add a comment that v = 0 means a smart contract signature in Safe Signature Encoding

modules/4337/contracts/Safe4337Module.sol
```diff
@@ -211,6 +211,41 @@ contract Safe4337Module is IAccount, HandlerContext, CompatibilityFallbackHandle
...
+        if (signatures.length < offset) return false;
+
+        result = true;
+        for (uint256 i = 0; i < threshold; i++) {
+            /* solhint-disable no-inline-assembly */
+            /// @solidity memory-safe-assembly
+            assembly {
+                let signaturePos := mul(0x41, i)
+                let s := calldataload(add(signatures.offset, add(signaturePos, 0x20)))
+                let v := byte(0, calldataload(add(signatures.offset, add(signaturePos, 0x40))))
```
mmv08, line 232: 1. You can move the `s` to the if block, because it's only used there.
2. variable names can be more descriptive - since we're dealing with smart contract signatures, the values are not `s` and `v` per ser se, but `s` is a pointer to the start of the signature data and `v` is the signature type

modules/4337/contracts/Safe4337Module.sol
akshay-ap: https://github.com/safe-global/safe-modules/pull/453/commits/af3b7088120302b3b14a9ef74de4c0605a0192f7

modules/4337/test/erc4337/ERC4337ModuleNew.spec.ts
akshay-ap: I thought having a test with Reference entrypoint should be enough. No harm in adding more tests with different signature types as you suggested.
Done here: https://github.com/safe-global/safe-modules/pull/453/commits/22fc197420ae76457903c47b26a72e8e8f40d5c7

modules/4337/contracts/Safe4337Module.sol
```diff
@@ -211,6 +211,44 @@ contract Safe4337Module is IAccount, HandlerContext, CompatibilityFallbackHandle
...
+            /* solhint-disable no-inline-assembly */
+            /// @solidity memory-safe-assembly
+            assembly {
+                let signaturePos := mul(0x41, i)
+                // read the Safe signature type (uint8) from calldata.
+                let signatureType := byte(0, calldataload(add(signatures.offset, add(signaturePos, 0x40))))
+
+                // signatureType = 0 indicates that signature is a smart contract signature in Safe Signature Encoding
+                if iszero(signatureType) {
+                    // For Safe smart contract signature the s value points to the start of the signature data in the signatures
```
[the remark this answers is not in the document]
akshay-ap: The `s` value is not relative.
https://github.com/safe-global/safe-modules/blob/c403aff29bed6e3e326ab36c820094f2ada41c23/modules/4337/src/utils/execution.ts#L114

or do you mean something else?

modules/4337/contracts/Safe4337Module.sol
akshay-ap: here: https://github.com/safe-global/safe-modules/pull/453/commits/53345a672f35578cb0d320b28f5f891b42386e70

modules/4337/contracts/Safe4337Module.sol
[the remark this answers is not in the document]
remedcu: I thought it was relative to the `signature.offset` as we use that for retrieving the length parameter at the start of the dynamic part:

https://github.com/safe-global/safe-modules/blob/53345a672f35578cb0d320b28f5f891b42386e70/modules/4337/contracts/Safe4337Module.sol#L240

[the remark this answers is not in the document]
akshay-ap: ohh.. now I get it. `s` value is relative to start of signature i.e. `signatures.offset` from calldata here.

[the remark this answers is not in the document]
akshay-ap: https://github.com/safe-global/safe-modules/pull/453/commits/1c31fda9ca0378a9d8e8ebea6cdc0056a9f7aa89

modules/4337/test/erc4337/ERC4337ModuleNew.spec.ts
```diff
@@ -119,6 +123,92 @@ describe('Safe4337Module - Newly deployed safe', () => {
...
+
+      const accountBalance = ethers.parseEther('1.0')
+      await user1.sendTransaction({ to: daughterSafe.address, value: accountBalance })
+      expect(await ethers.provider.getBalance(daughterSafe.address)).to.be.eq(accountBalance)
+
+      const transfer = ethers.parseEther('0.1')
+      const safeOp = buildSafeUserOpTransaction(
+        daughterSafe.address,
+        user1.address,
+        transfer,
```
remedcu, line 172:
```suggestion
      const safeOp = buildSafeUserOpTransaction(
        daughterSafe.address,
        user1.address,
        ethers.parseEther('0.1'),
```

As we use the value directly in other places as well.

remedcu (approved):
LGTM

modules/4337/contracts/Safe4337Module.sol
```diff
@@ -211,6 +211,43 @@ contract Safe4337Module is IAccount, HandlerContext, CompatibilityFallbackHandle
         operationHash = keccak256(operationData);
     }
 
+    /**
+     * @dev Checks if the signatures length is correct and does not contain additional bytes. The function does not check the integrity of the signature encoding.
```
nlordell, line 215: nit: Can you keep the line length down? I can't remember if we use 100 or 120 characters (take a look at other comments).

nlordell:
```suggestion
     * @dev Checks if the signatures length is correct and does not contain additional bytes. The function does not check the integrity of the signature encoding, as this is expected to be checked by the {Safe} implementation of {checkSignatures}.
```

modules/4337/contracts/Safe4337Module.sol
```diff
@@ -211,6 +211,43 @@ contract Safe4337Module is IAccount, HandlerContext, CompatibilityFallbackHandle
...
+            /// @solidity memory-safe-assembly
+            assembly {
+                let signaturePos := mul(0x41, i)
+                // read the Safe signature type (uint8) from calldata.
+                let signatureType := byte(0, calldataload(add(signatures.offset, add(signaturePos, 0x40))))
+
+                // signatureType = 0 indicates that signature is a smart contract signature in Safe Signature Encoding
+                if iszero(signatureType) {
+                    // For Safe smart contract signature the s value points to the start of the signature data in the signatures
+                    // s value is relative to the signature data.
```
nlordell, line 236: nit: Comment is outdated as there is no "s" value anymore. Maybe you could say the "second word"?

modules/4337/contracts/Safe4337Module.sol
```diff
@@ -211,6 +211,43 @@ contract Safe4337Module is IAccount, HandlerContext, CompatibilityFallbackHandle
...
+        uint256 offset = threshold * 0x41;
+
+        if (signatures.length < offset) return false;
+        bool pointsAtEnd = true;
+        for (uint256 i = 0; i < threshold; i++) {
+            /* solhint-disable no-inline-assembly */
+            /// @solidity memory-safe-assembly
+            assembly {
+                let signaturePos := mul(0x41, i)
+                // read the Safe signature type (uint8) from calldata.
```
nlordell, line 230:
```suggestion
                // read the Safe signature type byte from the signatures bytes, this is the 64th byte per signature.
```

modules/4337/contracts/Safe4337Module.sol
```diff
@@ -211,6 +211,43 @@ contract Safe4337Module is IAccount, HandlerContext, CompatibilityFallbackHandle
...
+     * @param signatures signatures data
+     * @param threshold Indicates the number of iterations to perform in the loop.
+     * @return result True if length check passes, false otherwise.
+     */
+    function _checkSignatureLength(bytes calldata signatures, uint256 threshold) internal pure returns (bool) {
+        uint256 offset = threshold * 0x41;
+
+        if (signatures.length < offset) return false;
+        bool pointsAtEnd = true;
+        for (uint256 i = 0; i < threshold; i++) {
```
nlordell, line 225: uber-nit: This is only needed in the `for` loop:

```suggestion
        for (uint256 i = 0; i < threshold; i++) {
            bool pointsAtEnd = true;
```

modules/4337/contracts/Safe4337Module.sol
```diff
@@ -211,6 +211,43 @@ contract Safe4337Module is IAccount, HandlerContext, CompatibilityFallbackHandle
...
     }
 
+    /**
+     * @dev Checks if the signatures length is correct and does not contain additional bytes. The function does not check the integrity of the signature encoding.
+     * @param signatures signatures data
+     * @param threshold Indicates the number of iterations to perform in the loop.
+     * @return result True if length check passes, false otherwise.
+     */
+    function _checkSignatureLength(bytes calldata signatures, uint256 threshold) internal pure returns (bool) {
+        uint256 offset = threshold * 0x41;
```
nlordell, line 221: This is the "expected offset" right?

modules/4337/contracts/Safe4337Module.sol
```diff
@@ -211,6 +211,43 @@ contract Safe4337Module is IAccount, HandlerContext, CompatibilityFallbackHandle
...
+
+                // signatureType = 0 indicates that signature is a smart contract signature in Safe Signature Encoding
+                if iszero(signatureType) {
+                    // For Safe smart contract signature the s value points to the start of the signature data in the signatures
+                    // s value is relative to the signature data.
+                    let signatureStartPointer := calldataload(add(signatures.offset, add(signaturePos, 0x20)))
+                    pointsAtEnd := eq(signatureStartPointer, offset)
+                    let contractSignatureLen := calldataload(add(signatures.offset, signatureStartPointer))
+                    // Update the required position of the next offset.
+                    offset := add(add(signatureStartPointer, 0x20), contractSignatureLen)
```
nlordell, line 241: I would explain "why add 0x20?":

```suggestion
                    // Update the expected offset of the next contract signature. This is the previous expected offset plus the total length of the current contract signature. Note that we add 0x20 to the contract signature length to account for the encoded length value.
                    offset := add(offset, add(0x20, contractSignatureLen))
```

modules/4337/contracts/Safe4337Module.sol
```diff
@@ -211,6 +211,43 @@ contract Safe4337Module is IAccount, HandlerContext, CompatibilityFallbackHandle
         operationHash = keccak256(operationData);
     }
 
+    /**
+     * @dev Checks if the signatures length is correct and does not contain additional bytes. The function does not check the integrity of the signature encoding.
+     * @param signatures signatures data
+     * @param threshold Indicates the number of iterations to perform in the loop.
+     * @return result True if length check passes, false otherwise.
+     */
+    function _checkSignatureLength(bytes calldata signatures, uint256 threshold) internal pure returns (bool) {
```
nlordell, line 220: There is a NatSpec issue here, `result` is not the name of the return value, but the first work in the documentation (since there is no named field).

Also, in the past audits, OZ pointed out that we were inconstent with named return values and asked us to be consistent everywhere (i.e. always, or never, use named returns) which is why I had suggested to use one previously:

```suggestion
     * @return isValue True if length check passes, false otherwise.
     */
    function _checkSignatureLength(bytes calldata signatures, uint256 threshold) internal pure returns (bool isValid) {
```

modules/4337/contracts/Safe4337Module.sol
```diff
@@ -211,6 +211,43 @@ contract Safe4337Module is IAccount, HandlerContext, CompatibilityFallbackHandle
...
+                    let contractSignatureLen := calldataload(add(signatures.offset, signatureStartPointer))
+                    // Update the required position of the next offset.
+                    offset := add(add(signatureStartPointer, 0x20), contractSignatureLen)
+                }
+            }
+            /* solhint-enable no-inline-assembly */
+            // If the signature data pointer is not pointing to the expected location, return false.
+            if (!pointsAtEnd) return false;
+        }
+        return signatures.length <= offset;
```
nlordell, line 248: Related to comment :point_up: 

```suggestion
        isValid = signatures.length <= offset;
```

nlordell (commented):
I think the tests don't have full coverage - All tests fail with the final check (i.e. whether or not the signature length matches the final expected offset), but AFAICT, none of the fail on the `if (!pointsToEnd)` check by inserting additional bytes in between the dynamic signatures.

modules/4337/contracts/Safe4337Module.sol
akshay-ap: https://github.com/safe-global/safe-modules/pull/453/commits/979b6a4574003e6c1a0beaad4aefe9e8aa653c83

akshay-ap:
> I think the tests don't have full coverage - All tests fail with the final check (i.e. whether or not the signature length matches the final expected offset), but AFAICT, none of the fail on the `if (!pointsToEnd)` check by inserting additional bytes in between the dynamic signatures.

https://github.com/safe-global/safe-modules/pull/453/commits/1ec98133eabbfd10ee60ba4186ab0d52241586a2

modules/4337/contracts/Safe4337Module.sol
```diff
@@ -211,6 +211,45 @@ contract Safe4337Module is IAccount, HandlerContext, CompatibilityFallbackHandle
...
+    function _checkSignatureLength(bytes calldata signatures, uint256 threshold) internal pure returns (bool isValid) {
+        uint256 expectedOffset = threshold * 0x41;
+
+        for (uint256 i = 0; i < threshold; i++) {
+            bool pointsAtEnd = true;
+            /* solhint-disable no-inline-assembly */
+            /// @solidity memory-safe-assembly
+            assembly {
+                let signaturePos := mul(0x41, i)
+                // read the Safe signature type byte from the signatures bytes, this is the 64th byte per signature.
```
nlordell, line 231:
```suggestion
                // read the Safe signature type byte from the signatures bytes, this is the 64th byte per signature.
```


I just realize that my suggestion was ambiguous, it is the 65th byte, or byte at index 64.

modules/4337/contracts/Safe4337Module.sol
```diff
@@ -222,10 +261,20 @@ contract Safe4337Module is IAccount, HandlerContext, CompatibilityFallbackHandle
...
+            validSignature = false;
+        }
+
+        if (validSignature) {
             // The timestamps are validated by the entry point, therefore we will not check them again
             validationData = _packValidationData(false, validUntil, validAfter);
-        } catch {
+        } else {
             validationData = _packValidationData(true, validUntil, validAfter);
         }
```
nlordell, line 279: I just realized, but this can be simplified to:

```solidity
        validationData = _packValidationData(!validSignature, validUntil, validAfter);
```

modules/4337/test/erc4337/Safe4337Module.spec.ts
```diff
@@ -268,6 +268,40 @@ describe('Safe4337Module', () => {
...
+      const userOp = buildPackedUserOperationFromSafeUserOperation({
+        safeOp,
+        signature,
+      })
+      const packedValidationData = packValidationData(1, validUntil, validAfter)
+      const entryPointImpersonator = await ethers.getSigner(await entryPoint.getAddress())
+      const safeFromEntryPoint = safeModule.connect(entryPointImpersonator)
+
+      expect(await safeFromEntryPoint.validateUserOp.staticCall(userOp, ethers.ZeroHash, 0)).to.eq(packedValidationData)
+    })
```
nlordell, line 304: You already added tests to the `ReferenceEntryPoint.spec.ts`, so no need to re-add them here, but my personal opinion is that getting the full branch coverage makes more sense with the unit tests here, and not in the above "light integration tests" from the `ReferenceEntryPoint.spec.ts` (and avoids the extra long test names because of the "Note: ...").

Additionally, I still think a test is missing - in particular, the new test with the  invalid dynamic part pointer would have failed _before_ this PR. While the invalid dynamic part pointer is nice to have, I think we also need to add a test with a **valid** Safe signature that only fails the `_checkSignatureLength` function.

modules/4337/contracts/Safe4337Module.sol
```diff
@@ -211,6 +211,45 @@ contract Safe4337Module is IAccount, HandlerContext, CompatibilityFallbackHandle
...
 
+    /**
+     * @dev Checks if the signatures length is correct and does not contain additional bytes. The function does not
+     * check the integrity of the signature encoding, as this is expected to be checked by the {Safe} implementation
+     * of {checkSignatures}.
+     * @param signatures signatures data
+     * @param threshold Indicates the number of iterations to perform in the loop.
+     * @return isValid True if length check passes, false otherwise.
+     */
+    function _checkSignatureLength(bytes calldata signatures, uint256 threshold) internal pure returns (bool isValid) {
```
nlordell, line 222:
```suggestion
    function _checkSignaturesLength(bytes calldata signatures, uint256 threshold) internal pure returns (bool isValid) {
```

modules/4337/contracts/Safe4337Module.sol
akshay-ap: https://github.com/safe-global/safe-modules/pull/453/commits/2743e53077f87e34811069ff1f2595d6a4f75492

modules/4337/contracts/Safe4337Module.sol
akshay-ap: https://github.com/safe-global/safe-modules/pull/453/commits/025234faf4e2a372a76c4b6c740c220e498513c4

modules/4337/test/erc4337/Safe4337Module.spec.ts
akshay-ap: Added tests here: https://github.com/safe-global/safe-modules/pull/453/commits/c54c89d722c732bee2da399ea7e15e8dd572ef14

I think having tests in both `ReferenceEntryPoint.spec.ts` and `SafeModule.spec.ts` would be good to have because `SafeModule.spec.ts` uses mock safe instance and `ReferenceEntryPoint.spec.ts` uses actual Safe instance. Signature verification flow for 4337 module being critical I think we should also have tests with actual Safe instance.

nlordell: I agree that we _should_ have a test in `ReferenceEntryPoint.spec.ts`, but since this is more of an "integration test light", I don't think we need to necessarily fully cover all paths in all files - and it felt a little strange to me that in the "unit test" file, we have the least amount of coverage, while in the "integration light test" files, we have the fuller coverage.

The tests are there, definitely don't take them out. Just something I wanted to point out.

modules/4337/test/erc4337/ReferenceEntryPoint.spec.ts
```diff
@@ -215,6 +215,208 @@ describe('Safe4337Module - Reference EntryPoint', () => {
...
+      safeOp,
+      signature: signature.concat('00'), // adding '00' invalidates signature length
+    })
+
+    await expect(entryPoint.handleOps([userOp], await relayer.getAddress()))
+      .to.be.revertedWithCustomError(entryPoint, 'FailedOp')
+      .withArgs(0, 'AA24 signature error')
+  })
+
+  it('should revert on invalid signature length - Smart contract signature (NOTE: would require a staked paymaster for ERC-4337)', async () => {
```
nlordell, line 250: The signature check fails, I don't think staked paymasters matter here:

```suggestion
  it('should revert on invalid signature length - Smart contract signature', async () => {
```

modules/4337/test/erc4337/ReferenceEntryPoint.spec.ts
```diff
@@ -215,6 +215,208 @@ describe('Safe4337Module - Reference EntryPoint', () => {
...
+      safeOp,
+      signature: signature.concat('00'), // adding '00' invalidates signature length
+    })
+
+    await expect(entryPoint.handleOps([userOp], await relayer.getAddress()))
+      .to.be.revertedWithCustomError(entryPoint, 'FailedOp')
+      .withArgs(0, 'AA24 signature error')
+  })
+
+  it('should revert when signature pointer points to invalid part of signature data - Smart contract signature (NOTE: would require a staked paymaster for ERC-4337)', async () => {
```
nlordell, line 304:
```suggestion
  it('should revert when signature pointer points to invalid part of signature data - Smart contract signature', async () => {
```

modules/4337/test/erc4337/ReferenceEntryPoint.spec.ts
```diff
@@ -215,6 +215,208 @@ describe('Safe4337Module - Reference EntryPoint', () => {
...
+      // Replace the 2nd word of static part of signature containing the pointer to dynamic part with invalid pointer value
+      signature: signature.slice(0, 67) + '00'.padStart(64, '0') + signature.slice(131),
+    })
+
+    await expect(entryPoint.handleOps([userOp], await relayer.getAddress()))
+      .to.be.revertedWithCustomError(entryPoint, 'FailedOp')
+      .withArgs(0, 'AA24 signature error')
+  })
+
+  it('should revert when expected offset and given dynamic pointer in signatures are not equal - Smart contract signature (NOTE: would require a staked paymaster for ERC-4337)', async () => {
```
nlordell, line 359:
```suggestion
  it('should revert when padded with additional bytes in-between signatures - Smart contract signature', async () => {
```

modules/4337/test/erc4337/ReferenceEntryPoint.spec.ts
```diff
@@ -215,6 +215,208 @@ describe('Safe4337Module - Reference EntryPoint', () => {
...
+    const userOp = buildPackedUserOperationFromSafeUserOperation({
+      safeOp,
+      // Replace the 2nd word of static part of signature containing the pointer to dynamic part with invalid pointer value
+      signature:
+        signature.slice(0, 66) +
+        ((signature.length - 2) / 2).toString(16).padStart(64, '0') +
+        '00' +
+        signature.slice(132) +
+        signature.slice(132),
+    })
```
nlordell, line 413: My personal opinion: It would be clearer to build the signature manually instead of first building it, then trying to change the signature to "make it work" for the test:

```suggestion
    const parentSafeSignature = user.signTypedData(
      {
        verifyingContract: parentSafe.address,
        chainId: await chainId(),
      },
      {
        SafeMessage: [{ type: 'bytes', name: 'message' }],
      },
      {
        message: opData,
      },
    )
    const signature = ethers.concat([
        ethers.zeroPadValue(parentSafe.address, 32), // address of the signer
        ethers.toHexBe(65 + "padding".length), // offset of the start of the signature
        "0x00", // contract signature type
        ethers.toUtf8Bytes("padding"), // extra illegal padding between signatures
        ethers.toHexBe(ethers.dataLength(parentSafeSignature)), // length of the dynamic signature
        parentSafeSignature,
    ])
```

(_disclaimer_: I didn't test the code out, but I found it more self-explanatory than all the slicing, especially since there are a lot of magic numbers there, since it is string slicing on `0x` prefixed hex instead of byte slicing).

modules/4337/test/erc4337/ERC4337ModuleNew.spec.ts
```diff
@@ -119,6 +123,147 @@ describe('Safe4337Module - Newly deployed safe', () => {
       expect(await ethers.provider.getBalance(safe.address)).to.be.eq(ethers.parseEther('0'))
     })
 
+    it('should revert when signature length contains additional bytes - EOA signature', async () => {
```
nlordell, line 126: I _think_ we keep the tests in `ERC4337ModuleExisting.spec` matching the ones here.

modules/4337/test/erc4337/ReferenceEntryPoint.spec.ts
```diff
@@ -215,6 +215,208 @@ describe('Safe4337Module - Reference EntryPoint', () => {
...
+            message: opData,
+          },
+        ),
+        dynamic: true,
+      },
+    ])
+
+    const userOp = buildPackedUserOperationFromSafeUserOperation({
+      safeOp,
+      // Replace the 2nd word of static part of signature containing the pointer to dynamic part with invalid pointer value
```
nlordell, line 406: I think this comment is outdated.

modules/4337/test/erc4337/ReferenceEntryPoint.spec.ts
```diff
@@ -215,6 +215,208 @@ describe('Safe4337Module - Reference EntryPoint', () => {
...
+      signature:
+        signature.slice(0, 66) +
+        ((signature.length - 2) / 2).toString(16).padStart(64, '0') +
```
nlordell, line 409: I think this offset is wrong - it points to the start of the padding and not the length. Notably this test *also* fails with `main` branch contracts, and is not covering the case of having a valid signature with extra bytes in between the dynamic signatures.

nlordell: In the current test suite, if we apply this diff:

```diff
diff --git a/modules/4337/contracts/Safe4337Module.sol b/modules/4337/contracts/Safe4337Module.sol
index 485ca91..62cea2a 100644
--- a/modules/4337/contracts/Safe4337Module.sol
+++ b/modules/4337/contracts/Safe4337Module.sol
@@ -246,7 +246,7 @@ contract Safe4337Module is IAccount, HandlerContext, CompatibilityFallbackHandle
             }
             /* solhint-enable no-inline-assembly */
             // If the signature data pointer is not pointing to the expected location, return false.
-            if (!pointsAtEnd) return false;
+            //if (!pointsAtEnd) return false;
         }
         isValid = signatures.length <= expectedOffset;
     }

```

The tests still pass - so we aren't covering this path fully.

akshay-ap: https://github.com/safe-global/safe-modules/pull/453/commits/42b216e6f756cb43fe396f1947367fe0898a9584
The test passes because ` isValid = signatures.length <= expectedOffset;` check prevents it. I also added signature  length manipulation check which justifies existence of `if (!pointsAtEnd) return false;`.

Still commenting the statement will lead to test to pass because Safe reverts while checking signatures. If we want failing test on removing `if (!pointsAtEnd) return false;`  then we need to use a Mock contract that passes signature checks even on invalid input.

modules/4337/test/erc4337/ReferenceEntryPoint.spec.ts
akshay-ap: marking as resolved

https://github.com/safe-global/safe-modules/pull/453/commits/42b216e6f756cb43fe396f1947367fe0898a9584

https://github.com/safe-global/safe-modules/pull/453/commits/0306c399fa9cab12bb70279e56179cacc2cdaa26

modules/4337/test/erc4337/ERC4337ModuleNew.spec.ts
nlordell: Not worth it.

nlordell (approved):

PR #455, opened 2024-07-03
fix-correct-example-address

vseehausen:
This PR fixes the directory in the README of the example 4337-passkeys. 

It also fixes two minor typos.

### DIFF
--- a/examples/4337-passkeys/README.md
+++ b/examples/4337-passkeys/README.md
@@ -8,7 +8,7 @@ This minimalistic example application demonstrates a Safe{Core} Smart Account de
 
 ```bash
 git clone https://github.com/safe-global/safe-modules.git
-cd safe-modules
+cd safe-modules/examples/4337-passkeys
 ```
 
 ### Install dependencies
--- a/examples/4337-passkeys/src/main.tsx
+++ b/examples/4337-passkeys/src/main.tsx
@@ -2,7 +2,7 @@ import React from 'react'
 import ReactDOM from 'react-dom/client'
 import { createBrowserRouter, RouterProvider } from 'react-router-dom'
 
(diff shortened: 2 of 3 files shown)


examples/4337-passkeys/src/main.tsx
```diff
@@ -2,7 +2,7 @@ import React from 'react'
 import ReactDOM from 'react-dom/client'
 import { createBrowserRouter, RouterProvider } from 'react-router-dom'
 
-// Keep the import of the wallets file to eval so w3m package can initialise
+// Keep the import of the wallets file to eval so w3m package can initialize
```
mmv08, line 5: I thought we default to the UK English 😂

examples/4337-passkeys/README.md
```diff
@@ -8,7 +8,7 @@ This minimalistic example application demonstrates a Safe{Core} Smart Account de
 
 ```bash
 git clone https://github.com/safe-global/safe-modules.git
-cd safe-modules
+cd safe-modules/examples/4337-passkeys
```
mmv08, line 11: One must run the `pnpm install` (the next step in the readme) in the project's root directory. Running it in the package directory may work as well, but I'm not sure if pnpm handles nested directories correctly

mmv08: I think it should work: https://github.com/orgs/pnpm/discussions/4735#discussioncomment-2770033 but please test it :pray:

nlordell: If we change the `cd` directory here, I think we should update other commands to also not assume that they are run from the repository root. For example this one:

https://github.com/safe-global/safe-modules/blob/8b85c79b52454606a11b5857018af797ee8f1c3d/examples/4337-passkeys/README.md?plain=1#L35-L37

PR #456, opened 2024-07-04
Emit Event on Shared Signer Configuration

### LINKED ISSUE #3: Does not emit event after writing into storage
**Github username:** --
**Twitter username:** --
**Submission hash (on-chain):** 0x44a5b7162971bec243ca1c07e1f22313e3e44ec1ed7e0660f4c97bdbab45ad1b
**Severity:** low

**Description:**
Description

`SafeWebAuthnSharedSigner::configue` sets Sets the signer configuration for the calling account but the issue it doesn't emit evet after writing into storage.

```solidity

    function configure(Signer memory signer) external onlyDelegateCall {
        uint256 signerSlot = SIGNER_SLOT;
        Signer storage signerStorage;

        // solhint-disable-next-line no-inline-assembly
        assembly ("memory-safe") {
            signerStorage.slot := signerSlot
        }

        signerStorage.x = signer.x;
        signerStorage.y = signer.y;
        signerStorage.verifiers = signer.verifiers;
    }

```
emit the event after writing into storage

nlordell:
Fixes https://github.com/hats-finance/Safe-0x2909fdefd24a1ced675cb1444918fa766d76bdac/issues/3

This PR adds an event that is emitted on shared signer configuration. Note that the event is emitted in the context of the configured Safe account. In order to prevent event `topic0` collisions, we chose an explicitly verbose name.

With respect to indexing, we explicitly chose **not** to index any of the fields so that it matches the `Created` event:

https://github.com/safe-global/safe-modules/blob/bb5e1f7360633d5ecfb175c5d8667b2a3a1b746d/modules/passkey/contracts/interfaces/ISafeSignerFactory.sol#L13-L20

To re-iterate on the reasons we chose to not index some of the fields (same reasoning applies to `Created` event above):
- `x` and `y`: These are single-use public key coordinates attached to a passkey. It doesn't make sense to filter events by "all events with `x` value", or similarly "all events with `y` value". Furthermore, you can compute the `signer` (which is indexed in the `Created` event) from the key coordinates, so you can to some degree search for signers for a given key coordinate.
- `verifiers`: This felt more like a configuration than something that Dapps would need to filter by. Why it might be interesting to see on chain which verifiers are being used, tools like Dune can help with this, and the additional cost of an indexed field is not worth it in our opinion.

### DIFF
--- a/modules/passkey/contracts/4337/SafeWebAuthnSharedSigner.sol
+++ b/modules/passkey/contracts/4337/SafeWebAuthnSharedSigner.sol
@@ -33,11 +33,6 @@ contract SafeWebAuthnSharedSigner is SignatureValidator {
     uint256 private constant _SIGNER_MAPPING_SLOT = 0x2e0aed53485dc2290ceb5ce14725558ad3e3a09d38c69042410ad15c2b4ea4e8;
 
     /**
-     * @notice An error indicating a `CALL` to a function that should only be `DELEGATECALL`-ed.
-     */
-    error NotDelegateCalled();
-
-    /**
      * @notice Address of the shared signer contract itself.
      * @dev This is used for determining whether or not the contract is being `DELEGATECALL`-ed when
      * setting signer data.
@@ -48,6 +43,23 @@ contract SafeWebAuthnSharedSigner is SignatureValidator {
      * @notice The starting storage slot on the account containing the signer data.
      */
     uint256 public immutable SIGNER_SLOT;
+
+    /**
+     * @notice Emitted when the shared signer is configured for an account.
+     * @dev Note that the configured account is not included in the event data. Since configuration
+     * is done as a `DELEGATECALL`, the contract emitting the event is the configured account. This
+     * is also why the event name is prefixed with `SafeWebAuthnSharedSigner`, in order to avoid
+     * event `topic0` collisions with other contracts (seeing as "configured" is a common term).
+     * @param x The x-coordinate of the public key.
+     * @param y The y-coordinate of the public key.
+     * @param verifiers The P-256 verifiers to use.
+     */
+    event SafeWebAuthnSharedSignerConfigured(uint256 x, uint256 y, P256.Verifiers verifiers);
+
+    /**
+     * @notice An error indicating a `CALL` to a function that should only be `DELEGATECALL`-ed.
+     */
+    error NotDelegateCalled();
 
     /**
      * @notice Create a new shared WebAuthn signer instance.
@@ -133,6 +145,8 @@ contract SafeWebAuthnSharedSigner is SignatureValidator {
         signerStorage.x = signer.x;
         signerStorage.y = signer.y;
         signerStorage.verifiers = signer.verifiers;
+
+        emit SafeWebAuthnSharedSignerConfigured(signer.x, signer.y, signer.verifiers);
     }
 
     /**
--- a/modules/passkey/test/4337/SafeWebAuthnSharedSigner.spec.ts
+++ b/modules/passkey/test/4337/SafeWebAuthnSharedSigner.spec.ts
@@ -170,6 +170,32 @@ describe('SafeWebAuthnSharedSigner', () => {
       expect(await ethers.provider.getStorage(account, BigInt(signerSlot) + 2n)).to.equal(config.verifiers)
     })
 
+    it('Should emit an event', async () => {
+      const { safeSingleton, proxyFactory, sharedSigner, mockVerifier } = await setupTests()
+
+      const config = {
+        x: ethers.id('publicKey.x'),
+        y: ethers.id('publicKey.y'),
+        verifiers: ethers.toBeHex(await mockVerifier.getAddress(), 32),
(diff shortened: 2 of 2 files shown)


modules/passkey/contracts/4337/SafeWebAuthnSharedSigner.sol
```diff
@@ -49,6 +44,23 @@ contract SafeWebAuthnSharedSigner is SignatureValidator {
...
+     * @notice Emitted when the shared signer is configured for an account.
+     * @dev Note that the configured account is not included in the event data. Since configuration
+     * is done as a `DELEGATECALL`, the contract emitting the event is the configured account. This
+     * is also why the event name is prefixed with `SafeWebAuthnSharedSigner`, in order to avoid
+     * event `topic0` collisions with other contracts (seeing as "configured" is a common term).
+     * @param x The x-coordinate of the public key.
+     * @param y The y-coordinate of the public key.
+     * @param verifiers The P-256 verifiers to use.
+     */
+    event SafeWebAuthnSharedSignerConfigured(uint256 x, uint256 y, P256.Verifiers verifiers);
```
mmv08, line 57: My first question when I saw this was: do we need them to be indexed or not? I recall we discussed and agreed that it shouldn't be, but perhaps it's worth a comment?

nlordell: Documented in the PR description.

I didn't include it in a comment to match what we did for `Created` event in the other contract. If you think this _should_ be comments, then I will do it in a separate PR and add comments for both events.

mmv08: I'm sorry if we've already talked about this, but let's say that given x and y coordinates, I want to find a safe where this passkey is configured as a signer. Would that be possible without indexed events?

IIRC, we concluded that this was possible without indexing, but I forgot how :see_no_evil:

nlordell: We said it was possible based on address for the proxy signer (you assume some standard `verifiers` configuration that you use for your Dapp and search for the `Created` event by `signer` field), but I don't think it is possible for this event at the moment.

One issue you would have as well, is that the event isn't emitted by a single contract but by the Safe contract itself (because of the `DELEGATECALL`), so you would have to search by topic0 and topic1 for all contracts (don't think that is an issue - just worth pointing out).

If we think "index Safes with this WebAuthn public key" is a usecase, then we should add a "public key hash" indexed field (that is computed as `keccak256(abi.encode(x, y))` or something similar) to make this easier without requiring two indexed fields.

mmv08: > We said it was possible based on address for the proxy signer (you assume some standard verifiers configuration that you use for your Dapp and search for the Created event by signer field)

But can anyone emit the `Created` event? I can directly create a signer without creating a Safe, can't I? So it's rather a non-specific one.

> so you would have to search by topic0 and topic1 for all contracts (don't think that is an issue - just worth pointing out).

Ah, this is a significant constraint, IMO. My use case was pretty much compatibility with autonomous Safe interfaces that don't require a backend, but it doesn't seem that adding this event will make it particularly friendly for this use case.

mmv08 (approved):

modules/passkey/contracts/4337/SafeWebAuthnSharedSigner.sol
mmv08: I'm a little lost. This is about the shared signer contract, right? So, no signer will actually be created, so you cannot use `AddOwner`. Do we have an alternative option then?

Regarding the validity of the use case, I do think it's very useful for trustless interfaces. We can ask the broader team for an opinion

akshay-ap (approved):

modules/passkey/contracts/4337/SafeWebAuthnSharedSigner.sol
nlordell: Sorry, let me try to clarify - the `Created` event is only emitted by the factory contract. The first group steps above are for "I want to find a safe where this passkey is configured as a signer" use case for the **factory** contract, and not the shared signer:

> I think we would then fetch AddOwner events by the signer that we derive the address for.

The second group of steps are for the same use case, but this time with the shared signer:

> For the shared signer, you could do with a index on `heccak256(abi.encode(x, y))`:

Sorry for the confusion - I just imagined that the "I want to find a safe where this passkey is configured as a signer" needs to be achievable from **both** versions of the signers (both factory+proxy and shared), so I highlighted steps from each of them.

nlordell: OK - I'm going to merge this PR as is, and address this comment in a separate PR (adding a user story to show this intended use case).

nlordell: https://github.com/safe-global/safe-modules/pull/463

PR #464, opened 2024-07-15
Signature Length Check FV

### LINKED ISSUE #461: Formally Verify Integrity of Signature Length Check
#453 Introduces new restrictions on the signature length for ERC-4337 user operation signatures in order to prevent malicious bundlers from griefing Safe owners into paying higher validation gas.

This issue captures the work to formally verify the integrity of the signature length check.

### Expected Outcome

- A CVL rule that verifies the integrity of the signature length check

remedcu:
Closes #461 

This PR adds a rule which verifies that the signature length check is ensured using the `_checkSignaturesLength(...)` in the `Safe4337Module`.

The rule ensures that a signature that the bundler could manipulate, which could clear the check in Safe using `checkSignatures(...)` will still get caught using the `_checkSignaturesLength(...)`. For this, we use a `canonicalSignatureHash(...)` which generates the same hash for valid signatures, with and without the excess data. An example is added to show this case before the function is written.

Also have added the workflow for checking the same in CI (Currently, it uses a script, which will be modified in a later PR to be consistent with our other repos).


modules/4337/certora/harnesses/Account.sol
```diff
@@ -53,6 +53,95 @@ contract Account is Safe {
...
+    "0000000000000000000000000000000000000000000000000000000000000024 deadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeef"  // length of bytes + data of bytes of signature 2
+    "0000000000000000000000000000000000000000000000000000000000000003 0000000000000000000000000000000000000000000000000000000000efbeef"          // length of bytes + data of bytes of signature 3
+
+    0x000000000000000000000000000000000000000000000000000000000000000100000000000000000000000000000000000000000000000000000000000000c30000000000000000000000000000000000000000000000000000000000000000020000000000000000000000000000000000000000000000000000000000000103000000000000000000000000000000000000000000000000000000000000000003000000000000000000000000000000000000000000000000000000000000014700000000000000000000000000000000000000000000000000000000000000000400000000000000000000000000000000000000000000000000000000deadbeef0000000000000000000000000000000000000000000000000000000000000024deadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeef00000000000000000000000000000000000000000000000000000000000000030000000000000000000000000000000000000000000000000000000000efbeef
```
nlordell, line 67: I think this encoding is wrong. In particular, the `deadbeef` and `efbeef` values are incorrectly aligned.

modules/4337/certora/specs/SignatureLengthCheck.spec
```diff
@@ -0,0 +1,27 @@
...
+    // Safe Contract functions
+    function safeContract.containsContractSignature(bytes, uint256) external returns(bool) envfree;
+    function safeContract.canonicalSignatureHash(bytes, uint256) external returns(bytes32) envfree;
+}
+
+// This rule verifies that if excess data is added to the dynamic part of the signature, then the signature check will fail.
+rule signatureLengthCheckDirectly(bytes signatures, bytes gasGriefingData, uint256 threshold) {
+    require signatures.length > 0;
+    require gasGriefingData.length > 0;
+    assert checkSignaturesLength(signatures, threshold) => !checkSignaturesLength(combineBytes(signatures,gasGriefingData), threshold);
```
nlordell, line 16:
```suggestion
    assert checkSignaturesLength(signatures, threshold) => !checkSignaturesLength(combineBytes(signatures, gasGriefingData), threshold);
```

modules/4337/certora/harnesses/Account.sol
remedcu: Did you mean, instead of this:

```
    "0000000000000000000000000000000000000000000000000000000000000004 00000000000000000000000000000000000000000000000000000000deadbeef"          // length of bytes + data of bytes of signature 1
    "0000000000000000000000000000000000000000000000000000000000000024 deadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeef"  // length of bytes + data of bytes of signature 2
    "0000000000000000000000000000000000000000000000000000000000000003 0000000000000000000000000000000000000000000000000000000000efbeef"          // length of bytes + data of bytes of signature 3
```

It will be like this?

```
    "0000000000000000000000000000000000000000000000000000000000000004 deadbeef"                                                                  // length of bytes + data of bytes of signature 1
    "0000000000000000000000000000000000000000000000000000000000000024 deadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeef"  // length of bytes + data of bytes of signature 2
    "0000000000000000000000000000000000000000000000000000000000000003 efbeef"                                                                    // length of bytes + data of bytes of signature 3
```

modules/4337/certora/specs/SignatureLengthCheck.spec
```diff
@@ -0,0 +1,27 @@
...
+}
+
+// This rule verifies that if excess data is added to the dynamic part of the signature, though it could pass in the safe contract's
+// `checkSignatures(...)`, it will be caught within the `_checkSignaturesLength(...)` call, as the dynamic length is checked.
+rule canonicalHashBasedLengthCheck(bytes signatures, bytes griefedSignatures, uint256 threshold) {
+    require safeContract.canonicalSignatureHash(signatures, threshold) == safeContract.canonicalSignatureHash(griefedSignatures, threshold);
+    require signatures.length < griefedSignatures.length;
+    require safeContract.containsContractSignature(signatures, threshold);
```
nlordell, line 24: This shouldn't be needed. What counter-example was it finding?

In particular, if you append extra bytes to EOA signatures, you would also get a griefing attack that this function should protect from.

remedcu: You are right, we can avoid this check. Updated 👍🏾

modules/4337/certora/specs/SignatureLengthCheck.spec
remedcu: Updated 👍🏾

modules/4337/certora/conf/SignatureLengthCheck.conf
```diff
@@ -0,0 +1,15 @@
+{
+    "files": [
+        "certora/harnesses/Safe4337ModuleHarness.sol",
+        "certora/harnesses/Account.sol",
+    ],
+	"optimistic_loop": true,
```
akshay-ap, line 6: Have you tried with different values of `loop_iter`? Default is 1.
Would be nice to use value 2 or 3 if possible.
https://docs.certora.com/en/latest/docs/prover/cli/options.html#loop-iter

modules/4337/certora/harnesses/Account.sol
akshay-ap: Contract signatures are not padded. so,
`0000000000000000000000000000000000000000000000000000000000000003 efbeef`

modules/4337/certora/specs/SignatureLengthCheck.spec
```diff
@@ -0,0 +1,25 @@
...
+
+methods {
+    function checkSignaturesLength(bytes, uint256) external returns(bool) envfree;
+    function combineBytes(bytes, bytes) external returns(bytes) envfree;
+
+    // Safe Contract functions
+    function safeContract.canonicalSignatureHash(bytes, uint256) external returns(bytes32) envfree;
+}
+
+// This rule verifies that if excess data is added to the dynamic part of the signature, then the signature check will fail.
```
akshay-ap, line 11:
```suggestion
// This rule verifies that if excess data is added to the the signature, then the signature check will fail.
```

akshay-ap: The rule doesn't seem to be specific to contract signature.

modules/4337/certora/specs/SignatureLengthCheck.spec
```diff
@@ -0,0 +1,25 @@
...
+rule signatureLengthCheckDirectly(bytes signatures, bytes gasGriefingData, uint256 threshold) {
+    require signatures.length > 0;
+    require gasGriefingData.length > 0;
+    assert checkSignaturesLength(signatures, threshold) => !checkSignaturesLength(combineBytes(signatures, gasGriefingData), threshold);
+}
+
+// This rule verifies that if excess data is added to the dynamic part of the signature, though it could pass in the safe contract's
```
akshay-ap, line 18: Same here as above comment. `_checkSignaturesLength` also checks lengths of EOA signatures and not just contract signatures.

modules/4337/certora/harnesses/Account.sol
remedcu: Thanks! Updated!

modules/4337/certora/conf/SignatureLengthCheck.conf
remedcu: Updated, and helped to figure out a counter-example with the first rule (though curious why it worked with just one iteration, still investigating), so had to remove it, as we have a stronger rule with the canonical hash one.

modules/4337/certora/specs/SignatureLengthCheck.spec
remedcu: Updated, the initial intention to write like this was because the excess data can only be added to the dynamic part and not to the static part.

modules/4337/certora/specs/SignatureLengthCheck.spec
remedcu: Updated 👍🏾

modules/4337/certora/harnesses/Account.sol
```diff
@@ -53,6 +53,78 @@ contract Account is Safe {
     function getValidUntilTimestamp(bytes calldata sigs) external pure returns (uint48) {
         return uint48(bytes6(sigs[6:12]));
     }
+
+    /*
+    Actual Signature:
+    "0x" +
+    "0000000000000000000000000000000000000000000000000000000000000001 00000000000000000000000000000000000000000000000000000000000000c3 00" +     // encoded EIP-1271 signature 1
```
akshay-ap, line 60: 1. I don't understand the purpose of having all different signature types here with actual values. I think just explaining in short why all of the signatures will have same hash and then linking to docs https://docs.safe.global/advanced/smart-account-signatures should be sufficient.

2. we are writing solidity code to verify some other solidity code. I would be careful in doing so as we are not verifying this solidity code here and preferably avoid writing additional code.  I understand that this code is needed to verify other code but, we are then relying on correctness of this code. We have done similar things in previous FV setups by having harnessed contracts but it was limited to reading contract state and simpler code snippets. Is there any alternative approach?

akshay-ap: > Ironically, the doc has some wrong values. Made a PR to update that: https://github.com/safe-global/safe-docs/pull/527 Maybe we can move the details to a doc and add that link here? Wdyt?

The docs would be fixed soon, right? I think linking to official docs would be the nicest way to do it rather than having another doc somewhere else with similar repeated info.

remedcu: But the docs don't show how the signature could be manipulated (which is good 😅 ) and still make the signature check in Safe pass directly, though will be caught within the 4337 Module.

modules/4337/certora/specs/SignatureLengthCheck.spec
```diff
@@ -0,0 +1,18 @@
+using Account as safeContract;
+
+methods {
+    function checkSignaturesLength(bytes, uint256) external returns(bool) envfree;
+    function combineBytes(bytes, bytes) external returns(bytes) envfree;
```
nlordell, line 5:
```suggestion
```

modules/4337/certora/harnesses/Safe4337ModuleHarness.sol
```diff
@@ -0,0 +1,15 @@
...
+contract Safe4337ModuleHarness is Safe4337Module {
+    constructor(address entryPoint) Safe4337Module(entryPoint) {}
+
+    function checkSignaturesLength(bytes calldata signatures, uint256 threshold) external pure returns (bool) {
+        return _checkSignaturesLength(signatures, threshold);
+    }
+
+    function combineBytes(bytes calldata signatures, bytes calldata data) external pure returns (bytes memory) {
+        return abi.encode(signatures, data);
+    }
```
nlordell, line 14: No longer used

```suggestion
```

modules/4337/certora/harnesses/Account.sol
nlordell: I agree, I find this comment a bit excessive TBH.

> we are writing solidity code to verify some other solidity code

Agree - this is a weakness to using a harness function. I think the issue is that expressing this in CVL is quite hard.

modules/4337/certora/specs/SignatureLengthCheck.spec
```diff
@@ -0,0 +1,18 @@
...
+    function safeContract.canonicalSignatureHash(bytes, uint256) external returns(bytes32) envfree;
+}
+
+// This rule verifies that if excess data is added to the signature, though it could pass in the safe contract's `checkSignatures(...)`,
+// it will be caught within the `_checkSignaturesLength(...)` call, as the dynamic length is checked.
+rule canonicalHashBasedLengthCheck(bytes signatures, bytes griefedSignatures, uint256 threshold) {
+    require safeContract.canonicalSignatureHash(signatures, threshold) == safeContract.canonicalSignatureHash(griefedSignatures, threshold);
+    require signatures.length < griefedSignatures.length;
+
+    assert checkSignaturesLength(signatures, threshold) => !checkSignaturesLength(griefedSignatures, threshold);
```
nlordell, line 17: I'm surprised this doesn't find a counter example, as I think I can come up with one.

modules/4337/certora/harnesses/Account.sol
remedcu: > I agree, I find this comment a bit excessive TBH.

Do you recommend removing it altogether, or moving it? Or just write the same in fewer words rather than the example?

modules/4337/certora/specs/SignatureLengthCheck.spec
remedcu: Updated 👍🏾

modules/4337/certora/harnesses/Account.sol
remedcu: I have written a shorter comment with a smaller example and a few lines of explanation.

modules/4337/certora/harnesses/Account.sol
```diff
@@ -53,6 +53,60 @@ contract Account is Safe {
...
+    "0x" +
+    "0000000000000000000000000000000000000000000000000000000000000001 0000000000000000000000000000000000000000000000000000000000000041 00" + // encoded EIP-1271 signature 1
+    "0000000000000000000000000000000000000000000000000000000000000004 deadbeef"                                                              // length of bytes + data of bytes of signature 1
+
+    0x00000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000041000000000000000000000000000000000000000000000000000000000000000004deadbeef
+
+    With Excess Data:
+    "0x" +
+    "0000000000000000000000000000000000000000000000000000000000000001 0000000000000000000000000000000000000000000000000000000000000041 00" + // encoded EIP-1271 signature 1
+    "0000000000000000000000000000000000000000000000000000000000000004 deadbeefdeadbeef"                                                      // length of bytes + data of bytes of signature 1 + excess data
```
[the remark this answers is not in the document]
remedcu: Updated 👍🏾

modules/4337/certora/conf/SignatureLengthCheck.conf
```diff
@@ -0,0 +1,16 @@
+{
+    "files": [
+        "certora/harnesses/Safe4337ModuleHarness.sol",
+        "certora/harnesses/Account.sol",
+    ],
+    "loop_iter": "3",
+	"optimistic_loop": true,
```
nlordell, line 7: nit: formatting:
```suggestion
    "optimistic_loop": true,
```

modules/4337/certora/harnesses/Account.sol
```diff
@@ -53,6 +53,59 @@ contract Account is Safe {
...
+    "0000000000000000000000000000000000000000000000000000000000000001 0000000000000000000000000000000000000000000000000000000000000041 00" + // encoded EIP-1271 signature 1
+    "0000000000000000000000000000000000000000000000000000000000000004 deadbeef"                                                              // length of bytes + data of bytes of signature 1
+
+    With Excess Data:
+    "0x" +
+    "0000000000000000000000000000000000000000000000000000000000000001 0000000000000000000000000000000000000000000000000000000000000041 00" + // encoded EIP-1271 signature 1
+    "0000000000000000000000000000000000000000000000000000000000000004 deadbeef"                                                              // length of bytes + data of bytes of signature 1
+    "ffffff"                                                                                                                                 // excess data
+
+    Both will have the same canonical hash: 0xc860b1a81652620308a8138a17ef5105d9b18e6e766516ffd3de9897260d1320
```
nlordell, line 72: We can also link to the Safe documentation pages (since this won't be on-chain anyway).

modules/4337/certora/conf/SignatureLengthCheck.conf
remedcu: Updated 👍🏾

modules/4337/certora/harnesses/Account.sol
remedcu: Added 👍🏾

modules/4337/certora/specs/SignatureLengthCheck.spec
remedcu: We `require` in the canonical hash function for the signature length to be > 0

https://github.com/safe-global/safe-modules/blob/f72107828db4c06d3a81ee1563dca721a54e845a/modules/4337/certora/harnesses/Account.sol#L92

So, the above example is not valid.

nlordell: Why? Safe contract signatures don't require this:

https://github.com/safe-global/safe-smart-account/blob/8f80a8372d193be121dcdb52e869a258824e5c0f/contracts/Safe.sol#L231

The exact same counter example can also exist for signatures of any length - the "trick" is that two signers have identical signatures:

```
"0000000000000000000000000000000000000000000000000000000000000001" + (65*2).toString(16).padStart(64, "0") + "00" +
"0000000000000000000000000000000000000000000000000000000000000002" + (65*2).toString(16).padStart(64, "0") + "00" +
"0000000000000000000000000000000000000000000000000000000000000001" + "ff"
```

and

```
"0000000000000000000000000000000000000000000000000000000000000001" + (65*2).toString(16).padStart(64, "0") + "00" +
"0000000000000000000000000000000000000000000000000000000000000002" + (65*2 + 32).toString(16).padStart(64, "0") + "00" +
"0000000000000000000000000000000000000000000000000000000000000001" + "ff" +
"0000000000000000000000000000000000000000000000000000000000000001" + "ff"
```

remedcu: Updated, thanks for the pair programming session @nlordell!

nlordell (approved):

PR #465, opened 2024-07-18
Use Slightly More Gas For 4337 Test

nlordell:
Under certain conditions, an E2E test can fail because it uses slightly more gas. This PR adds a larger gas buffer to the E2E test to prevent intermittent failures.

Example failure: https://github.com/safe-global/safe-modules/actions/runs/9990468990/job/27611219692

```
  1) Singleton Signers [@4337]
       should deploy a new Safe with alternate signing scheme accessing associated storage:
     Error: could not coalesce error (error={ "code": -32500, "message": "verificationGas should have extra 2000 gas. has only 353" }, payload={ ... }, code=UNKNOWN_ERROR, version=6.13.1)
```


modules/4337/certora/harnesses/Account.sol
```diff
@@ -59,7 +59,7 @@ contract Account is Safe {
      * The canonical format ensures the signatures are tightly packed one after the other in order.
      *
      * For more details on signature encoding: https://docs.safe.global/advanced/smart-account-signatures
-    */
+     */
```
nlordell, line 62: Unrelated formatting changes from running `prettier` on the project.

remedcu (approved):

PR #470, opened 2024-07-22
Certora Script Update & Minor changes

### LINKED ISSUE #468: Safe Modules Certora Setup Refactor

remedcu:
Fixes #468 

This PR eliminates the use of a script for the Certora FV and uses the Certora CLI command directly.

In addition, extra small changes included in this PR:
- Natspec changes in Harness solidity code.
- Inclusion of Recovery Module in the README.
- Using the latest version of checkout (`v4`) for GitHub CI.


modules/4337/certora/requirements.txt
```diff
@@ -0,0 +1 @@
+certora-cli==7.3.0
```
mmv08, line 1: can we switch to latest https://pypi.org/project/certora-cli/ ? would be a good thing to do together with updating the scripts

mmv08 (approved):

modules/4337/certora/requirements.txt
remedcu: As mentioned in today's sync, I will create a new issue for the upgrade as this requires some changes to the specs which I want to discuss with Certora first.

PR #471, opened 2024-07-22
Add comment in _checkSignaturesLength regarding dynamic signatures

akshay-ap:
This PR adds a comment on dynamic signatures and signature length manipulation in `_checkSignaturesLength` Natspec

### DIFF
--- a/modules/4337/contracts/Safe4337Module.sol
+++ b/modules/4337/contracts/Safe4337Module.sol
@@ -223,9 +223,9 @@ contract Safe4337Module is IAccount, HandlerContext, CompatibilityFallbackHandle
      * `_checkSignaturesLength` function checks for the presence of any padded bytes to the `signature` data.
      * However, there is an edge case that `_checkSignaturesLength` function cannot detect.
      * A malicious bundler can manipulate the field(s) storing the signature length and pad additional bytes to the
-     * dynamic part of the signatures which will make `_checkSignaturesLength` to return true. In such cases, it is the
-     * responsibility of the Safe signature validator (which can be a Safe or any other contract) that supports
-     * ERC-1271 and is the owner of the Safe to check for additional bytes to it smart contract signature data.
+     * dynamic part of the signatures which will make `_checkSignaturesLength` to return true.  In such cases, it is
+     * the responsibility of the Safe signature validator implementation, as an account owner, to check for additional
+     * bytes to it smart contract signature data.
      * @param signatures Signatures data.
      * @param threshold Signer threshold for the Safe account.
      * @return isValid True if length check passes, false otherwise.
--- a/modules/4337/contracts/Safe4337Module.sol
+++ b/modules/4337/contracts/Safe4337Module.sol
@@ -217,15 +217,16 @@ contract Safe4337Module is IAccount, HandlerContext, CompatibilityFallbackHandle
      * of {checkSignatures}.
      * @dev Safe account has two types of signatures: EOA and Smart Contract signatures. While the EOA signature is 
      * fixed in size, the Smart Contract signature can be of arbitrary length. Safe encodes the Smart Contract
-     * signature length in the signature data. A malicious bundler can pad additional bytes to the `signature` data,
-     * causing the account to pay more gas than needed for user operation validation and reach the
-     * verificationGasLimit if the verifier does not implement appropriate length checks. `_checkSignaturesLength`
-     * function checks for presence of any padded bytes to the `signature` data. However, there is an
-     * edge case that `_checkSignaturesLength` function cannot detect. Since, the `signature` field in UserOp is not
-     * part of the UserOp hash a malicious bundler can manipulate the field storing the signature length and pad
-     * additional bytes to the dynamic part of the signatures which will make `_checkSignaturesLength` to return true.
-     * In such cases, it is the responsibility of the signature verifier to check for additional padded bytes to the
-     * signatures data.
+     * signature length in the signature data. If appropriate length checks are not performed during the signature
+     * verification then a malicious bundler can pad additional bytes to the signatures data and make the account pay
+     * more gas than needed for user operation validation and reach the verificationGasLimit.
+     * `_checkSignaturesLength` function checks for the presence of any padded bytes to the `signature` data.
+     * However, there is an edge case that `_checkSignaturesLength` function cannot detect.
+     * Since the `signature` field in UserOp is not part of the UserOp hash a malicious bundler can manipulate the
+     * field(s) storing the signature length and pad additional bytes to the dynamic part of the signatures which will
+     * make `_checkSignaturesLength` to return true. In such cases, it is the responsibility of the signature verifier
+     * (which can be a Safe or any other contract) that supports ERC-1271 and is the owner of the Safe to check for
+     * additional padded bytes to the signatures data.
      * @param signatures Signatures data.
      * @param threshold Signer threshold for the Safe account.
      * @return isValid True if length check passes, false otherwise.
--- a/modules/4337/contracts/Safe4337Module.sol
+++ b/modules/4337/contracts/Safe4337Module.sol
@@ -222,11 +222,10 @@ contract Safe4337Module is IAccount, HandlerContext, CompatibilityFallbackHandle
      * more gas than needed for user operation validation and reach the `verificationGasLimit`.
      * `_checkSignaturesLength` function checks for the presence of any padded bytes to the `signature` data.
      * However, there is an edge case that `_checkSignaturesLength` function cannot detect.
-     * Since the `signature` field in UserOp is not part of the UserOp hash a malicious bundler can manipulate the
-     * field(s) storing the signature length and pad additional bytes to the dynamic part of the signatures which will
-     * make `_checkSignaturesLength` to return true. In such cases, it is the responsibility of the Safe signature validator
-     * (which can be a Safe or any other contract) that supports ERC-1271 and is the owner of the Safe to check for
-     * additional bytes to it smart contract signature data.
+     * A malicious bundler can manipulate the field(s) storing the signature length and pad additional bytes to the
+     * dynamic part of the signatures which will make `_checkSignaturesLength` to return true. In such cases, it is the
+     * responsibility of the Safe signature validator (which can be a Safe or any other contract) that supports
+     * ERC-1271 and is the owner of the Safe to check for additional bytes to it smart contract signature data.
      * @param signatures Signatures data.
      * @param threshold Signer threshold for the Safe account.
      * @return isValid True if length check passes, false otherwise.
--- a/modules/4337/contracts/Safe4337Module.sol
+++ b/modules/4337/contracts/Safe4337Module.sol
@@ -219,7 +219,7 @@ contract Safe4337Module is IAccount, HandlerContext, CompatibilityFallbackHandle
      * fixed in size, the Smart Contract signature can be of arbitrary length. Safe encodes the Smart Contract
      * signature length in the signature data. If appropriate length checks are not performed during the signature
      * verification then a malicious bundler can pad additional bytes to the signatures data and make the account pay
-     * more gas than needed for user operation validation and reach the verificationGasLimit.
+     * more gas than needed for user operation validation and reach the `verificationGasLimit`.
      * `_checkSignaturesLength` function checks for the presence of any padded bytes to the `signature` data.
      * However, there is an edge case that `_checkSignaturesLength` function cannot detect.
      * Since the `signature` field in UserOp is not part of the UserOp hash a malicious bundler can manipulate the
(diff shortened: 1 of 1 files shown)


modules/4337/contracts/Safe4337Module.sol
```diff
@@ -212,9 +212,17 @@ contract Safe4337Module is IAccount, HandlerContext, CompatibilityFallbackHandle
...
     /**
-     * @dev Checks if the signatures length is correct and does not contain additional bytes. The function does not
+     * @notice Checks if the signatures length is correct and does not contain additional bytes. The function does not
      * check the integrity of the signature encoding, as this is expected to be checked by the {Safe} implementation
      * of {checkSignatures}.
+     * @dev A malicious bundler can pad additional bytes to the `signatures` data, causing the account to pay more gas
+     * than needed for user operation validation. Safe account has two types of signatures: EOA and Smart Contract
+     * signatures. While the EOA signature is fixed in size, the Smart Contract signature can be of arbitrary length.
+     * Safe encodes the Smart Contract signature length in the signature data. Since, the `signature` field in UserOp
+     * is not part of the UserOp hash a malicious bundler can manipulate the field storing the signature length and pad
```
remedcu, line 222: Doubt: I thought only the signature offset and additional bytes added to the dynamic part could be manipulated. Won't a change in signature length change the actual signature taken by the Safe Contracts for verification, thus resulting in a definite failure of signature verification?

akshay-ap: It depends on the account implementation whether the signature verification will fail or pass. In case of Safe, it will pass.

The test case here will fail because verification passed though there were padded bytes in signature. 
https://github.com/safe-global/safe-modules/commit/c825117819b5ddc3d61a503745b44b441f061282

modules/4337/contracts/Safe4337Module.sol
```diff
@@ -212,9 +212,17 @@ contract Safe4337Module is IAccount, HandlerContext, CompatibilityFallbackHandle
     }
 
     /**
-     * @dev Checks if the signatures length is correct and does not contain additional bytes. The function does not
+     * @notice Checks if the signatures length is correct and does not contain additional bytes. The function does not
      * check the integrity of the signature encoding, as this is expected to be checked by the {Safe} implementation
      * of {checkSignatures}.
+     * @dev A malicious bundler can pad additional bytes to the `signatures` data, causing the account to pay more gas
+     * than needed for user operation validation. Safe account has two types of signatures: EOA and Smart Contract
```
nlordell, line 219: I think this sentence needs to be clarified. To me it reads like it is currently possible (which AFAIU, it isn't in the case for smart contract verifiers that have signature length checks).

akshay-ap: updated https://github.com/safe-global/safe-modules/pull/471/commits/36c0742db90438c850b1cf19403e0fb346418fec

modules/4337/contracts/Safe4337Module.sol
```diff
@@ -212,9 +212,20 @@ contract Safe4337Module is IAccount, HandlerContext, CompatibilityFallbackHandle
     }
 
     /**
-     * @dev Checks if the signatures length is correct and does not contain additional bytes. The function does not
+     * @notice Checks if the signatures length is correct and does not contain additional bytes. The function does not
      * check the integrity of the signature encoding, as this is expected to be checked by the {Safe} implementation
      * of {checkSignatures}.
+     * @dev Safe account has two types of signatures: EOA and Smart Contract signatures. While the EOA signature is 
+     * fixed in size, the Smart Contract signature can be of arbitrary length. Safe encodes the Smart Contract
+     * signature length in the signature data. A malicious bundler can pad additional bytes to the `signature` data,
```
nlordell, line 220:
```suggestion
     * signature length in the signature data. A malicious bundler can pad additional bytes to a smart contract `signature` data,
```

nlordell, line 220: I still find this a bit confusing :see_no_evil: - in particular, we call the group of owner signatures for the Safe `signatures`, but here `signature` (I _think_. At first, I thought this was referring to the data for a single Smart Contract signature, but then realized it wasn't).

In particular, this sentence still reads like it is possible to pad `signatures` bytes, but it isn't.

modules/4337/contracts/Safe4337Module.sol
```diff
@@ -212,9 +212,20 @@ contract Safe4337Module is IAccount, HandlerContext, CompatibilityFallbackHandle
...
+     * causing the account to pay more gas than needed for user operation validation and reach the
+     * verificationGasLimit if the verifier does not implement appropriate length checks. `_checkSignaturesLength`
+     * function checks for presence of any padded bytes to the `signature` data. However, there is an
+     * edge case that `_checkSignaturesLength` function cannot detect. Since, the `signature` field in UserOp is not
+     * part of the UserOp hash a malicious bundler can manipulate the field storing the signature length and pad
+     * additional bytes to the dynamic part of the signatures which will make `_checkSignaturesLength` to return true.
+     * In such cases, it is the responsibility of the signature verifier to check for additional padded bytes to the
+     * signatures data.
```
nlordell, line 228: I personally feel like there are some conflated terms here, I think `signature` is used in places where `signatures` should be used, and it isn't always clear what "signature verifier" means (it looks like it is the Safe in some places but a smart contract owner in others)?

modules/4337/contracts/Safe4337Module.sol
akshay-ap: https://github.com/safe-global/safe-modules/pull/471/commits/a50f197ad3c46ecb36e61b15466ce958bab26167

modules/4337/contracts/Safe4337Module.sol
akshay-ap: https://github.com/safe-global/safe-modules/pull/471/commits/a50f197ad3c46ecb36e61b15466ce958bab26167

modules/4337/contracts/Safe4337Module.sol
```diff
@@ -212,9 +212,21 @@ contract Safe4337Module is IAccount, HandlerContext, CompatibilityFallbackHandle
     }
 
     /**
-     * @dev Checks if the signatures length is correct and does not contain additional bytes. The function does not
+     * @notice Checks if the signatures length is correct and does not contain additional bytes. The function does not
      * check the integrity of the signature encoding, as this is expected to be checked by the {Safe} implementation
      * of {checkSignatures}.
+     * @dev Safe account has two types of signatures: EOA and Smart Contract signatures. While the EOA signature is
+     * fixed in size, the Smart Contract signature can be of arbitrary length. Safe encodes the Smart Contract
+     * signature length in the signature data. If appropriate length checks are not performed during the signature
```
nlordell, line 220: Can you move the first sentence further down to where it is relevant? I feel like this leading statement is completely unrelated to the following sentence and makes the documentation confusing for me :see_no_evil:.

modules/4337/contracts/Safe4337Module.sol
```diff
@@ -212,9 +212,21 @@ contract Safe4337Module is IAccount, HandlerContext, CompatibilityFallbackHandle
...
+     * verification then a malicious bundler can pad additional bytes to the signatures data and make the account pay
+     * more gas than needed for user operation validation and reach the verificationGasLimit.
```
nlordell, line 222:
```suggestion
     * more gas than needed for user operation validation and reach the `verificationGasLimit`.
```

modules/4337/contracts/Safe4337Module.sol
```diff
@@ -212,9 +212,21 @@ contract Safe4337Module is IAccount, HandlerContext, CompatibilityFallbackHandle
...
+     * `_checkSignaturesLength` function checks for the presence of any padded bytes to the `signature` data.
```
nlordell, line 223: This is _technically_ not true (and we could get a bug submission with a counter example).

For example:

```
0000000000000000000000001111111111111111111111111111111111111111    # signer 1
00000000000000000000000000000000000000000000000000000000000000c3    # offset 1&2
00                                                                  # contract signature
0000000000000000000000002222222222222222222222222222222222222222    # signer 2
00000000000000000000000000000000000000000000000000000000000000c3    # offset 1&2
00                                                                  # contract signature
0000000000000000000000003333333333333333333333333333333333333333    # signer 3
00000000000000000000000000000000000000000000000000000000000000e8    # offset 3
00                                                                  # contract signature
0000000000000000000000000000000000000000000000000000000000000001 aa # signature 1 & 2
deadbeef                                                            # padding
0000000000000000000000000000000000000000000000000000000000000001 bb # signature 3
```

This will pass the length check with padding.

I would argue that what this does is:
> `_checkSignaturesLength` ensures that the `signatures` data cannot be longer than the canonical encoding of Safe signatures, thus setting a strict upper bound on how long the `signatures` bytes can be, greatly limiting a malicious bundler's ability to pad signature bytes.

modules/4337/contracts/Safe4337Module.sol
```diff
@@ -212,9 +212,21 @@ contract Safe4337Module is IAccount, HandlerContext, CompatibilityFallbackHandle
...
+     * However, there is an edge case that `_checkSignaturesLength` function cannot detect.
```
nlordell, line 224: Here is where I think the leading sentence is most relevant - the kinds of signatures and specifically the smart contract signature matter for this edge case.

modules/4337/contracts/Safe4337Module.sol
```diff
@@ -212,9 +212,21 @@ contract Safe4337Module is IAccount, HandlerContext, CompatibilityFallbackHandle
...
+     * Since the `signature` field in UserOp is not part of the UserOp hash a malicious bundler can manipulate the
```
nlordell, line 225: > Since the `signature` field in UserOp is not part of the UserOp hash

To me this is somewhat self-evident. If it were, you wouldn't need `_checkSignatureLength` in the first place. I wouldn't include it here (its a bit confusing IMO).

modules/4337/contracts/Safe4337Module.sol
```diff
@@ -212,9 +212,21 @@ contract Safe4337Module is IAccount, HandlerContext, CompatibilityFallbackHandle
...
+     * field(s) storing the signature length and pad additional bytes to the dynamic part of the signatures which will
+     * make `_checkSignaturesLength` to return true. In such cases, it is the responsibility of the signature verifier
```
nlordell, line 227: I believe we call them "signature validators" ([source](https://github.com/safe-global/safe-smart-account/blob/main/contracts/interfaces/ISignatureValidator.sol)).

```suggestion
     * make `_checkSignaturesLength` to return true. In such cases, it is the responsibility of the Safe signature validator
```

modules/4337/contracts/Safe4337Module.sol
```diff
@@ -212,9 +212,21 @@ contract Safe4337Module is IAccount, HandlerContext, CompatibilityFallbackHandle
...
+     * (which can be a Safe or any other contract) that supports ERC-1271 and is the owner of the Safe to check for
```
nlordell, line 228: It doesn't use ERC-1271 but legacy 1271 in Safe <=1.4.1 and 1271 in Safe >=1.5.0. I would simplify this to:

```suggestion
     * make `_checkSignaturesLength` to return true. In such cases, it is the responsibility of the Safe signature validator
     * implementation, as an account owner, to check for
```

modules/4337/contracts/Safe4337Module.sol
```diff
@@ -212,9 +212,21 @@ contract Safe4337Module is IAccount, HandlerContext, CompatibilityFallbackHandle
...
+     * additional padded bytes to the signatures data.
```
nlordell, line 229:
```suggestion
     * additional bytes to it smart contract signature data.
```

Should be singular here no?

modules/4337/contracts/Safe4337Module.sol
akshay-ap: https://github.com/safe-global/safe-modules/pull/471/commits/07f6e216efee71071df28c00b1f1cfb576a99f5d

modules/4337/contracts/Safe4337Module.sol
akshay-ap: https://github.com/safe-global/safe-modules/pull/471/commits/3fad06560a3b9b87cabd95c3af1459e2c39d1cd8

modules/4337/contracts/Safe4337Module.sol
akshay-ap: https://github.com/safe-global/safe-modules/pull/471/commits/3dff65b5656955f1b83d40b27641f72c5d864e97

modules/4337/contracts/Safe4337Module.sol
akshay-ap: https://github.com/safe-global/safe-modules/pull/471/commits/3dff65b5656955f1b83d40b27641f72c5d864e97

modules/4337/contracts/Safe4337Module.sol
akshay-ap: https://github.com/safe-global/safe-modules/pull/471/commits/86b8c96c64bf700c5f3dcab99e3d391342b3fa39

nlordell (approved):

PR #474, opened 2024-07-23
Fix Flaky WebAuthn Shim Test

### LINKED ISSUE #369: Ambigious behaviour with WebAuthn shim
## Description

The test "Should authorise and verify an existing credential" in the `WebAuthn Shim` section fails unexpectedly ([Source](https://github.com/safe-global/safe-modules/actions/runs/8597366623/job/23555916815))

## Steps to reproduce

This fails rarely, as the next run of the job returns `true`.

## Additional context

The problem was found during changes created in PR #356 (though no change in the concerned file was made).

nlordell:
Fixes #369

This PR bumps the version of `@simplewebauthn/server` to include a fix to which was causing flacky tests for our WebAuthn shim. It turns out that the library was not correctly dealing with leading zeros in the signature `r` or `s` values, which would cause the test to periodically fail (since the signature was random in each run - as both the private key and the challenge were randomly generated, meaning that we would occasionally have an `r` or `s` value with a leading `0` byte).

### DIFF
--- a/modules/4337/package.json
+++ b/modules/4337/package.json
@@ -54,7 +54,6 @@
     "@openzeppelin/contracts": "^5.0.2",
     "@safe-global/safe-4337-local-bundler": "workspace:^0.0.0",
     "@safe-global/safe-4337-provider": "workspace:^0.0.0",
-    "@simplewebauthn/server": "^10.0.0",
     "@types/chai": "^4.3.16",
     "@types/mocha": "^10.0.7",
     "@types/node": "^20.14.10",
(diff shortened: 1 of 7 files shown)


modules/4337/package.json
```diff
@@ -54,7 +54,6 @@
     "@openzeppelin/contracts": "^5.0.2",
     "@safe-global/safe-4337-local-bundler": "workspace:^0.0.0",
     "@safe-global/safe-4337-provider": "workspace:^0.0.0",
-    "@simplewebauthn/server": "^10.0.0",
```
nlordell, line 57: unused dependency.

remedcu (approved):

PR #478, opened 2024-08-08
Fix TypeScript Lints

nlordell:
This PR fixes some TS lints from the gas metering example.

I am not sure how to test my changes, should I just run the scripts (cc @remedcu).


remedcu:
Yes, we can verify the script by running it and seeing if it results in the same (or similar results with the gas output).

nlordell:
Tested with Pimlico, and it seems to work: https://dashboard.tenderly.co/tx/base-sepolia/0x8793b68afdd3ef5cf47b693af61f368b80250f24ba94e6203d855a20b116d1be/gas-usage

remedcu (approved):
uber-nit: While testing `erc20-paymaster`, it required more than 1 USDC, thus failing. Should we add in the `README` that if the paymaster validation is failing on `erc20-paymaster` runs, then we should increase the ERC20 token payment slightly?

nlordell:
Captured in https://github.com/safe-global/safe-modules/issues/480. Since it is unrelated to this PR - I won't address it here.

PR #482, opened 2024-08-14
Use Arachnid CREATE2 Deployer For Some Contracts

### LINKED ISSUE #481: Daimo deployment update
The latest version of the `passkey` module uses `safe-singleton-factory` for deployment, but we use the official deployment from [here](https://p256.eth.limo/) which uses the Deterministic method.

This issue captures the work for deploying Daimo with the Deterministic approach while keeping the rest of the deployment as it is (Daimo should not be deployed twice, just once with the Deterministic approach).

More context: https://github.com/safe-global/safe-modules/pull/477#discussion_r1713852370

nlordell:
Fixes #481

Some contracts included in our deployment script are expected to use the Arachnid CREATE deployment proxy instead of the Safe singleton factory. We change our deployment scripts to manually use the Arachnid CREATE2 deployer in those specific cases, so the contracts end up at the expected addresses.

Affected contracts:
- ERC-4337 entry point
- Daimo P-256 verifier

You can verify this works by running `pnpm run deploy hardhat` in the Passkey package and checking the deployment addresses are what you expect them to be:

```
...
deploying "EntryPoint" (tx: 0x58b5726a3381344205508251ec394b4bf74b6f8bc669be0b1c9de8d75ce1f8ef)...: deployed at 0x0000000071727De22E5E9d8BAf0edAc6f37da032 with 3650209 gas
...
deploying "DaimoP256Verifier" (tx: 0xe18ff421fb23f117cdb2e60917f68703459743d0a26a14b79eae2cc01def4a20)...: deployed at 0xc2b78104907F722DABAc4C69f826a522B2754De4 with 814337 gas
...
deploying "SafeWebAuthnSignerFactory" (tx: 0x3e6df468f44301ff0bd64fd2403a42214417e25eb865a3c8cb68c79cc5ce042f)...: deployed at 0x1d31F259eE307358a26dFb23EB365939E8641195 with 1031063 gas
deploying "SafeWebAuthnSharedSigner" (tx: 0xc9d5201f64e0f0d537870bafb8a6445c70970f3a13fdda763eee0cb0519bfbf9)...: deployed at 0x94a4F6affBd8975951142c3999aEAB7ecee555c2 with 690536 gas
```

### DIFF
--- a/modules/passkey/CHANGELOG.md
+++ b/modules/passkey/CHANGELOG.md
@@ -13,11 +13,8 @@ EVM target: Paris
 
 - `SafeWebAuthnSignerFactory` at `0x1d31F259eE307358a26dFb23EB365939E8641195`
 - `SafeWebAuthnSharedSigner` at `0x94a4F6affBd8975951142c3999aEAB7ecee555c2`
+- `DaimoP256Verifier` at `0xc2b78104907F722DABAc4C69f826a522B2754De4` ([source](https://p256.eth.limo/))
 - `FCLP256Verifier` at `0xA86e0054C51E4894D88762a017ECc5E5235f5DBA`
-
-### Official Deployment Address from 3rd party
-
-- `DaimoP256Verifier` at `0xc2b78104907F722DABAc4C69f826a522B2754De4` ([Source](https://p256.eth.limo/))
 
 ## Changes
 
--- a/modules/passkey/src/deploy/verifiers.ts
+++ b/modules/passkey/src/deploy/verifiers.ts
@@ -1,24 +1,43 @@
+import type { HardhatRuntimeEnvironment } from 'hardhat/types'
 import type { DeployFunction } from 'hardhat-deploy/types'
 
 import DaimoP256Verifier from '../vendor/daimo-eth/P256Verifier.json'
 
-const deploy: DeployFunction = async ({ deployments, getNamedAccounts }) => {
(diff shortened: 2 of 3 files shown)


remedcu (commented):
~The `DaimoP256Verifier` address is now using the Arachnid one, but the `Entrypoint` is still using the `safe-singleton-factory` when I checked with `pnpm run deploy hardhat`.~

It was using the old files, rectified after doing `build` using the `pnpm install` in the root directory.

remedcu (approved):
Unrelated: `benchmark` check is failing, but it is unrelated to this PR.

mmv08:
Will we be able to deploy the contract to networks that enforce EIP-155?

nlordell:
This only affects:
- The Daimo P-256 verifier contract (which we will not be able to deploy)
- The EntryPoint contract (which we will also not be able to deploy)

For the Daimo P-256 verifier, we can easily have exceptions for networks that don't have the Arachnid CREATE2 deployer (so it will just end up on a "non-canonical" address). It is also not a "core" Safe contract, so I don't see this as a huge issue.

For the EntryPoint contract, it is a bit more complicated (as the Safe4337Module will also end up on a different address), and I assume that getting bundler compatibility will also be a challenge. I would wait to see how the 4337 ecosystem handles this particular case before deciding what we do.

PR #484, opened 2024-08-15
Fix EIP-7212 Dev Node Docker Build

nlordell:
The docker build for the Optimism version of Geth with 7212 support was failing in CI. It turns out that this is related to a recent release of Go 1.23 which is causing Ethereum Geth and forks (such as the Optimism fork we use) to not build.

The fix it to pin to an older version of Golang (1.22 being the latest version that works). I also, in experimenting, updated the version of Optimism Geth we use to see if it would help - which it didn't, but decided to leave the change in anyhow (it is just a version bump to the source that we use to build and a small change to the patch file to account for the changed code).


remedcu (approved):
Nice catch!

PR #486, opened 2024-08-19
Fix Passkey Verification

nlordell:
In deploying the Passkey v0.2.1 contracts, I noticed that the local verification script wasn't working as well as Etherscan verification of the FCL verifier contract. This PR fixes both of these things and is related to a few issues:

1. The `solc` package version was not correctly specified for each `localVerify` script. It was pulling in a `solc` version based on transient dependencies which doesn't necessarily match what the contracts are compiled with (and thus affects local verification). This was an issue shared by all Solidity packages.
2. Local verification was not correctly building the Solidity compiler input JSON. It turns out that the way we build compiler input would sometimes only pass in the Keccak-256 hash of the source instead of the actual Solidity source - so we add an additional condition in the task to read the source from disk if necessary (this is a `hardhat-deploy` inconsistency).
3. The `etherscan-verify` task from the `hardhat-deploy` package does not work for contracts with Solidity compiler settings overrides such as the FCL P-256 verifier contract (which has different compiler settings from the rest of the contracts). We have a special manual `hardhat-verify` plugin call in the `deploy-all` task to work around the issue.


.gitignore
```diff
@@ -4,15 +4,14 @@ coverage/
 coverage.json
 deployments/
 dist/
-typechain-types
+typechain-types/
+venv/
```
mmv08, line 8: we do python now?

mmv08 (approved):

.gitignore
nlordell: The certora verification uses pip for installing the Certora CLI.

mmv08: I am not a Python expert, but it doesn't seem like the virtual env gets initialised anywhere in the script?

nlordell: No it does not - I use it locally, since our different projects use different versions of `certora-cli`, so installing it globally isn't great. I basically follow the instructions from:

https://github.com/safe-global/safe-locking?tab=readme-ov-file#formal-verification

Issue #395, opened 2024-04-22
Explore adding a ESM build to passkey/4337 packages


mmv08:
## Description

Currently, we're exporting all the builds as `commonjs` and a lot of tooling is expecting es module builds nowadays, and sometimes the interop mechanism has bugs (e.g., https://github.com/safe-global/safe-modules/pull/389#discussion_r1574665647)

Hardhat doesn't support ESM with TypeScript, so the only solution we can try is to add a second typescript config and use it for the build

mmv08:
I looked into the original error message and discovered the problem was in the `cbor`'s package incompatibility with the browser environment. I fixed it by switching to a web-compatible version, and, therefore, I propose not to add an ESM build. We can postpone that to later and see if we get any requests from the users (we do not market the passkeys package as the SDK for Safe + Passkeys and have a dedicated SDK in the safe-core-sdk)

PR #488, opened 2024-08-28
Update dependencies to use `cbor-web` instead of `cbor` in `@safe-global/safe-passkey` and update imports in the example app

mmv08:
When trying to solve https://github.com/safe-global/safe-modules/issues/395, I looked at the original error that triggered the discussion of adding an ESM build to the packages. I thought that the error was due to a bug in the bundler's ESM interoperability mechanism, but I realised that actually the issue was in the `cbor` package and compatibility with the browser environment.

When you import a js module, the code will get evaluated regardless of it use: https://www.perplexity.ai/search/does-javascript-evaluate-an-im-y4Zv1HKiQNmqxnHd1HJ1SA#0 (we may not use any of the functionality that's actually incompatible but because the code gets evaluated and it may imported a package that's unavailable in browser)

This PR:
- Fixes compatibility of the `@safe-global/safe-passkey` with the browser environment by switching from `cbor` to `cbor-web` package as recommended in the package [readme](https://github.com/hildjj/node-cbor)
- Use the `@safe-global/safe-passkey` code in the example app
- Updates the upstream bundler image TAG because it seems like they switched branch names

### DIFF
--- a/modules/passkey/package.json
+++ b/modules/passkey/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@safe-global/safe-passkey",
-  "version": "0.2.2",
+  "version": "0.2.1-1",
   "author": "@safe-global",
   "description": "Safe Passkey Owner",
   "homepage": "https://github.com/safe-global/safe-modules/tree/main/modules/passkey",
--- a/modules/passkey/package.json
+++ b/modules/passkey/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@safe-global/safe-passkey",
-  "version": "0.2.1",
+  "version": "0.2.2",
   "author": "@safe-global",
   "description": "Safe Passkey Owner",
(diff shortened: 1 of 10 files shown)


modules/passkey/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@safe-global/safe-passkey",
-  "version": "0.2.1",
+  "version": "0.2.2",
```
nlordell, line 3: I think we should use 0.2.1-1 here, since it is an update to supporting libraries and not SC code. Otherwise, it could get confusing WRT `safe-modules-deployments`.

nlordell (commented):
LGTM, just one comment on the version

modules/passkey/package.json
mmv08: Fair enough, will change

nlordell (approved):

PR #490, opened 2024-08-29
Refactor dependencies to use latest versions and update CI workflows for safe-modules-allowance.

mmv08:
This PR:
- Updates the dependencies to use the latest versions
- Adds a CI job for Safe Allowance module
- Fixes a interface name collision between `ISafe` used in tests and in the production code
- For some reason the tests were not running due to TS complaints, I had to be more specific with the installed packages for `pnpm` in the Safe Allowance module


nlordell (approved):
LGTM, but CI is failing.

.github/workflows/ci_allowance.yml
```diff
@@ -0,0 +1,43 @@
...
+        with:
+          version: 9
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 20.x
+          cache: pnpm
+      - run: |
+          pnpm install
+          pnpm run --filter "@safe-global/safe-allowance-module" lint
+          pnpm run --filter "@safe-global/safe-allowance-module" fmt
```
remedcu, line 42:
```suggestion
          pnpm run --filter "@safe-global/safe-allowance-module" fmt:check 
```

And in `package.json` we add a `"fmt:check": "prettier --check contracts/**/*.sol",`

modules/4337/certora/requirements.txt
```diff
@@ -1 +1 @@
-certora-cli==7.10.1
+certora-cli==7.10.2
```
remedcu, line 1: Ubernit: Newline

modules/allowances/package.json
```diff
@@ -26,8 +26,11 @@
   "license": "ISC",
```
remedcu, line 26: The `author` email is still `@gnosis.io`? 😅

modules/allowances/package.json
```diff
@@ -17,17 +17,25 @@
...
+    "fmt:check": "pnpm run fmt:sol:check && npm run fmt:ts:check",
+    "fmt:sol:check": "prettier 'contracts/**/*.sol' --check",
+    "fmt:ts:check": "prettier 'tasks/**/*.ts' 'test/**/*.ts' --check",
+    "fmt:sol:write": "prettier 'contracts/**/*.sol' --write",
+    "fmt:ts:write": "prettier 'tasks/**/*.ts' 'test/**/*.ts' --write"
   },
-  "author": "richard@gnosis.io",
-  "license": "ISC",
+  "author": "safe-global",
+  "license": "GPL-3.0",
```
akshay-ap, line 31: FYI: GPL-3.0 is deprecated. https://spdx.org/licenses/GPL-3.0.html

mmv08: This is the one we use in the other files, I'd raise this with legal. It says the identifier has been deprecated but not the license. We need a translation from legalese

mmv08: I changed it to GPL because this is the one we use in other files. I see it's deprecated and asked in Slack which one we should use.

Let's address this later in a different PR (I can also revert the change until we get more clarity)

modules/4337/certora/conf/Safe4337Module.conf
```diff
@@ -15,9 +15,10 @@
     ],
     "rule_sanity": "basic",
     "solc": "solc8.23",
+    "solc_allow_path":"../../node_modules",
```
mmv08, line 18: Previously there was a bug with solc's allow path option and certora-cli, now it's fixed we can use it and get rid of the crazy direct links

modules/allowances/package.json
nlordell: > This is the one we use in the other files, I'd raise this with legal. It says the identifier has been deprecated but not the license. We need a translation from legalese

The non-deprecated identifier is `GPL-3.0-only`.

nlordell: Basically, the `GPL-3.0` identifier was ambiguous because GPL has a "license update clause" (meaning that if they come up with a GPL-4, then all GPL-3 code will also be licensed under v4). In SPDX, this was represented by `GPL-3.0-or-later`. `GPL-3.0` changed to `GPL-3.0-only` to avoid confusion (as it is the GPL license opting out of the "license update clause").

modules/4337/certora/conf/TransactionExecutionMethods.conf
```diff
@@ -18,7 +18,7 @@
     "solc_allow_path":"../../node_modules",
     "verify": "Safe4337Module:certora/specs/TransactionExecutionMethods.spec",
     "packages": [
-        "@account-abstraction=../../node_modules/.pnpm/@account-abstraction+contracts@0.7.0/node_modules/@account-abstraction",
-        "@safe-global=../../node_modules/.pnpm/@safe-global+safe-contracts@1.4.1-build.0_ethers@6.13.2_bufferutil@4.0.8_utf-8-validate@5.0.10_/node_modules/@safe-global"
+        "@account-abstraction=node_modules/@account-abstraction",
+        "@safe-global=node_modules/@safe-global"
```
nlordell, line 22: :tada

modules/allowances/package.json
nlordell: Looking at the contract code, this should be `LGPL-3.0-only` (see the SPDX identifier used in the Solidity code).

mmv08:
I updated the licenses everywhere to be LGPL-3.0 only, except the recovery module, because I wasn't sure if we could change the license since it's a distribution of another project.

modules/4337/certora/requirements.txt
```diff
@@ -1 +1,2 @@
-certora-cli==7.14.3
+# TODO: remove beta when 7.21.0 is stable
```
mmv08, line 1: TODO: Merge when 7.21.0 is stable and switch back to certora-cli instead of the beta version

PR #492, opened 2024-09-09
L-1: Add a reset period check to prevent the transaction failing and using all gas

mmv08:
This PR fixes a low severity audit finding by Ackee:
> It is possible to set the resetTimeMin to 0. When the resetBaseMin is bigger
than 0 and the resetTimeMin is 0, then the branch with modulo operation is
triggered and it will cause a division by zero. Since it is also in Solidity version
<0.8.0, it consumes all gas.


akshay-ap (approved):
LGTM

nlordell (approved):

PR #508, opened 2025-03-27
Adds zksolc support for AllowanceModule

akshay-ap:
Changes in PR:
- Modify contract deployment for AllowanceModule and associated contracts
- Use imports for compiling contracts so that contracts can be deployed using `hardhat-deploy`

To do:
- Run all tests in ci
- Fix failing test in when running `npm run test:zk`


modules/allowances/test/test-helpers/deploySafeProxy.ts
```diff
@@ -1,27 +1,10 @@
...
-    data: iface.encodeFunctionData('createProxyWithNonce', [mastercopy, initializer, ZeroHash]),
-  })
-
-  return calculateProxyAddress(initializer, factory, mastercopy)
-}
-
-function calculateInitializer(owner: string): string {
-  const iface = new Interface(ArtifactSafe.abi)
+export function calculateInitializer(owner: string, zkSync: boolean = false): string {
+  const iface = new Interface(zkSync ? ArtifactSafeZk.abi : ArtifactSafe.abi)
```
mmv08, line 7: The abi should be the same regardless of the network, why is it needed?

modules/allowances/test/allowanceSingle.spec.ts
```diff
@@ -118,7 +121,7 @@ describe('AllowanceModule allowanceSingle', () => {
       await allowanceModule.setAllowance.populateTransaction(
         alice.address,
         ZeroAddress, // zero means ether
-        OneEther,
+        ethers.parseEther('1'),
```
mmv08, line 124: can use `OneEther` here too

modules/allowances/test/allowanceRecurring.spec.ts
```diff
@@ -15,14 +21,21 @@ describe('AllowanceModule allowanceRecurring', () => {
...
+    const { safe, allowanceModule, token, owner, alice, bob } = await setupTests()
 
     const safeAddress = await safe.getAddress()
     const tokenAddress = await token.getAddress()
 
     // add alice as delegate
     await execSafeTransaction(safe, await allowanceModule.addDelegate.populateTransaction(alice.address), owner)
 
+    // Set the timestamp to the current time for zk based networks
+    if (hre.network.zksync) {
```
mmv08, line 33: why is this needed? the comment should explain why, it can be seen what it is doing from code

modules/allowances/package.json
```diff
@@ -52,6 +60,10 @@
     "solidity-coverage": "^0.8.14",
     "ts-node": "^10.9.2",
     "typechain": "^8.3.2",
-    "typescript": "^5.7.2"
+    "typescript": "^5.7.2",
+    "zksync-ethers": "^6.16.2"
+  },
+  "dependencies": {
+    "patch-package": "^8.0.0"
```
mmv08, line 67: where is this dependency used?

akshay-ap: good catch. Its unused now. I was using it to patch dependencies initially

modules/allowances/test/allowanceRecurring.spec.ts
akshay-ap: When running with zk flag enabled, the node starts with a very low timestamp value (probably 0)  which causes the tests to fail because of underflow in calculating test values. So, this code sets the timestamp of the node to current value. I will add it in comment.

modules/allowances/test/test-helpers/deploySafeProxy.ts
akshay-ap: correct.

mmv08 (approved):

